PR #123, opened 2015-01-02
Add sqlite as dependency for RPM/Debian packages

fatalbanana:


vstakhov:
I think perl should be added here as well since snowball uses perl for building (but not in runtime, of course).

fatalbanana:
Ok added. :) Made sure this doesn't become a runtime dependency. ;)

PR #126, opened 2015-01-09
Fix building on Fedora

fatalbanana:


fatalbanana:
Forgot to reference issue in commit: https://github.com/vstakhov/rspamd/issues/88

vstakhov:
Merged, thank you!

PR #129, opened 2015-01-18
Fix build on SunOS

wiedi:
- the struct dirent d_type is not available on SunOS
- tm_gmtoff is not a POSIX field


vstakhov:
Merged, thank you!

PR #145, opened 2015-02-11
Various LUA plugin issues

eneq123:
1. headers.lua:235: attempt to concatenate local 'header_from' (a nil value)
2. once_received.lua:53: bad argument #3 to 'resolve_ptr' (string expected, got nil)
3. ratelimit.lua:188: attempt to index local 'from' (a nil value)

### DIFF
--- a/src/plugins/lua/once_received.lua
+++ b/src/plugins/lua/once_received.lua
@@ -48,7 +48,7 @@ local function check_quantity_received (task)
 			if not r['real_hostname'] or string.lower(r['real_hostname']) == 'unknown' or 
 				string.match(r['real_hostname'], '^%d+%.%d+%.%d+%.%d+$') then
 				
-				if r['real_ip'] and r['real_ip']:to_string() then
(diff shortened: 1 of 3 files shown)


src/plugins/lua/once_received.lua
```diff
@@ -48,10 +48,10 @@ local function check_quantity_received (task)
 			if not r['real_hostname'] or string.lower(r['real_hostname']) == 'unknown' or 
 				string.match(r['real_hostname'], '^%d+%.%d+%.%d+%.%d+$') then
 				
-				if r['real_ip'] then
+				if r['real_ip'] and r['real_ip']:to_string() then
```
vstakhov: Shouldn't we check here `:is_valid()` instead of `to_string()`?

eneq123: Oh, yes we should.

vstakhov:
Merged, thank you!

PR #154, opened 2015-02-12
rbl.lua: Don't ask for user unless called for by config

fatalbanana:


src/plugins/lua/rbl.lua
```diff
@@ -63,14 +63,16 @@ local function rbl_cb (task)
...
     (function()
-      if not notgot['user'] and not havegot['user'] then
-	havegot['user'] = task:get_user()
-	if havegot['user'] == nil then
-	  notgot['user'] = true
-	end
-      end
-      if havegot['user'] ~= nil and rbl['user'] == false then
-	return
+      if rbl['user'] == false then
```
vstakhov: false? is it logically correct?

fatalbanana: Yes, 'user' and 'default_user' answer the question: 'should I use this RBL if the user is authenticated?' so to activate the exclusion it is set to false. ;)

PR #172, opened 2015-02-15
rbl.lua: Support site-local IP address / subnet exclusions

fatalbanana:


fatalbanana:
It would be nice if this worked for IPv6. :) And if we could create radix trees from Lua for the is_private_ip stuff (assuming that may work better than what we're doing?).

PR #236, opened 2015-03-20
Add IPv6 loopback to secure_ip setting

fatalbanana:


conf/worker-controller.inc
```diff
@@ -3,5 +3,5 @@
 type = "controller";
 count = 1;
 password = "q1";
-secure_ip = "127.0.0.1";
+secure_ip = "::1 127.0.0.1";
```
vstakhov: I'd better use implicit array syntax:

    secure_ip = "127.0.0.1";
    secure_ip = "::1"

PR #270, opened 2015-04-23
extract all URLs from a text

awhitesong:
Here is the basic implementation for this issue. I have created a new function `lua_url_all`. Please review if any changes. Thanks.

### DIFF
--- a/src/lua/lua_url.c
+++ b/src/lua/lua_url.c
@@ -56,6 +56,7 @@ LUA_FUNCTION_DEF (url, to_table);
 LUA_FUNCTION_DEF (url, is_phished);
 LUA_FUNCTION_DEF (url, get_phished);
 LUA_FUNCTION_DEF (url, create);
+LUA_FUNCTION_DEF (url, all);
 
 static const struct luaL_reg urllib_m[] = {
 	LUA_INTERFACE_DEF (url, get_length),
@@ -73,6 +74,7 @@ static const struct luaL_reg urllib_m[] = {
 
 static const struct luaL_reg urllib_f[] = {
 	LUA_INTERFACE_DEF (url, create),
+	LUA_INTERFACE_DEF (url, all),
 	{NULL, NULL}
 };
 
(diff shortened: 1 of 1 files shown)


src/lua/lua_url.c
```diff
@@ -370,6 +372,49 @@ lua_url_create (lua_State *L)
 }
 
 static gint
+lua_url_all (lua_State *L)
+{
+	struct rspamd_url *url;
+	rspamd_mempool_t *pool = rspamd_lua_check_mempool (L, 1);
+	const gchar *text,*end;
+	gint length,i;
```
vstakhov: `i` used uninitialized.

src/lua/lua_url.c
```diff
@@ -370,6 +372,49 @@ lua_url_create (lua_State *L)
...
+	const gchar **pos;
+
+	if (pool == NULL) {
+		lua_pushnil (L);
+	}
+	else {
+		text = luaL_checkstring (L, 2);
+
+		if (text != NULL) {
+			length = strlen(text);
```
vstakhov: You can get length of the string by just using `luaL_checklstring (L, 2, &length);`

src/lua/lua_url.c
vstakhov: length should have type `size_t`.

src/lua/lua_url.c
```diff
@@ -370,6 +372,49 @@ lua_url_create (lua_State *L)
...
+			*pos=text;
+			end=text+length;
+			lua_newtable(L);
+			while(*pos<=end){
```
vstakhov: some style comment: there should be space before opening bracket and spaces before and after `<=`

src/lua/lua_url.c
```diff
@@ -370,6 +372,49 @@ lua_url_create (lua_State *L)
...
+				url = rspamd_url_get_next (pool, text, pos, NULL);				
+
+				if (url!=NULL) {
```
vstakhov: But if your `url` is NULL, you'll get a gap in lua table indicies, e.g. [1, 2, 4, 6]

src/lua/lua_url.c
```diff
@@ -370,6 +372,49 @@ lua_url_create (lua_State *L)
...
+					lua_pushinteger (L, i);
+					lua_pushlstring (L, url->string, url->urllen);
+					lua_settable (L, -3);
```
vstakhov: If `url` is NULL you should just call `break` there.

src/lua/lua_url.c
```diff
@@ -370,6 +372,49 @@ lua_url_create (lua_State *L)
...
+			length = strlen(text);
+			*pos=text;
+			end=text+length;
+			lua_newtable(L);
+			while(*pos<=end){
+				url = rspamd_url_get_next (pool, text, pos, NULL);				
+
+				if (url!=NULL) {
+					lua_pushinteger (L, i);
+					lua_pushlstring (L, url->string, url->urllen);
```
vstakhov: Shouldn't you push `rspamd{url}` here? See https://github.com/vstakhov/rspamd/blob/master/src/lua/lua_url.c#L358

awhitesong:
Hi, so i have modified and made specified changes. Just a little documentation is left for now. Please tell if any further changes needed. Thanks.

src/lua/lua_url.c
```diff
@@ -370,6 +372,55 @@ lua_url_create (lua_State *L)
...
+
+				if (url != NULL) {
+					lua_url = lua_newuserdata (L, sizeof (struct rspamd_lua_url));
+					rspamd_lua_setclass (L, "rspamd{url}", -1);
+					lua_url->url = url;
+					lua_pushinteger (L, i++);
+					lua_pushlstring (L, url->string, url->urllen);
+					lua_settable (L, -3);
+				}
+				else
```
vstakhov: Small style comment: in rspamd, every condition and loop *must* be enclosed within braces, even if it is one line. E.g.
```C
if (some)
   do();
else
   done();
```

should be written as:
```C
if (some) {
   do();
}
else {
   done();
}
```

awhitesong:
Changed the `if else` formatting.

vstakhov:
According to http://pgl.yoyo.org/luai/i/lua_settable `lua_settable` pops both key and value from the stack. I doubt that extra `lua_pop` is required in this case.

PR #271, opened 2015-04-23
Add pbkdf2 password support to controller

dark-al:


src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
 
+static gboolean
+rspamd_is_encrypted_password (const gchar * password)
+{
+        const gchar *start, *end;
+        gint64 id;
+        gsize size;
+        gboolean ret = TRUE;
+        
+        if (g_str_has_prefix (password, "$")) {
```
vstakhov: Why not `if (password[0] == '$')`?

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+                /* Parse id */
+                start = password + 1;
+                end = start;
+                size = 0;
+                while (g_ascii_isdigit (*end)) {
```
vstakhov: You should also check for `\0` here or you are introducing a security hole...

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+                /* Parse id */
+                start = password + 1;
+                end = start;
+                size = 0;
+                while (g_ascii_isdigit (*end)) { 
+                        size++;
+                        end++;
+                }
+                if (size) {
+                        id = g_ascii_strtoll (g_strndup (start, size), (char **) NULL, 10);
```
vstakhov: You don't need to do strdup here: just do the following:

```C
gchar *endptr;
id = strtoul (start, &endptr, 10);

if ((endptr != NULL && *endptr != *end) || id != VALID_ID) {
    return FALSE;
}
```

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+static const gchar *
+rspamd_encrypted_password_get_str (const gchar * password, gsize skip, gsize * length)
+{
+        const gchar *str, *start, *end;
+        gsize size;
+        
+        start = password + skip;   
+        end = start;
+        size = 0;
+        while (g_ascii_isdigit (*end) || g_ascii_isalpha (*end)) {
```
vstakhov: No check for `\0` again. It is crucial!
`g_ascii_isdigit (*end) || g_ascii_isalpha (*end)` can be replaced with `g_ascii_isalnum (*end)`. But strictly speaking you should check base32 symbols.

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+        
+        start = password + skip;   
+        end = start;
+        size = 0;
+        while (g_ascii_isdigit (*end) || g_ascii_isalpha (*end)) { 
+             size++;
+             end++;
+        }
+        if (size) {
+                str = g_strndup (start, size);
```
vstakhov: You don't need to duplicate string here, just return the position found and the size.

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+static gboolean
+rspamd_check_encrypted_password (const gchar * password, const gchar * check)
+{
+        const gchar *salt, *hash;
+        gsize salt_len, hash_len;
+	guint8 *key = NULL;
+        gboolean ret = TRUE;
+        
+        /* get salt */
+        salt = rspamd_encrypted_password_get_str (check, 3, &salt_len);
```
vstakhov: What if `salt` is NULL here and &salt_len is undefined? I'd predict crash...

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+        /* get hash */
+        hash = rspamd_encrypted_password_get_str (check, 3 + salt_len + 1, &hash_len);
+        /* decode salt */
+        salt = rspamd_decode_base32 (salt, salt_len, &salt_len);
```
vstakhov: Again, if you allocate `salt` previously then here you'd loose pointer to it and this would lead to a memory leak. If you will modify your `get_str` routine, then this won't happen

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+{
+        const gchar *salt, *hash;
+        gsize salt_len, hash_len;
+	guint8 *key = NULL;
+        gboolean ret = TRUE;
+        
+        /* get salt */
+        salt = rspamd_encrypted_password_get_str (check, 3, &salt_len);
+        /* get hash */
+        hash = rspamd_encrypted_password_get_str (check, 3 + salt_len + 1, &hash_len);
```
vstakhov: Cannot `hash` be NULL? Obviously it can!

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+        /* decode salt */
+        salt = rspamd_decode_base32 (salt, salt_len, &salt_len);
+        /* if we have salt and hash */
+        if (salt_len && hash_len) {
+	        key = g_malloc (BLAKE2B_OUTBYTES);
```
vstakhov: why not using just a static buffer, like `guchar key[BLAKE2B_OUTBYTES]`

src/controller.c
```diff
@@ -127,6 +130,94 @@ struct rspamd_controller_session {
...
+                rspamd_cryptobox_pbkdf (password, strlen(password), salt, salt_len, key, BLAKE2B_OUTBYTES, 16000);
+                key = rspamd_encode_base32 (key, BLAKE2B_OUTBYTES);
```
vstakhov: And you loose your malloc'ed pointer here again!

src/controller.c
```diff
@@ -1561,7 +1665,7 @@ init_controller_worker (struct rspamd_config *cfg)
 
 	rspamd_rcl_register_worker_option (cfg, type, "enable_password",
 		rspamd_rcl_parse_struct_string, ctx,
-		G_STRUCT_OFFSET (struct rspamd_controller_worker_ctx, password), 0);
+		G_STRUCT_OFFSET (struct rspamd_controller_worker_ctx, enable_password), 0);
```
vstakhov: Good catch!

PR #469, opened 2016-01-03
Unconditionally disable SPF for authenticated users

### LINKED ISSUE #443: spf check when authenticated
When sending myself an email, the R_SPF_SOFTFAIL	 symbol is added, but it shouldn't, as i'm sending thru an authenticated smtp session.

Received: from [IPv6:xxxxxxx] (unknown [IPv6:xxxxxx])
	by xxxx.xxx.xx (Postfix) with ESMTPSA id 0792CF11

IPs in received headers containing "with ESMTPSA" should be ignored by SPF checks

fatalbanana:
Shall we just unconditionally disable SPF for authenticated users?- I can't think of a reason someone would want it, so seems OK to me.

Fixes #443

### DIFF
--- a/src/plugins/spf.c
+++ b/src/plugins/spf.c
@@ -353,25 +353,31 @@ spf_symbol_callback (struct rspamd_task *task, void *unused)
 	struct spf_resolved *l;
 
 	if (radix_find_compressed_addr (spf_module_ctx->whitelist_ip,
-			task->from_addr) == RADIX_NO_VALUE) {
-		domain = get_spf_domain (task);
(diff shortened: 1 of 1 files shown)


xpunkt:
and add skip spf check to ipv4 127.0.0.1 and ipv6 ::1

fatalbanana:
We could do that using the existing (but not documented ;)) `whitelist` feature of SPF module though it'd only work where people are not applying that in their configs already. Private IP map in RBL module config could be suitable: perhaps we should have such a map somehow built-in.

PR #567, opened 2016-03-22
[Feature] Add list support to `mime types` module configuration

moisseev:
- this change allows to configure several mime types for an extension like
```txt = [ "message/disposition-notification", "text/plain", "text/rfc822-headers" ];```
- `type` variable replaced with `mtype` as `type` is lua reserved word


vstakhov:
That looks good, thank you!

PR #608, opened 2016-04-28
[Minor] Improve subject rules

### LINKED ISSUE #548: Why SUBJ_ALL_CAPS isn't in metrics.conf
It might not be an issue. However, I'd like to know why SUBJ_ALL_CAPS isn't in metrics.conf?

Another question is if I want all scores in metrics.conf divided by 3 since the reject threshold with my server need to be 5. might there be some problems? I already know one because SUBJ_ALL_CAPS isn't in metrics.conf. Any other symbols are not in metrics.conf ?  Thanks.

moisseev:
Make scores depend on subject length
Also resolves #548


vstakhov:
I think I have an idea to leave this as rules, not as symbols

vstakhov:
I did it :)

PR #618, opened 2016-05-02
Updates to phishing module

moisseev:


conf/metrics.conf
```diff
@@ -833,8 +833,9 @@ metric {
 
     group "phishing" {
         symbol "PHISHING" {
-            weight = 3.0;
```
vstakhov: I think this weight change might be dangerous.

moisseev: I'll leave it up to you.
Without `one_shot = true;` I had sometimes more than 12. Considering nobody has complained about it for a long time I think `6.0` should be safe, at least it is better than it was. On the other hand a lot of legitimate mail use redirectors, so I understand your concern.

vstakhov: I have some larger collections of redirectors and I **really**, no **REALLY** should write some documentation about rspamd url redirector that has been used for many years in Rambler and proved its efficiency.

moisseev: Then would you mind to share your collection by adding it to the redirectors white list?

PR #636, opened 2016-05-26
[Feature] Add throughput graph for RRD backend to WebUI

moisseev:
The `Throughput` button is disabled. To enable it uncomment corresponding line in the `index.html`.

The reasons:
1. The widget is highly experimental and requires some improvements. In particular data representation is not pretty good, especially on long time intervals.
2. There are issues with the RRD backend: https://github.com/vstakhov/rspamd/issues/615#issuecomment-220404928, https://github.com/vstakhov/rspamd/issues/405#issuecomment-188132960

I have an idea to enable the button conditionally (if RRD is enabled in rspamd configuration) when 1 and 2 will be resolved.

### DIFF
--- a/interface/css/rspamd.css
+++ b/interface/css/rspamd.css
@@ -653,3 +653,11 @@ td.maps-cell {
 #historyLog_wrapper div.row:last-child > div {
     padding: 5px 20px 0 20px;
     }
+
+/* Throughput graph controls */
+#graph_controls select {
+    margin: 10px 20px 0;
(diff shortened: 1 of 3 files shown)


vstakhov:
Wow, thank you! I just think that d3evolution should be somehow bundled in webui.

moisseev:
I merely don't want to pollute the rspamd repository. Taking into account the reason 1, I think it might be better if you keep it as external links for now.

PR #638, opened 2016-05-26
Correct migration.md, small change to writing_rules.md

larryhynes:


vstakhov:
Thank you very much for this work!

PR #656, opened 2016-06-09
Correct architecture/index.md

larryhynes:
Line 76, in the new version of `architecture/index.md`, says: "Therefore, it is recommended to increase the size of such statfiles." I would like to see an explanation of *how* to do that, or recommend you just delete that line.

I don't know what 'form URL encoding' means in the first paragraph (Protocol basics) of `architecture/protocol.md`.


moisseev:
> Therefore, it is recommended to increase size for such statfiles.

Another question: is there some criteria how much statistics is enough? Statfiles of any size will be filled eventually.

I've already hit the limit and really need advice.
~~~
Statfile: BAYES_SPAM type: sqlite3; length: 92.08M; free blocks: 0; total blocks: 1.28M; free: 0.00%; learned: 4772; users: 1; languages: 3
Statfile: BAYES_HAM type: sqlite3; length: 95.69M; free blocks: 0; total blocks: 1.29M; free: 0.00%; learned: 5857; users: 1; languages: 3
Total learns: 10629
~~~

fatalbanana:
@larryhynes To shrink/grow (old mmap backend) statfiles: set desired size in config & restart rspamd.

@moisseev Ideal size might be about as much as you can comfortably fit in RAM. MMap backend is incompatible with newer backends: you should probably migrate.

larryhynes:
I'm currently using redis as the backend. It tells me there's `0.00%` free and `0` free blocks, but it seems to be doing OK. :smile:

```
Statfile: BAYES_HAM type: redis; length: 15.23M; free blocks: 0; total blocks: 400.99k; free: 0.00%; learned: 1015; users: 62; languages: 0
Statfile: BAYES_SPAM type: redis; length: 6.45M; free blocks: 0; total blocks: 169.91k; free: 0.00%; learned: 742; users: 8; languages: 0
```

My point in the comment with the PR relates to the documentation, it wasn't in relation to my current `rspamd` setup. I feel that if that line about increasing the size of statfiles is to stay in the documentation then it should expand a little on how to increase it, or else just not mention it at all.

Now that you have piqued my interest in my own setup, I've been labouring under the assumption that redis will automatically increase the size of its dump as and when required e.g. as new tokens are added the size of the redis dump will grow. Also, I see nothing in the `statistic.conf` file to suggest there is a setting for `size`. Should I be doing something else here?!?

fatalbanana:
I might be inclined to suggest removing it (& expanding on something like this in statistics documentation instead).

Redis/sqlite statistics will grow forever. In case of Redis: you can use its `maxmemory` setting to limit DB size (writes will fail when this limit is reached; failure to set this could mean Redis crashing and/or performance degrading when database gets too large).

moisseev:
Is any non-zero values of `free blocks` and `free` possible for `sqlite/redis` in `rspamc stat` output? Maybe it should be removed for all backends except `mmap`?

vstakhov:
Thank you, @larryhynes I really appreciate your help!

larryhynes:
Very welcome; thanks for such a great spam filter! :smile:

PR #664, opened 2016-06-15
[Minor] Add `application/octet-stream` mime type for `pdf` extension

moisseev:
The `application/octet-stream` mime type is commonly used  for `pdf` files by variety of automated mailing systems.


vstakhov:
I was hit by this issue as well indeed.

PR #693, opened 2016-07-02
[Test] Add & rework functional tests

fatalbanana:
- Machine formatted test data
 - More idiomatic use of test framework
 - Added tests for fuzzy & redis statistics


test/functional/cases/statistics/compat-plain.robot
```diff
@@ -0,0 +1,19 @@
+*** Settings ***
+Suite Setup     Statistics Setup
+Suite Teardown  Statistics Teardown
+Default Tags    broken
```
fatalbanana: Do you plan to fix it? (mmap statistics backend) :) Should we be testing this?

PR #730, opened 2016-07-19
[Fix] Deal with certain nested dependencies in SpamAssassin module

fatalbanana:
- Also some minor cleanup and fixing of logging
 - Import fun.lua stuff into main namespace

### DIFF
--- a/src/plugins/lua/spamassassin.lua
+++ b/src/plugins/lua/spamassassin.lua
@@ -25,8 +25,6 @@ local rspamd_trie = require "rspamd_trie"
 local util = require "rspamd_util"
 require "fun" ()
 
---local dumper = require 'pl.pretty'.dump
-
 -- Known plugins
 local known_plugins = {
   'Mail::SpamAssassin::Plugin::FreeMail',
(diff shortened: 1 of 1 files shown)


src/plugins/lua/spamassassin.lua
```diff
@@ -1443,28 +1436,28 @@ local function post_process()
...
       rules))
 
   -- Check meta rules for foreign symbols and register dependencies
-  _.each(function(k, r)
+  each(function(k, r)
       if r['expression'] then
         local expr_atoms = r['expression']:atoms()
 
         for i,a in ipairs(expr_atoms) do
-          if not atoms[a] then
```
vstakhov: It is slightly wrong. You need to register dependency not on `a` but on external dependencies of `a`. Furthermore, you need to adjust the current external deps for this meta to include dependencies from `a`. And finally, you need to perform this operation unless there are no changes in deps lists.

src/plugins/lua/spamassassin.lua
```diff
@@ -23,9 +23,7 @@ local rspamd_expression = require "rspamd_expression"
 local rspamd_mempool = require "rspamd_mempool"
 local rspamd_trie = require "rspamd_trie"
 local util = require "rspamd_util"
-local _ = require "fun"
```
vstakhov: I thought about it, though I suppose this change should be in a separate commit.

PR #754, opened 2016-07-26
[WIP] Try avoid sqlite locking issues

fatalbanana:


vstakhov:
Hum, but we have lock here granted for the particular process. Why EBUSY could occur here I wonder?

fatalbanana:
Let me see if I can get tests to fail again and try figure out what if anything is appropriate to do here (currently I am struggling to reproduce failures [running for hours without failures]).

Issue #748, opened 2016-07-24
rspamd: writev (127.0.0.1), Connection refused


xeyownt:
I'm still trying to install rmilter/rspamd on a debian Jessie/postfix/dovecot stack. So far I managed to get everything to work correctly, including rspamd webui, but unfortunately by editing some of the package files (`/lib/systemd/system/*`) to force ipv4 sockets.

I'd like to revert all my changes in these files, and only keep configuration changes in /etc as it should be.

Currently, file `/lib/systemd/system/rspamd.socket` contains this:

```
[Socket]
ListenStream=11333
ListenStream=[::1]:11334
BindIPv6Only=both
```
With this configuration, on my system, the port 11334 is only open on ipv6. This is not a problem because I can tell Apache to reverse proxy to ipv6 [::1]:11334, and this works. Note that I can't connect to 127.0.0.1:11334 (despites the BindIPv6Only option).

The problem is that I can't tell rmilter to connect to [::1]:11333. Depending on whether I use square brackets or not, I get parsing errors.

```
... rmilter[15852]: config file parse error! line: 49, text: [::1], reason: syntax error
... ober rmilter[15852]: yyparse: cannot parse config file, 1 errors
... ober rmilter[15892]: config file parse error! line: 49, text: :1:11333, reason: syntax error
... ober rmilter[15892]: yyparse: cannot parse config file, 1 errors
```
And using address 127.0.0.1:11333, I get the errors:
```
... rmilter[1832]: <7a6b3b759d>; spamdscan: start scanning message on 127.0.0.1
... rmilter[1832]: <7a6b3b759d>; rspamd: writev (127.0.0.1), Connection refused
... rmilter[1832]: <7a6b3b759d>; spamdscan: finish scanning message on 127.0.0.1
... rmilter[1832]: <7a6b3b759d>; rspamdscan: failed to scan, retry, 127.0.0.1, /tmp/msg.XX57AIw6
```
I tried all possible settings for rspamd `secure_ip` parameter (either "127.0.0.1", "::1", "::1 127.0.0.1" or even empty). It seems the problem is more in the fact that the ipv6 socket is not open for both ipv4/ipv6, despite the systemd config and that my system seems correctly configured:

```bash
sysctl -a|grep v6|grep bind
# net.ipv6.bindv6only = 0
```

fatalbanana:
`BindIPv6Only` won't apply in case listening address is specified, so controller listens only on IPv6 under systemd socket activation.

[...]

Issue #740, opened 2016-07-21
rspamc stat fails on out-of-the-box install on Debian Jessie


xeyownt:
I follow the quickstart guide from https://rspamd.com/doc/quickstart.html.
My system is a Debian Jessie on VPS, with systemd.

```bash
apt-get --no-install-recommends install rspamd
dpkg -l | grep rspamd
# ii  rspamd              1.2.8-1~jessie        amd64     Rapid spam filtering system
systemctl enable rspamd.socket
systemctl start rspamd.socket
rspamc stat
# Results for command: stat (0.000 seconds)
# IO write error: Connection refused
```

Note that `systemctl enable rspamd.socket` is not necessary because it was enabled already.

Now I can get it to work with

```bash
rspamc -h ::1:11334 stat
# Results for command: stat (5.980 seconds)
# Messages scanned: 0
# ...
rspamc -h ::1:11334 stat
# Results for command: stat (0.004 seconds)
# Messages scanned: 0
# ...
```
First start is very slow because service must be started.

Oddly enough it does not work on localhost ipv4:
```bash
rspamc -h 127.0.0.1:11334 stat
# Results for command: stat (0.000 seconds)
# IO write error: Connection refused
```

Is it intended behaviour? If so, maybe the quickstart guide should be updated with the new rspamc line.
 
It took me several hours to figure this out, first because I had ipv6 disabled on my kernel, and second because of trying to understand systemd stuff and obscure `rspamd.conf` file (`bind_socket = "systemd:0"`, what does this mean?).

fatalbanana:
`systemd:0` means first systemd socket (FD 3), `systemd:1` is the second (FD 4) and so on.

There are indeed issues with IPv4 vs IPv6 under systemd and quickstart should be updated. Something like `rspamc < /dev/null` should work.

xeyownt:
I confirm your suggestion works:

```bash
systemctl stop rspamd.service                                                                                   
# Warning: Stopping rspamd.service, but it can still be activated by:
#  rspamd.socket
systemctl status rspamd.service                                                                       
# ● rspamd.service - rapid spam filtering system
#    Loaded: loaded (/lib/systemd/system/rspamd.service; disabled)
#    Active: inactive (dead) since Thu 2016-07-21 19:16:09 CEST; 5s ago
# ...
```
We see that service is inactive. Now let's trigger it:
```bash
rspamc < /dev/null
# Results for file: stdin (0.000 seconds)
[...]

Issue #452, opened 2015-12-16
How to get more information about IO timeout errors?


vadz:
I'm using `rspamc` from Exim as a pipe command and globally it works just fine but I've already got 3 times (for ~1000 messages) an apparently bogus timeout error, i.e. the received (spam) message has the following headers:

```
X-Spam-Scanner: rspamc 1.1.0
X-Spam-Scan-Time: 5.005
X-Spam-Error: IO timeout
```

and there is no trace of it in `rspamd.log` at all. Running `rspamc` with it on the comment line successfully recognizes it as spam and takes less than one second (I tried it several times and there is a pretty big variation which is also strange, but it's never close to 5s).

I suppose that there must have been some transient network problem which prevented an RBL lookup from completing or something like this, but how can I really know it? I.e. is there any way to get more information in case a timeout occurs, e.g. add some description of the operation that timed out to `X-Spam-Error` header? Otherwise I'm really not sure how am I supposed to prevent this from happening in the future...

vstakhov:
Well, that's unfortunately exim issue. I don't know how to setup exim timeout for rspamd reply waiting to tell the truth. Obviously, high scan times could be caused by slow DNS replies, for example. There is also parameter called `task_timeout` that limits the overall processing time allowed for a single task. You might want to set it a bit lower than exim timeout, for example by the following configuration:

```
worker {
  type = "normal";
  ...
  task_timeout = 4s;
}
```

Please mention that you need to modify **worker's** section not `options` one. In future, the corresponding option might be added as well.

fatalbanana:
Default timeout on Exim's pipe transport is `1h` which should be plenty. You can increase timeout used by `rspamc` by passing `-t` flag.

vadz:
Thanks for the tip about `rspamc -t`, this does mitigate the problem somewhat, but it still happens, albeit more rarely. I think it would be much better if rspam could just skip the DNS lookup or whatever that is taking so long (I still have no idea what it is...) and use the other checks as it seems all but certain that most of the messages that now end up in my inbox with `X-Spam-Error: IO timeout` header would have been recognized as spam anyhow.

And I can confirm that it is definitely _not_ an Exim issue.

vstakhov:
Well, I've added logic to remove delayed filters when timeouting a task. However, please mention that it won't work for post-filters.

vadz:
[...]

Issue #430, opened 2015-11-21
Rspamd and Exim "Connection reset by peer"


Twilek-de:
I am trying to use Rspamd (1.0.10) with Exim (4.86). I have implemented the options as posted in the Wiki. When a new message comes in Exim log will show:

2015-11-21 15:53:18 1a09XS-0004jF-Ht spam acl condition: error reading from spamd socket: Connection reset by peer
2015-11-21 15:53:18 1a09XS-0004jF-Ht H=mout.gmx.net [212.227.17.21] Warning: ACL "warn" statement skipped: condition test deferred

Rspamd log will show
(normal) <06859d>; task; accept_socket: accepted connection from ::ffff:127.0.0.1 port 44402
(normal) <06859d>; task; rspamd_message_parse: loaded message; id: <trinity-2fe66e10-3e6a-4c3a-9beb-65aecae8b3ef-1448117443853@3capp-gmx-bs55>; queue-id: <1a09Uy-0004dC-6i>
(normal) <06859d>; task; lua_metric_symbol_callback: call to (HTML_SHORT_LINK_IMG_1) failed: /usr/share/rspamd/rules/html.lua:32: attempt to index local 'hc' (a nil value); trace: [1]:{/usr/share/rspamd/rules/html.lua:32 - <unknown> [Lua]};
(normal) <06859d>; task; lua_metric_symbol_callback: call to (R_EMPTY_IMAGE) failed: /usr/share/rspamd/rules/html.lua:82: attempt to index local 'hc' (a nil value); trace: [1]:{/usr/share/rspamd/rules/html.lua:82 - <unknown> [Lua]};
(normal) <06859d>; task; rspamd_protocol_http_reply: id: <trinity-2fe66e10-3e6a-4c3a-9beb-65aecae8b3ef-1448117443853@3capp-gmx-bs55>, qid: <1a09Uy-0004dC-6i>, ip: 212.227.15.19, from: xxx@xxx.de, (default: F (no action): [0.39/15.00] [RCVD_IN_DNSWL_LOW,MIME_HTML_ONLY,HFILTER_MID_NOT_FQDN,R_SPF_ALLOW,RWL_MAILSPIKE_POSSIBLE]), len: 1689, time: 1014.103ms real, 14.551ms virtual, dns req: 23,
(normal) <c9675d>; task; accept_socket: accepted connection from ::ffff:127.0.0.1 port 44674
[...]

PR #762, opened 2016-07-27
[Rework] Remove systemd socket activation (Fixes #759)

### LINKED ISSUE #759: Fix systemd socket activation fiasco
Remaining known issues with systemd socket activation are:

1) Controller listens on IPv6 only, which means that controller commands like `rspamc stat` might not work without passing `-h` depending on system configuration.
2) Socket file will fail on systems without IPv6 support

~~I believe I can fix both. Unfortunately, it seems that we will need two `.socket` files where we used to have one, which is potentially disruptive & confusing to users. But I think the change is worth it & we could try manage this as best possible in packaging.~~ Actually multiple socket files results in unpredictable behaviour, and stuff we need to deal with this is not available in versions of systemd we want to support, so just give up all hope & rip out socket activation completely?

3) Socket activation is unfamiliar & confusing to many users.

Is not really something we can fix. Could be addressed somewhat by improving documentation. If we rip out socket activation it becomes a non-issue.

fatalbanana:
Confirmed to provide a clean & reboot-proof upgrade-path on CentOS 7, Debian Jessie & Ubuntu Xenial- so I suppose we can ship it.

TODO:

- Update documentation
- Update rspamd-build-scripts
- rmilter


fatalbanana:
Fixes #430 
Maybe fixes #452
Fixes #748 
Fixes #740

Glandos:
Could you please at least keep the .socket file, without enabling it at install. I have both IPv4 and IPv6, and I experienced no issue with that.
Socket activation is great for fast boot, which is important in my use case. I don't reboot often, but when I do, I appreciate the low 2 minute downtime.

fatalbanana:
Packaging will likely break socket-activated setups on upgrade so providing this as an option requires some careful testing with/without socket activation on different platforms (and probably changes to packaging) which I don't have time for.

jhass:
I'm not sure I follow. Either the package installs a socket activated setup or it doesn't. Then on upgrade

* It does nothing if it's not common for the distro to handle updates (eg. Archlinux).
* It does a `systemctl daemon-reload`.
* It does a `systemctl restart rspamd.service` for non-socket activated setups.
* It does a `systemctl stop rspamd.server` for socket activated setups.

fatalbanana:
Packaging maintained within rspamd project (Debian/RPM) does _not_ use socket-activation anymore (and we don't recommend using it). Something like `restart rspamd.service` happens on upgrade- which breaks socket-activated setups. Packaging also enables `rspamd.service` to run on startup- which breaks socket-activated setups. 3rd party packagers can still do whatever they like.

PR #776, opened 2016-07-31
[Test] Some minor fixes for tests

fatalbanana:
I think to just change behaviour of `Check Rspamc` as functionality that I've baked in seems both useless & annoying.


fatalbanana:
Adding this to #775

PR #814, opened 2016-08-05
[Minor] Fix plugins

fatalbanana:


fatalbanana:
Errr, nevermind- we still want this AFAIU.

PR #818, opened 2016-08-06
[Minor] Don't claim lua is fast

frozencemetery:
Neither the linked benchmark nor [this other one people use a lot](https://benchmarksgame.alioth.debian.org/u64q/compare.php?lang=lua&lang2=gcc) show Lua to perform comparably to C.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,12 +92,17 @@ against a set of multiple patterns. Or it uses radix bit trie for storing IP add
 
 ## Extensions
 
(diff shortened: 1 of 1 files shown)


vstakhov:
Lua is fast for the purposes of Rspamd. LuaJIT is even faster that has been proven by a lot of practical experiences. I don't care about some damned artificial benchmarks at all.

PR #833, opened 2016-08-10
[WIP] Some fixes for code coverage

fatalbanana:
Unfortunately does not deal with the worst of problems


circle.yml
```diff
@@ -11,6 +11,7 @@ test:
     - cmake --build . --target install
     - sudo robot test/functional/cases
     - lcov --no-external -b . -d . -c --output-file coverage.info
+    - lcov -r coverage.info *.rl
```
fatalbanana: Meh... Actually I'm not sure: if it makes more sense to track coverage of `.rl` files (guessing that works), or the generated `.rl.c` files, or both. Coverage of these parts is broken anyway- issue there appears to be to do with paths. Unsure of issue(s) with other missing coverage (but could be somewhat similar).

PR #837, opened 2016-08-11
[Feature] rspamd_stats: support log directory reading

moisseev:
rspamd_stats will read (and decompress) multiple log files in the specified directory.

### DIFF
--- a/utils/rspamd_stats.pl
+++ b/utils/rspamd_stats.pl
@@ -339,7 +339,7 @@ B<Total hits>: total number of hits and percentage of symbol hits divided by tot
 
 =item *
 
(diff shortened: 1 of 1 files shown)


utils/rspamd_stats.pl
```diff
@@ -13,9 +13,18 @@
...
+my $exclude_logs = 0;
 my $man = 0;
 my $help = 0;
 
+# Associate file extensions with decompressors
+my %decompressor = (
+    'bz2' => 'bzcat',
+    'gz'  => 'zcat',
+    'xz'  => 'xzcat',
+);
```
vstakhov: All these tools are absent on Linux

PR #863, opened 2016-08-19
Allow for applying ratelimit per bare IP or ASN

fatalbanana:


fatalbanana:
Deferring until adaptive buckets are done (and for 1.4).

PR #910, opened 2016-08-30
[Minor] Update rule style: headers.lua

fatalbanana:


vstakhov:
That's awesome: I wanted to do it many times but was constantly scared about this work.

PR #912, opened 2016-08-30
[Minor] Disable IP Score for authenticated users & local networks

fatalbanana:


moisseev:
We also need a list (or maybe 2 separate lists) of checks disabled for users and local networks documented somewhere. For now configuring of outgoing mail scanning is blindfold.

fatalbanana:
Current approach seems to be to flood logs with messages indicating scanning is disabled (SPF/DKIM/DMARC do this; greylist/RBL/hfilter unfortunately do not). Irrespective of that, we should probably have such a reference indeed (could form part of a resource about outbound scanning since that seems to be a point of interest as well as confusion for many).

PR #915, opened 2016-08-30
dnssec is default enabled, fixing that

xpunkt:
dont blame me


vstakhov:
It is disabled by default.

PR #936, opened 2016-09-08
[Feature] Add R_DKIM_PERMFAIL symbol

fatalbanana:


src/plugins/dkim_check.c
```diff
@@ -560,62 +586,71 @@ dkim_module_check (struct dkim_check_result *res)
...
-					g_list_prepend (NULL,
-							rspamd_mempool_strdup (sel->task->task_pool,
-									rspamd_dkim_get_domain (sel->ctx))));
-		}
-	}
-
-	if (all_done) {
-		rspamd_session_watcher_pop (res->task->s, res->w);
-	}
+	rspamd_session_watcher_pop (res->task->s, res->w);
```
vstakhov: Shouldn't it be done merely if `all_done` is TRUE?

src/plugins/dkim_check.c
```diff
@@ -560,62 +586,71 @@ dkim_module_check (struct dkim_check_result *res)
...
-					sel = cur;
-				}
+			else if (cur->res == DKIM_CONTINUE) {
+				symbol = dkim_module_ctx->symbol_allow;
+				symbol_weight = cur->mult_allow * 1.0;
+				messages = g_list_prepend (messages,
+						rspamd_mempool_strdup (cur->task->task_pool,
+						rspamd_dkim_get_domain (cur->ctx)));
+			}
+			else if (!got_allow && cur->res == DKIM_PERM_ERROR && cur->err->message != NULL) {
```
vstakhov: What if `cur->err` is NULL? Or it cannot be NULL?

src/plugins/dkim_check.c
fatalbanana: I suppose so. I'm not sure why I have it otherwise. ;)

src/plugins/dkim_check.c
fatalbanana: It should never be NULL in case of DKIM_PERM_ERROR but it doesn't hurt to check this properly.

PR #951, opened 2016-09-14
[Feature] Add `mid` Lua module

moisseev:


src/plugins/lua/mid.lua
```diff
@@ -0,0 +1,112 @@
...
+      for _,map in ipairs(maps) do
+        local v = map:get_key(dkim_domain)
+        if v then
+          if v == '' then
+            if not header then
+              task:insert_result(settings['symbol_known_no_mid'], 1, dkim_domain)
+              return
+            end
+          else
+            re[dkim_domain] = rspamd_regexp.create_cached(v)
```
vstakhov: This can return `nil` if a regexp is incorrect.

vstakhov: Another issue with this approach is that regexps are indefinitely stored in cache in this case and are not updated upon subsequent map changes. However, this might be regexp cache issue indeed.

src/plugins/lua/mid.lua
```diff
@@ -0,0 +1,112 @@
...
+  end
+end
+
+local opts =  rspamd_config:get_all_opt('mid')
+if opts then
+  for k,v in pairs(opts) do
+    settings[k] = v
+  end
+
+  if settings['url'] and #settings['url'] > 0 then
```
vstakhov: In fact, each map now can contain multiple sources. Hence, your logic here seems incorrect to me.

src/plugins/lua/mid.lua
moisseev: What would you suggest? Maybe I should invert plugin logic?
Create regexp map and check Message-ID against it. If match found check DKIM for corresponding map value (domain).

vstakhov: For the first issue, let's assume that incorrect regexp is a warning and add symbol in this case (like if it was no regexp).
For the second issue, I'll think about expiration to the regexp cache (so you can leave `create_cached` safely).

PR #1012, opened 2016-10-07
[Fix] Avoid segfault in settings module

fatalbanana:


fatalbanana:
Ugh. Nope, that is infact not a suitable fix.

PR #1013, opened 2016-10-07
[Minor] SPF: Request one of A/AAAA records only

fatalbanana:
Currently, `a` elements in SPF records will cause PERMFAIL unless both A & AAAA records are available. The RFC recommends looking up one or the other depending on the type of IP connecting:

`When any mechanism fetches host addresses to compare with <ip>, when
   <ip> is an IPv4, "A" records are fetched; when <ip> is an IPv6
   address, "AAAA" records are fetched.
`

... and by doing this we mitigate the issue to some extent.


fatalbanana:
That will not play well with caching ...

PR #1019, opened 2016-10-10
[Feature] Some composites to unify SPF/DKIM/DMARC scoring

fatalbanana:
What do you think about this? Idea is to have somewhat consistent scoring for missing/failed/good auth.


fatalbanana:
Not fond of it I suppose. ;)

PR #1034, opened 2016-10-16
Added copytruncate arg into logrotate config

qk4l:
Hi,

On Ubuntu 16.04 I noticed that after logrotate work rspamd continue write logs to old file which was renamed.
Seems SIGUSR1 (reopenlog call) not work properly and need to fix.

```
$/var/log/rspamd# ls -la
total 32
drwxr-xr-x  2 _rspamd _rspamd 4096 Oct 16 13:41 .
drwxr-xr-x 15 root    syslog  4096 Oct 16 06:25 ..
-rw-r--r--  1 _rspamd _rspamd *5397* Oct 16 13:48 rspamd.log.1
-rw-r--r--  1 _rspamd _rspamd  749 Oct 16 13:41 rspamd.log.2.gz
-rw-r--r--  1 _rspamd _rspamd 3257 Oct 16 13:40 rspamd.log.3.gz
-rw-r--r--  1 _rspamd _rspamd 4274 Oct 16 13:39 rspamd.log.4.gz
root@mxeu1:/var/log/rspamd#  service rspamd reopenlog 
 * Reopen logs for rapid spam filtering system rspamd
   ...done.
$/var/log/rspamd# rspamc < ~test.mail
... some output ..

root@mxeu1:/var/log/rspamd# ls -la
total 32
drwxr-xr-x  2 _rspamd _rspamd 4096 Oct 16 13:41 .
drwxr-xr-x 15 root    syslog  4096 Oct 16 06:25 ..
-rw-r--r--  1 _rspamd _rspamd *6615* Oct 16 13:49 rspamd.log.1
-rw-r--r--  1 _rspamd _rspamd  749 Oct 16 13:41 rspamd.log.2.gz
-rw-r--r--  1 _rspamd _rspamd 3257 Oct 16 13:40 rspamd.log.3.gz
-rw-r--r--  1 _rspamd _rspamd 4274 Oct 16 13:39 rspamd.log.4.gz
```

Actually logrotate provide  _copytruncate_  behavior in which it copy data from original file (ex.rspamd.log) to rotated one so application continuous write to same file.

### DIFF
--- a/debian/rspamd.logrotate
+++ b/debian/rspamd.logrotate
@@ -4,6 +4,7 @@
        delaycompress
        compress
        notifempty
+       copytruncate
        missingok
        postrotate
           service rspamd reopenlog >/dev/null 2>&1 || true


vstakhov:
There is something with your setup and not with Rspamd. Rspamd correctly processes `USR2` and reopens log files. Your proposed change is, on the contrary, not correct as it introduces extra cost of logs rotation which is not needed.

qk4l:
@vstakhov you are right. I shoud check if rspamd preccess `SIGUSR1` by sending it directly by `kill`.
In my installation rspamd started by systemd daemon and systemd do not create PID file so `do_reopenlog()` in init rc script can not send signal

```
do_reopenlog() {
        echo "start-stop-daemon --stop --signal 10 --quiet --pidfile $PIDFILE"
        start-stop-daemon --stop --signal 10 --quiet --pidfile $PIDFILE
        return 0
}
```

```
● rspamd.service - rapid spam filtering system
   Loaded: loaded (/lib/systemd/system/rspamd.service; enabled; vendor preset: enabled)
   Active: active (running) since Sun 2016-10-16 14:47:44 MSK; 2h 15min ago
 Main PID: 30675 (rspamd)
```

P.S. Installed `rspamd` from deb package.

vstakhov:
If you use systemd then it is assumed by Rspamd package that systemd manages all logs and their rotation. You are trying to mix systemd-adopted installation with rc scripts. That won't work this way. I have very little experience with systemd (and I don't regret about it!) and I cannot thus provide more details.

qk4l:
> You are trying to mix systemd-adopted installation with rc scripts.

I don't want to do that. As a user I took the deb package from rspamd repo and installed it on Ubuntu 16.04 (it uses systemd =( ). Then I configured `rspamd` to write logs into file (type=file) because I don't want to use systemd for logging and got this problem.

Please reject this PR.
To resolve this problem `-p` argument should be added to ExecStart

```
[Service]
NonBlocking=true
ExecStart=/usr/bin/rspamd -c /etc/rspamd/rspamd.conf -f -p /run/rspamd/rspamd.pid
User=_rspamd
RuntimeDirectory=rspamd
RuntimeDirectoryMode=0755
Restart=always
```

PR #1038, opened 2016-10-16
[Feature] Whitelist hostnames for greylisting

AlexeySa:


moisseev:
Correct name of `the last 2octet in a hostname` is `second-level domain` (SLD). An octet is an eight-bit byte.
In fact, we need to whitelist `effective second-level domain` (eSLD). So, interesting domain name part can have 2 dots (and 3 domain parts) as well, for instance: https://github.com/vstakhov/rspamd/pull/1039/commits/a76420b8ef6180e96732eee38328042786fed49a#diff-12d1ae862532ac757f3113c840c820e7R178

AlexeySa:
Whitelist third and second level domains https://github.com/vstakhov/rspamd/pull/1041

PR #1067, opened 2016-10-24
[Feature] Allow for excluding messages from AV scanning based on size

fatalbanana:


src/plugins/lua/antivirus.lua
```diff
@@ -162,18 +162,24 @@ local function sophos_config(opts)
   return nil
 end
 
+local function message_too_large(task, rule)
+  local max_size = tonumber(rule['max_size'])
+  if not max_size then return end
+  if task:get_size() > max_size then return true end
```
vstakhov: That's a bit inconsistent: you return 3 values from a function: `true`, `false` and `nil`. I think it's not a good idea.

fatalbanana: I've made it consistent (and also inverted this function's logic).

src/plugins/lua/antivirus.lua
```diff
@@ -162,18 +162,24 @@ local function sophos_config(opts)
   return nil
 end
 
+local function message_not_too_large(task, rule)
+  local max_size = tonumber(rule['max_size'])
+  if not max_size then return true end
```
vstakhov: Shouldn't you return `false` here?

fatalbanana: I've inverted logic (so `true` means it's scanned now). :) If we don't have `max_size` setting (or it's not a number) we want to just scan.

PR #1072, opened 2016-10-25
Change logic for MX-check plugin

AlexeySa:
Old version: FROM <one@test.vov.ru> realy check "vov.ru"
New version: FROM <one@test.vov.ru> realy check "test.vov.ru"


vstakhov:
Are you sure that this logic is preferrable? @moisseev?

fatalbanana:
Seems preferable to me FWIW: better to test MX of the actual address seen than whichever TLD it belongs to.

vstakhov:
Ok, LGTM then.

PR #1086, opened 2016-11-01
[Feature] Add subject type to multimap

andfgon:
Currently the header type of multimap does not return the value decoded if it is base64.
The method task:get_header_full() always makes the value fields and decoded with the subject content in base64, not decoded.

The change provide the possibility of using the subject type in the multimap.conf configuration file, using the method task:get_header('subject') to return the subject decoded.

Multimap configuration example:

```
[root@mail ~]# cat /usr/local/etc/rspamd/local.d/multimap.conf 
BLACKLIST_SUBJECT_REGEX {
        type = "subject";
        regexp = true;
        map = "/usr/local/etc/rspamd/myrules/blacklist-subject.map";
}
```

```
[root@mail ~]# cat /usr/local/etc/rspamd/myrules/blacklist-subject.map
/.*Total\sSales.*/i
```

### DIFF
--- a/src/plugins/lua/multimap.lua
+++ b/src/plugins/lua/multimap.lua
@@ -38,6 +38,9 @@ local value_types = {
   header = {
     get_value = function(val) return val end,
   },
+  subject = {
+    get_value = function(val) return val end,
+  },
   rcpt = {
     get_value = function(val) return val end,
   },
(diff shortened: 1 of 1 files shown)


fatalbanana:
Actually, what multimap is matching is supposed to be a decoded value but API returns the wrong thing: https://github.com/vstakhov/rspamd/pull/1088

vstakhov:
Right, thank you Andrew. I'll thus close these requests.

andfgon:
Perfect, I still suggest adding the subject type as a new feature, making it easy to create rules by subject.

PR #1112, opened 2016-11-08
[Feature] Module to push messages/metadata to redis pubsub

fatalbanana:
Maybe this is worth having around...

Possible use-cases include quarantines and feedback loops.


vstakhov:
The only thing I doubt about is the name of the plugin. `pubsub` sounds quite confusing. Perhaps, something like `metadata_export` would be a more descriptive name?

PR #1125, opened 2016-11-13
[Minor] Further simplify conditionals

moisseev:
@AlexeySa I don't understand why strict check`do_greylisting_required ~= "1"` is necessary. More over, I think this [condition](https://github.com/vstakhov/rspamd/blob/4cfb22716f09aa901aaa046fffa5d8c202edf9c1/src/plugins/lua/greylist.lua#L279) will never match. Did I miss something?
So this code
```lua
  if action == 'reject' then return end
  if do_greylisting_required and do_greylisting_required ~= "1" then
    return
  else
    if action == 'no action' then return end
  end
```
could be repaced with
```lua
  if action == 'reject' then return end
  if not do_greylisting_required and action == 'no action' then return end
```
or
```lua
  if action == 'reject' or
      not do_greylisting_required and action == 'no action' then
    return
  end
```

### DIFF
--- a/src/plugins/lua/greylist.lua
+++ b/src/plugins/lua/greylist.lua
@@ -275,11 +275,9 @@ local function greylist_set(task)
   end
 
   local action = task:get_metric_action('default')
-  if action == 'reject' then return end
-  if do_greylisting_required and do_greylisting_required ~= "1" then
+  if action == 'reject' or
+      not do_greylisting_required and action == 'no action' then
     return
-  else
-    if action == 'no action' then return end
   end
   local body_key = data_key(task)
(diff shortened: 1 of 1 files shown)


AlexeySa:
@moisseev, yhank you. The simpler the better :).
I use this feature in my plugins:
```
task:get_mempool():set_variable("grey_greylisted_required", "1")
```

moisseev:
@AlexeySa I believe exact value doesn't matter in this case. So merely check if "grey_greylisted_required" defined or not should be enough.

AlexeySa:
Yeah, right

PR #1127, opened 2016-11-14
Update freemail hosts

AlexeySa:
From https://github.com/willwhite/freemail/blob/master/data/free.txt


vstakhov:
@AlexeySa: I've added mirrors of these maps, so I believe there is no need to include this file into Rspamd distribution any longer

vstakhov:
Freemail and disposable domains are now mirrored at rspamd.com:

https://rspamd.com/freemail/free.txt.zst
https://rspamd.com/freemail/disposable.txt.zst

They can be used for `add_map` code.

AlexeySa:
Ok

PR #1128, opened 2016-11-14
[Minor] Update sample whitelist for greylist

AlexeySa:


vstakhov:
I'll also remove `freemail.inc` from the source tree.

PR #1156, opened 2016-11-21
Rules updates

smfreegard:


AlexeySa:
@smfreegard , thenk you!

I think we should make exceptions for local and auth users.

rules/misc.lua
```diff
@@ -395,3 +395,585 @@ rspamd_config.MISSING_FROM = {
...
+    local to = task:get_recipients(2)
+    if not (to and to[1]) then return false end
+    -- Check if FROM == TO
+    if (#to == 1 and to[1].addr:lower() == from[1].addr:lower()) then
+      task:insert_result('TO_EQ_FROM', 1.0)
+    elseif (#to == 1 and to[1].domain:lower() == from[1].domain:lower()) then
+      task:insert_result('TO_DOM_EQ_FROM_DOM', 1.0)
+    end
+  end
+}
```
vstakhov: I'd suggest to use `rspamd_config:register_symbols` instead of fake rules.

rules/misc.lua
```diff
@@ -395,3 +395,585 @@ rspamd_config.MISSING_FROM = {
...
+
+rspamd_config.TO_DOM_EQ_FROM_DOM = {
+  callback = function()
+    -- Set by CHECK_FROM
+  end,
+  description = 'To domain is the same as the From domain',
+  score = 0.0
+}
+
```
vstakhov: This all looks very bad and have some performance implications...

rules/misc.lua
```diff
@@ -395,3 +395,585 @@ rspamd_config.MISSING_FROM = {
...
+    end
+
+    -- See if header recipients match envelope recipients
+    if (to_match_envrcpt == #to) then
+      task:insert_result('TO_MATCH_ENVRCPT_ALL', 1.0)
+    elseif (to_match_envrcpt > 0) then
+      task:insert_result('TO_MATCH_ENVRCPT_SOME', 1.0)
+    end
+  end
+}
```
vstakhov: Same as previous: `rspamd_config:register_symbols` is more appropriate.

rules/misc.lua
```diff
@@ -395,3 +395,585 @@ rspamd_config.MISSING_FROM = {
...
+  description = 'Message-ID RHS matches From domain',
+  score = 1.0
+}
+
+rspamd_config.CHECK_RECEIVED = {
+  callback = function (task)
+    local received = task:get_received_headers()
+    task:insert_result('RCVD_COUNT_' .. #received, 1.0)
+  end
+}
```
vstakhov: We have `mid` check plugin, so I'd suggest to add these rules to the plugin.

rules/misc.lua
```diff
@@ -395,3 +395,585 @@ rspamd_config.MISSING_FROM = {
...
+  score = 0.0
+}
+
+rspamd_config.REPLYTO_DN_EQ_FROM_DN = {
+  callback = function ()
+    -- Set by CHECK_REPLYTO
+  end,
+  description = 'Reply-To display name matches From',
+  score = 0.0
+}
```
vstakhov: Same comment.

rules/misc.lua
```diff
@@ -395,3 +395,585 @@ rspamd_config.MISSING_FROM = {
...
+
+rspamd_config.MIME_NA_MISSING_HTML = {
+  callback = function ()
+    -- Set by CHECK_MIME
+  end,
+  description = 'MIME multipart/alternative missing text/html part',
+  score = 2.0
+}
+
```
vstakhov: Same here.

rules/misc.lua
```diff
@@ -395,3 +395,585 @@ rspamd_config.MISSING_FROM = {
...
+-- Used to be called IS_LIST
+rspamd_config.PREVIOUSLY_DELIVERED = {
+  callback = function(task)
+    if not task:has_recipients(2) then return false end
+    local to = task:get_recipients(2)
+    local rcvds = task:get_header_full('Received')
```
vstakhov: You can get this information parsed when using `task:get_received`. Extra searches seems to be unneeded.

smfreegard:
@vstakhov - updated as requested.

@AlexeySa 
> I think we should make exceptions for local and auth users.

For which rules?

> MID check gives a lot of false positives.

Which particular symbols?  I don't see anything particularly dangerous in there.

rules/misc.lua
smfreegard: Damn - I forgot this one.   Will update again...

AlexeySa:
Many good mails has bad mid-headers. Sample: @rcron.myorderbox.aus-tx.col, @localhost, @ru...
Exceptions for local/auth users: helo, subject, message-id checks...

Could you arrange the code as a plugin? The user should be able to disable your rules.

PR #1157, opened 2016-11-21
Add freemail/disposable maps

smfreegard:


vstakhov:
What about scores for these rules?

xpunkt:
possible to test CC: aswell for both maps files ?

smfreegard:
@vstakhov 
> What about scores for these rules?

They're designed as informational rules and for use in composites rather than for outright scoring.   I would need to run these through a corpus check to work of what percentages they hit before I would score them singly.

@xpunkt 
> possible to test CC: aswell for both maps files ?

Sure - I'll update.

PR #1181, opened 2016-11-25
Debug logging

fatalbanana:


vstakhov:
I think about adding slightly different function: `logger.debugm('module', task|config|pool|string_tag, 'fmt string', ...)` What do you think?

fatalbanana:
SGTM.

vstakhov:
Now you can use it :)

PR #1196, opened 2016-11-28
cmake: Use Find Modules to find Luajit and Lua

cryptomilk:
I was not able to build rspamd on openSUSE Leap 42.2 or Tumbleweed. The cmake code is really strange so I simply used Find Modules and it works now.

You should use FindModules for the rest too. I wrote a simple script to generate them. See

https://xor.cryptomilk.org/scripts/cmake_create_findpackage

Good cmake examples are https://cmocka.org or https://libssh.org


vstakhov:
This change will likely break lua-5.2 and lua-5.3 compatibility. I haven't seen any issues with the existing code btw.

cryptomilk:
Can you be more precise what you think will break?

I will try with lua 5.3

vstakhov:
You search for lua-5.1 only from what I see. And I really don't see any improvements over the existing approach: I've played with FindModules in the past and found it extremely stupid and not suitable for all things I want to have in Rspamd.

cryptomilk:
```find_package(Lua 5.1 REQUIRED)```

This menas the minimum required version is Lua 5.1!

```cmake --help-module find_package | less ```

The current code does not work on openSUSE 42.2 and Leap and I'm trying to package it there. In CMake Find Modules or cmake-configs are used to find include directories and libraries.

**Comment by cryptomilk** [comment]
This patchset works correctly on Fedora 25 with Lua 5.3.

**Comment by vstakhov** [comment]
Ok, but now the problem is that your change breaks my build scripts... I use static luajit in `${RSPAMD_SEARCH_PATH}`. Another issue is that you can have multiple lua versions in your system (e.g. in FreeBSD) and I have completely no idea which version will be chosen.

**Comment by cryptomilk** [comment]
Ok, so we need HINTS to simplify that. I will add set to FindLuaJIT.cmake.

HINTS are used to simplify pointing cmake to the location to find a library. If you want a special version, you can always set them already at the command line:

```cmake -DLUA_INCLUDE_DIR=/opt/lua555/include -DLUA_LIBRARY=/opt/lua555/lib/liblua555.so /path/to/source```

If you use HINTS it gets simpler:

```cmake -DLUAJIT_ROOT_DIR=/opt/luajit```

A find module normally looks always for the latest version and if not find look for a lower version:

lua 5.3 -> 5.2 -> 5.1 -> 5.0

File: cmake/Modules/FindLuaJIT.cmake [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+if (UNIX)
+  find_package(PkgConfig)
+  if (PKG_CONFIG_FOUND)
+    pkg_check_modules(_LUAJIT luajit)
+  endif (PKG_CONFIG_FOUND)
+endif (UNIX)
+
+set(_LUAJIT_ROOT_HINTS
+    ${RSPAMD_SEARCH_PATH}
+    ${LUJIT_ROOT_DIR}
```
vstakhov: LUJIT looks like a misprint.

cryptomilk:
Ups. Fixed :-)

vstakhov:
I'll revert this PR because it breaks OSX build.

cryptomilk:
If you can tell me what exactly broke I can fix it.

vstakhov:
It doesn't link with the proper library: it finds headers from lua-5.3 and links with luajit. Anyway, I won't like to change this code I'm afraid. I agree that your version is cleaner in terms of cmake, but I've spent a lot of time to make it working on different OSes (e.g. FreeBSD, OpenBSD, NetBSD, OSX, Solaris and others) and different lua/luajit combinations. I'd suggest to fix it somehow for your Suse setup because I won't likely change it significantly unless it is proven to work on non-Linux OSes.

cryptomilk:
Is it a header or a linking problem now? Is pkg-config incorrect on OSX?

How did you get lua on OSX through some open source package management? I have a OSX machine here so I can try it.

In the long run it is much cleaner and easier to use Find Modules.

vstakhov:
It is linking problem. The problem is that with `ENABLE_LUAJIT=ON`, Rspamd should just ignore all but luajit. OSX has `macports` which places files into `/opt/local` folder by default. Pkg-config works:

~~~
% pkg-config --libs lua
-L/opt/local/lib -llua -lm
% pkg-config --libs lua-5.1
-L/opt/local/lib/lua-5.1 -llua-5.1 -lm
% pkg-config --libs luajit
-L/opt/local/lib -lluajit-5.1
~~~

I agree that modules code is better but I want to keep builds stable at least for this time (I'm going to release 1.4.1 very soon).

PR #1197, opened 2016-11-28
Revert "cmake: Use Find Modules to find Luajit and Lua"

vstakhov:
Reverts vstakhov/rspamd#1196


cryptomilk:
What is the issue on OSX, so that I can address it?

PR #1201, opened 2016-11-29
Fix typo when calling rspamd_redis_make_request for write operation.

andrejzverev:


andrejzverev:
FYI, i did check for all other modules, and seems this is only place to fix.

PR #1205, opened 2016-11-29
Adjust scores and add new rules

smfreegard:


ntninja:
@smfreegard

Hallo!

Do you happen to remember why you added this rule:

```
reconf['X_PHPOS_FAKE'] = {
  re = 'X-PHP-Originating-Script=/^\\d{7}:/Hi',
  description = 'Fake X-PHP-Originating-Script header',
  score = 3.0,
  group = 'headers'
}
```

It’s causing issues on modern OS’s: #4091

fatalbanana:
> It’s causing issues on modern OS’s: #4091

The idea is that such a number is too large to be a valid UID (but it mightn't hold true).

ntninja:
Yes, it is no longer true given that at least Linux has 32-bit UIDs nowadays (it’s just a matter of configuration whether they are used or not – particularly NFS and containerization are common users).

ntninja:
May I submit a patch to either remove or change the limit to be above `2**32 - 1` instead?

ntninja:
Opened PR: #5745

Issue #1230, opened 2016-12-06
Webui / Focus on password field


madorin:
On first connect, if user is not authenticated do an inpuit_password.focus() to avoid mouse gestures ;)

madorin:
Wow, it was quick, but i think better to put it in dialog.show() callback, because it is called async.
Something like:

              $(dialog).show(function(){
                  if ($(#connectHost).is(":visible"))
                    $('#connectHost').focus();
                  else
                    $('#connectPassword').focus();
              });
 
Code is not tested.

PR #1231, opened 2016-12-06
[WebUI] Set focus to password field (#1230)

fatalbanana:
Reported-by: @madorin


madorin:
Wow, it was quick :)

By the way, i think it will be better to put focus in a callback, because show is async, and it may take a while on slow browsers, like:

$(dialog).show(function(){
    $("connectPassword", dialog).focus();
});

Code is not tested.

PR #1236, opened 2016-12-07
Make optional dependency on hiredis really optional

vaygr:
I've been running my rspamd installation on OpenBSD with this patch since `1.4.0` (before `1.4.0` building without hiredis support worked just fine).

I never used Redis on my box and never needed to. As current configuration options allow to turn off compilation with hiredis support and as [documentation outlines](https://rspamd.com/doc/quickstart.html#caching-setup) it's used for caching, I'm creating a PR to fix this.

Currently I'm running rspamd with the following modules:

- forged_recipients
- maillist
- chartable
- once_received
- phishing
- spf
- dkim
- regexp
- hfilter
- whitelist
- mime_types

none of which requires Redis support.


vstakhov:
I'll merge this change for now but please bear in mind that I can break it again in future. In fact, I do not test this configuration since I'm not convinced why do you need Rspamd without hiredis.

src/libserver/fuzzy_backend.c
```diff
@@ -85,8 +85,9 @@ static const struct rspamd_fuzzy_backend_subr fuzzy_subrs[] = {
 		.id = rspamd_fuzzy_backend_id_sqlite,
 		.periodic = rspamd_fuzzy_backend_expire_sqlite,
 		.close = rspamd_fuzzy_backend_close_sqlite,
-	},
-	[RSPAMD_FUZZY_BACKEND_REDIS] = {
+	}
+#ifdef ENABLE_HIREDIS
```
vstakhov: There is no such macro JFYI...

vaygr: `WITH_HIREDIS`, right, thanks a lot

PR #1241, opened 2016-12-07
respect -u option in cmd line

andrejzverev:


fatalbanana:
`-u` belongs to `--user`.

fatalbanana:
And `--ucl` is just a duplicate of `--json`. :)

andrejzverev:
I don't see any ``user`` here and i doubt it will be.
Since we are declaring to -u available it don't really matter if this just a duplicate, really.
``rspamadm control -u stat`` must just work even if this is by default. Another way is remove ``-u`` from usage output. Well, something like that.   

```
./src/rspamadm/rspamadm control --help
Usage:
  rspamadm [OPTION...] control - manage rspamd main control interface

Summary:
  Rspamd administration utility version 1.5.0
  Release id: 8ebc1e5

Help Options:
  -h, --help        Show help options

Application Options:
  -j, --json        Output json
  -c, --compact     Output compacted
  -u, --ucl         Output ucl (default)
  -s, --socket      Use the following socket path
  -t, --timeout     Set IO timeout (1s by default)
```

andrejzverev:
this is only cosmetic fix ;-) c'mon

fatalbanana:
Sorry I thought this was rspamc....

PR #1245, opened 2016-12-08
[Feature] Add rmilter_headers module (Fixes #1227)

### LINKED ISSUE #1227: Module for easily adding/removing headers
Idea is to provide some options for adding/removing headers that doesn't require one to write a script. Not entirely sure about how this should look (some library of functions which can be combined freely in configuration I suppose).

fatalbanana:
```
rmilter_headers {
        use = "x-spamd-bar";
        routines {
                x-spamd-bar {
                        negsymbol = "";
                }
        }
}
```


vstakhov:
What about authentication results?

fatalbanana:
That's a possibility. :)

PR #1255, opened 2016-12-12
[WebUI] Fix read only

AnnaStakhova:


moisseev:
We need to disable history reset button as well.

PR #1264, opened 2016-12-14
[Fix] Do not set pre-result and update records for no `Queue-ID` messages

moisseev:


vstakhov:
Can you please explain what is the purpose of this change?

fatalbanana:
Idea is for greylisting module not to inhibit testing with rspamc.

moisseev:
Yes, this is for messages that did not come from MTA (scanned with `rspamc` or WebUI). Obviously, greylisting records should not be updated in this case as well.

PR #1269, opened 2016-12-19
[Feature] Support DKIM signing in Lua plugins

fatalbanana:


vstakhov:
But that's actually not a support of DKIM signing in lua plugins - it is `sign_condition` that lives in `rspamd_plugins.dkim.sign` isn't it?

fatalbanana:
Humm, `rspamd_plugins.dkim.sign` shares nothing with `sign_condition` though it duplicates much of the code that's used there. Unlike `sign_condition` it can be called from a plugin. :)

PR #1271, opened 2016-12-20
[Feature] Add preliminary version of DKIM signing module

fatalbanana:
How does this look to you? :) It's not entirely like rmilter but it's close. :)


vstakhov:
It looks good to me, thanks!

Issue #1279, opened 2016-12-22
DATE_IN_FUTURE FP


moisseev:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [x] Other bug
* [ ] Feature
* [ ] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
1.5.0-git.b75412f

### Additional information:
Rspamd adds the `DATE_IN_FUTURE(4.00)[]` symbol for correct time stamp if `local name` for the local time zone is specified in the `Date:` header.

- `Date: Thu, 22 Dec 2016 14:11:48 +0300 (MSK)` gives FP
- `Date: Thu, 22 Dec 2016 14:11:48 +0300` is fine

fatalbanana:
I can't reproduce that.

```From: "barf@google.com" <barf@rspamd.tk>
Date: Thu, 22 Dec 2016 13:42:48 +0300 (BLAH)
Content-type: text/plain

test```

**Comment by vstakhov** [comment]
I cannot reproduce it as well. I've also printed that date from Lua and I don't see any difference between date with `(MSK)` and without.

**Comment by moisseev** [comment]
To reproduce the issue the timestamp should be close enough to current time. I guess, it is a problem with timezone parsing, like it uses `0` instead of `+3`.
```
mx# date
четверг, 22 декабря 2016 г. 17:03:00 (MSK)
mx# rspamc -p
Date: Thu, 22 Dec 2016 13:42:48 +0300 (MSK)
Content-type: text/plain

Results for file: stdin (7.357 seconds)
[Metric: default]
Spam: true
Score: 12.90 / 15.00
Action: soft reject
Symbol: DMARC_NA (0.00)[No From header]
Symbol: GREYLIST (0.00)[Thu, 22 Dec 2016 14:08:14 GMT, greylisted, body]
Symbol: MIME_GOOD (-0.10)[text/plain]
Symbol: MISSING_FROM (2.00)
Symbol: MISSING_MID (2.50)
Symbol: MISSING_MIME_VERSION (2.00)
Symbol: MISSING_SUBJECT (2.00)
Symbol: MISSING_TO (2.00)
Symbol: RCPT_COUNT_0 (0.00)
[...]

PR #1281, opened 2016-12-23
[Fix] Date header timezone adjustments (#1279)

fatalbanana:


src/libutil/util.c
```diff
@@ -2648,7 +2648,7 @@ rspamd_tm_to_time (const struct tm *tm, glong tz)
 	result += tm->tm_sec;
 
 	/* Now apply tz offset */
-	result -= offset;
```
vstakhov: Are you sure in this change?

fatalbanana: Yep. Positive time adjustments were actually subtracted, maybe negative adjustments would not have worked correctly either. Both cases seem to be handled correctly now.

PR #1302, opened 2017-01-04
Rspamd 1.4: backport memory leak fix/other fixes

fatalbanana:


vstakhov:
You need two commits here I'm afraid.

vstakhov:
https://github.com/vstakhov/rspamd/commit/16c8c89c354a62aae901c63a43688a164b4d9ea0 - this one. Otherwise it will just blow up.

PR #1318, opened 2017-01-07
[WebUI] Add summary to the Throughput tab

moisseev:


dehnli:
rspamd-webui-troughput-pie-bug

It occurs always when pressing left-side rrd-pie entry

PR #1368, opened 2017-01-24
Rework ratelimit module changes

fatalbanana:
As discussed

### DIFF
--- a/src/plugins/lua/ratelimit.lua
+++ b/src/plugins/lua/ratelimit.lua
@@ -539,7 +539,7 @@ local function parse_string_limit(lim)
   local t = lpeg.match(grammar.limit, lim)
 
   if t and t[1] and t[2] and t[2] ~= 0 then
-    return t[1] / t[2]
+    return t[1] / t[2], t[1]
   end
 
   rspamd_logger.errx(rspamd_config, 'bad limit: %s', lim)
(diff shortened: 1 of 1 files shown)


src/plugins/lua/ratelimit.lua
```diff
@@ -560,14 +560,11 @@ if opts then
...
-          settings[t] = lim
-          local plim = parse_string_limit(lim[2])
-          if plim then
-            settings[t][2] = plim
-          end
+        settings[t] = lim
+      elseif type(lim) == 'string' then
+        local plim = parse_string_limit(lim)
+        if plim then
+          settings[t] = {0, plim, 1}
```
vstakhov: Erm, that doesn't something I've meant. Why 0? The limit of bucket should be the first value (before slash).

fatalbanana: If we have `some_ratelimit = "2 / 1m";` and "2" is the size of the bucket then what is leak rate? Is it supposed to be 1 per minute? Code seems to expect that this in fact part of the leak rate and not bucket size: which I understood we'd want to always be zero, rather than specified in configuration at all.

vstakhov: The bucket size is 2 and leak rate is 2 / 60. So it will be empty in two minutes allowing 2 messages to be accepted.

fatalbanana: That sounds reasonable. :)

PR #1388, opened 2017-01-31
rspamadm grep

fatalbanana:
- Read from stdin & funny file extensions
- Avoid undefined behaviour
- Try print and clear matches as soon as possible


fatalbanana:
Oh, this should become `rspamadm grep` I suppose.

PR #1397, opened 2017-02-05
[WebUI] Add server selector to navbar

moisseev:


vstakhov:
That's something I wanted to do as well. Thank you!

PR #1499, opened 2017-03-08
[Fix] Fix log line duplication in `rspamadm grep`

fatalbanana:
- Reported by: @croessner


fatalbanana:
Actually no, hold on... it's still bad. :)

PR #1501, opened 2017-03-08
Add task:get_symbols_all() ; apply it in metadata_exporter

fatalbanana:


src/lua/lua_task.c
```diff
@@ -2811,6 +2828,43 @@ lua_task_get_symbol (lua_State *L)
...
+	gpointer k;
+	gboolean found = FALSE;
+	gint i = 1;
+
+	if (task) {
+		metric = task->cfg->default_metric;
+		mres = g_hash_table_lookup (task->results, DEFAULT_METRIC);
+		if (mres) {
+			found = TRUE;
+			lua_newtable (L);
```
vstakhov: It is better to use lua_createtable here since we know the upper bound of the table.

src/lua/lua_task.c
```diff
@@ -2811,6 +2828,43 @@ lua_task_get_symbol (lua_State *L)
...
+			g_hash_table_iter_init (&it, mres->symbols);
+			while (g_hash_table_iter_next (&it, &k, NULL)) {
+				lua_push_symbol_result (L, task, metric, k, FALSE, TRUE);
```
vstakhov: I'd recommend to split `lua_push_symbol_result` into two parts: one with hash table lookup and one that accepts the complete struct rspamd_symbol. This will optimise execution slightly.

src/lua/lua_task.c
```diff
@@ -2708,53 +2715,75 @@ static inline gboolean
...
+	const gboolean add_metric,
+	const gboolean add_name)
 {
 	struct rspamd_metric_result *metric_res;
 	struct rspamd_symbol_result *s;
-	gint j;
+	gint j, e;
+
+	if (!symbol_result) {
+		metric_res = g_hash_table_lookup (task->results, metric->name);
```
vstakhov: What if `metric_res` is NULL?

PR #1512, opened 2017-03-10
[Test] Fix F-Prot/ClamAV test stability

fatalbanana:


mischasan:
Pardon me, but the two bugs in acism are now fixed. Are you still using it?

vstakhov:
@mischasan Thank you very much for your update. I'm working on porting it to Rspamd so far. In the meantime, I have two questions:

1. https://github.com/mischasan/aho-corasick/blob/master/acism_create.c#L84 - I don't understand why do you use `nhash` here - it is not initialised anywhere.
2. What do you think about relaxing LGPL requirements: embedding it to Rspamd would have some benefits from link-time-optimizations which are now enabled for Rspamd packages.

vstakhov:
The (1) has introduced by https://github.com/mischasan/aho-corasick/commit/3c04651d5c53d3e9f570d943d2fb682a21a92752

vstakhov:
So my change is the following: https://github.com/vstakhov/rspamd/commit/8164a21476e2f343a761c30f05c065199ea092ba

mischasan:
Thanks for the catch. I'm having a cert problem with git push, and made
manual mods using the github editor.
There were two lines missing from the file on github:

     for (tp = troot + nnodes, nhash = 0; --tp > troot;)
         nhash += tp->match && tp->child;

I have updated the file to match, and will post when I have verified all
files against my local git repo.

I am happy to relax the license for rspamd specifically. GPL for me is a
way to find out who uses (or wants to use) the package --- e.g. rspamd.
I have no other interest in licensing (beyond BSD self-protection). I see
that rspamd uses the Apache license. From your experience, is there a
simple way to handle this?

fatalbanana:
How about [Mozilla Public License](https://www.mozilla.org/en-US/MPL/2.0/FAQ/)?

mischasan:
I'm going to ask our local legal about the easiest way to handle this.
I had a problem with another piece of SSE2 code that I relabelled for
someone with a BSD license.
I found out two years later that it was being widely used elsewhere without
my knowledge ... defeating my only reason for GPL in the first place.
Because of the LGPL, I've been able to follow its course in cancer (genome)
research and ANN work (which somehow were less affected by that bug!?).
LGPL is a heavyweight way of doing this; ANYTHING else would be preferrable.

I recognize the big effort rspamd is, with lots of contributors. FWIW my
prev job included work on Sophos' spam filter.
So I have a strong motivation to enable your use of the source Apache-style.
I'm hunting for a way to preserve the "paper trail".

Issue #1521, opened 2017-03-14
Multiple DKIM Signing Keys and Domains not working


jnhmn:
I have following dking_signing.conf in the local.d folder:
```
   If false, messages with empty envelope from are not signed                                                                                                 
  allow_envfrom_empty = true;
  # If true, envelope/header domain mismatch is ignored
  allow_hdrfrom_mismatch = true;
  # If true, multiple from headers are allowed (but only first is used)
  allow_hdrfrom_multiple = false;
  # If true, username does not need to contain matching domain
  allow_username_mismatch = true;
  # If false, messages from authenticated users are not selected for signing
  auth_only = true;
  # Default path to key, can include '$domain' and '$selector' variables
  #path = "/etc/mail/$domain/$selector.private";
  # Default selector to use 
  selector = "default";
  # If false, messages from local networks are not selected for signing
  sign_local = true;
  # Symbol to add when message is signed
  symbol = "DKIM_SIGNED";
  # Whether to fallback to global config
  try_fallback = true;
  # Domain to use for DKIM signing: can be "header" or "envelope"
  use_domain = "envelope";
  # Whether to normalise domains to eSLD
  use_esld = false;
  # Whether to get keys from Redis
  use_redis = false;
  # Hash for DKIM keys in Redis
  hash_key = "DKIM_KEYS";

  # Domain specific settings
    domain {
      lists.ex.de {
        path = "/etc/mail/lists.ex.de/default.private";
        selector = "default";
      }   
      ex.de {
        path = "/etc/mail/ex.de/default.private";
        selector = "default";
      }   
      sub.sub.tld.de {
        path = "/etc/mail/sub.sub.tld.de/default.private";
        selector = "default";
      }   
    } 
```

The server still tries to find the signing keys in ` /var/lib/rspamd/dkim/$domain.$selector.key` and ignores the domain specific configurations

fatalbanana:
Should be fixed in this PR: https://github.com/vstakhov/rspamd/pull/1522

PR #1522, opened 2017-03-14
[Fix] DKIM signing: use domain-specific signing key (#1521)

fatalbanana:
Reported by: @jnhmn


beckzg:
Is this fix included in version 1.8.3? I'm using this version and the problem still persists, it's looking for key in /var/db/rspamd/dkim/$domain.$selector.key despite I'm using per domain config.

PR #1548, opened 2017-03-22
[Minor] Use SMTP recipients for SPOOF_REPLYTO rule

fatalbanana:
This avoids some possible trouble with mailing lists.


vstakhov:
LGTM

PR #1550, opened 2017-03-23
[WebUI] Fixes overlapping, fixes updates from empty to non-empty table, move symbols data processing to function

andryyy:


vstakhov:
Great job, thank you!

PR #1570, opened 2017-03-30
added sampled_out dmarc report

cpragadeesh:
Also applied next policy by severity in case of sampled_out emails as per [RFC7489](https://tools.ietf.org/html/rfc7489#section-6.6.4).

### DIFF
--- a/src/plugins/lua/dmarc.lua
+++ b/src/plugins/lua/dmarc.lua
@@ -73,14 +73,14 @@ end
 
 local dmarc_grammar = gen_dmarc_grammar()
 
-local function dmarc_report(task, spf_ok, dkim_ok, disposition)
+local function dmarc_report(task, spf_ok, dkim_ok, disposition, sampled_out)
   local ip = task:get_from_ip()
   if not ip:is_valid() then
     return nil
   end
-  local res = string.format('%d,%s,%s,%s,%s', task:get_date(0),
+  local res = string.format('%d,%s,%s,%s,%s,%s', task:get_date(0),
     ip:to_string(), tostring(spf_ok), tostring(dkim_ok),
-    disposition)
+    disposition, (sampled_out and 'sampled_out' or ''))
 
   return res
 end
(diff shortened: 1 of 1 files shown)


src/plugins/lua/dmarc.lua
```diff
@@ -321,11 +321,16 @@ local function dmarc_callback(task)
         if not pct or pct == 100 or (math.random(100) <= pct) then
           task:insert_result(dmarc_symbols['quarantine'], res, lookup_domain .. ' : ' .. reason_str, dmarc_policy)
           disposition = "quarantine"
+        else
+          task:insert_result(dmarc_symbols['softfail'], res, lookup_domain .. ' : ' .. reason_str, "none", "sampled_out")
```
fatalbanana: You'll want to set `disposition = 'none'` here

fatalbanana:
That's a good start, thanks, but this information is still missing from DMARC reporting information in Redis.

src/plugins/lua/dmarc.lua
cpragadeesh: disposition is set to 'none' initially here:
`local disposition = 'none'`

Should I set it to 'none' again regardless?

fatalbanana: No, sorry. :) It should be fine in fact.

fatalbanana: Oh, "none" here should be dmarc_policy.

src/plugins/lua/dmarc.lua
```diff
@@ -321,11 +321,16 @@ local function dmarc_callback(task)
...
         end
       elseif dmarc_policy == 'reject' then
         if not pct or pct == 100 or (math.random(100) <= pct) then
           task:insert_result(dmarc_symbols['reject'], res, lookup_domain .. ' : ' .. reason_str, dmarc_policy)
           disposition = "reject"
+        else
+          task:insert_result(dmarc_symbols['quarantine'], res, lookup_domain .. ' : ' .. reason_str, "quarantine", "sampled_out")
```
fatalbanana: "quarantine" should be dmarc_policy.

cpragadeesh:
@fatalbanana Hope I got it right this time :)

PR #1586, opened 2017-04-11
New rules

smfreegard:
CTYPE_MIXED_BOGUS
HTTP_TO_HTTPS
HTTP_TO_IP
MIME_BASE64_TEXT


smfreegard:
Ok - updated.   

I've zero'd out the score for MIME_BASE64_TEXT as it might be useful in composites or improved in the future.

I've also added early loop exits into HTTP_TO_HTTPS and HTTP_TO_IP.

vstakhov:
Cool, thank you, Steve!

PR #1601, opened 2017-04-18
[Feature] UCL: register parser variables from Lua

fatalbanana:


contrib/libucl/lua_ucl.c
```diff
@@ -627,6 +627,76 @@ lua_ucl_parser_parse_file (lua_State *L)
...
+	value = luaL_checkstring (L, 3);
+
+	if (parser != NULL && name != NULL && value != NULL) {
+		ucl_parser_register_variable (parser, name, value);
+		lua_pushboolean (L, true);
+		ret = 1;
+	}
+	else {
+		lua_pushboolean (L, false);
+		lua_pushstring (L, "invalid arguments");
```
vstakhov: It is better to return fatal error by `luaL_error` here.

contrib/libucl/lua_ucl.c
```diff
@@ -627,6 +627,76 @@ lua_ucl_parser_parse_file (lua_State *L)
...
+{
+	struct ucl_parser *parser;
+	const char *name, *value;
+	int ret = 2;
+
+	parser = lua_ucl_parser_get (L, 1);
+
+	if (parser != NULL && lua_type (L, 2) == LUA_TTABLE) {
+		for (lua_pushnil (L); lua_next (L, 2); lua_pop (L, 1)) {
+			name = luaL_checkstring (L, -2);
```
vstakhov: No, that won't work. You cannot just touch keys when iterating over a table (tostring will corrupt it). Use the following:

```c
lua_pushvalue (L, -2);
name = luaL_checkstring (L, -1);
value = luaL_checkstring (L, -2);
ucl_parser_register_variable (parser, name, value);
lua_pop (L, 1);
```

contrib/libucl/lua_ucl.c
```diff
@@ -627,6 +627,76 @@ lua_ucl_parser_parse_file (lua_State *L)
...
+			value = luaL_checkstring (L, -1);
+			ucl_parser_register_variable (parser, name, value);
+		}
+
+		lua_pushboolean (L, true);
+		ret = 1;
+	}
+	else {
+		lua_pushboolean (L, false);
+		lua_pushstring (L, "invalid arguments");
```
vstakhov: Same thing: use `luaL_error`

vstakhov:
LGTM, merging

PR #1607, opened 2017-04-26
[Conf] Remove ratelimits from default configuration

fatalbanana:
I don't think these should feature in default configuration as we can't very well suggest reasonable values for them and they're likely to take users by surprise.


vstakhov:
That looks reasonable.

PR #1615, opened 2017-05-05
[Minor] Fix result parsing for SAVAPI

croessner:


src/plugins/lua/antivirus.lua
```diff
@@ -23,46 +23,6 @@ local redis_params
 
 local N = "antivirus"
 
-if confighelp then
```
fatalbanana: Why is this removed?

croessner: That is what I tried to tell you in IRC ;-) You must please cherry-pick :-) The 1.5.7 release does not know about confighelp. That came later.

fatalbanana: Commits applied to the master branch likely pick apply cleanly to rspamd-1.5. Applying your changes to rspamd-1.5 and then reverse cherry-picking is less likely to work well (but could). Copying files wholesale between branches won't work.

croessner: Is there something that I can do right now, or can you doe it? If you look at my changes, they are really very little. In fact I kicked the trim() function and changed some lines concerning the 310 parsing. This is much cleaner than before. And hopefully the parsing is fixed now and forever ;-)

fatalbanana: A well-formed PR would be most helpful. :)

PR #1619, opened 2017-05-08
Read DKIM selector from Redis

andryyy:


src/plugins/lua/dkim_signing.lua
```diff
@@ -185,6 +186,25 @@ local function dkim_signing_cb(task)
   end
 
   if settings.use_redis then
+    if settings.selector_prefix then
+      local function redis_key_cb(err, data)
+        if err or type(data) ~= 'string' then
+          rspamd_logger.infox(rspamd_config, "cannot make request to load DKIM selector for domain %s: %s", p.domain, err)
+        else
+          if data then
+            p.selector = data
```
fatalbanana: I fear this won't be called before it's needed. Probably, rest of the code applicable to redis signing should be moved into a function and called inside this callback (or if selector is not needed).

andryyy: Thanks, I will test the new method and report back. :)

PR #1624, opened 2017-05-09
[Feature] Preliminary DMARC reporting implementation

fatalbanana:
* Aggregation is being done in Lua; should happen in Redis rather
* Full DMARC report data for a given domain is in memory
* Whole base64-encoded report is in memory
* One report sent at a time
* Size limits not implemented
* Reporting address override not implemented
* Quite fragile: may skip reporting for a domain/orphan data or stop working completely in some cases
* Probably more issues

### DIFF
--- a/src/plugins/lua/dmarc.lua
+++ b/src/plugins/lua/dmarc.lua
@@ -17,7 +17,11 @@ limitations under the License.
 
 -- Dmarc policy filter
 
+local hash = require "rspamd_cryptobox_hash"
 local rspamd_logger = require "rspamd_logger"
+local mempool = require "rspamd_mempool"
+local rspamd_tcp = require "rspamd_tcp"
+local rspamd_url = require "rspamd_url"
 local rspamd_util = require "rspamd_util"
 local check_local = false
 local check_authed = false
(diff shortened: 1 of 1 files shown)


src/plugins/lua/dmarc.lua
```diff
@@ -77,14 +226,19 @@ end
...
     return nil
   end
-  local res = string.format('%d,%s,%s,%s,%s,%s', task:get_date(0),
-    ip:to_string(), tostring(spf_ok), tostring(dkim_ok),
-    disposition, (sampled_out and 'sampled_out' or ''))
+  local dkim_pass = table.concat(dres.pass or E, '|')
+  local dkim_fail = table.concat(dres.fail or E, '|')
+  local dkim_temperror = table.concat(dres.temperror or E, '|')
+  local dkim_permerror = table.concat(dres.permerror or E, '|')
+  local res = string.format('%d,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s',
```
vstakhov: I think, table + `table.concat` would be better here.

vstakhov:
Cool, merging.

PR #1639, opened 2017-05-13
 [Minor] debian: Fix lintian issues and refresh packaging

sbadia:
Hello, thanks again for rspamd/rmilter software!

As you know rmilter and rspamd are not in Debian anymore, we are trying to reintroduce rspamd and rmilter into Debian official repos with @jeromelebleu!

If you are ok with that, here is a little packaging refresh :)

Thanks in advance

This pr is linked to https://github.com/vstakhov/rmilter/pull/182


vstakhov:
Hello, thank you very much for your work. I just want to ask you couple of questions regarding of your proposed Debian packaging:

1. Rspamd is rapidly developed project, so old versions could have some awful bugs or be not reliable. Currently, I follow `mainline` and `stableline` model, where `mainline` is currently developed version and `stableline` is currently maintained stable branch. For example, now 1.6 is current `mainline` (no releases yet) and 1.5 is now `stableline`. I do not support anything beyond these versions. If there will be a chance to provide rotten versions to users then I will be strongly **against** from including Rspamd to any official repo.

2. There were couple of issues with embedded software bundled with Rspamd. First of all, there are couple of minified js scripts in Rspamd webui. I suggest to add Debian specific patches to load those scripts from some CDN or from some external repo (and minifying during build). There is also `hiredis` client embedded in Rspamd. It is **impossible** to use unpatched one with Rspamd so far.

In the meantime, I'll merge the proposed changes for both Rspamd and Rmilter. Please keep in touch with me if you plan some more efforts for this task. But I would really appreciate to know your thoughts about the questions I've arisen here.

sbadia:
Hello @vstakhov !

Many thanks for your valuable comments and questions, sorry for the late reply, we have not abandoned the idea to get back rspamd in Debian (because it's a valuable software!), we are now ready and we use `collab-maint` (a collaborative space to maintain software inside Debian). We (Jerome, and I) use rspamd on a daily basis.

About your first question and the developments cycles of Debian, I completely understand that LTS versions are a complex problem in term of maintenance and as you said awful bugs. In Debian we can easily import and package released version of rspamd (git tag/tarball) in the current testing distribution (currently buster), use debian-backports to reach the stable distribution (currently jessie) (and sure if needed an older distributions, of course it's more difficult for an older distribution than stable/oldstable). But I sure we can discuss how version you are more confident to be like a LTS or not; we can maybe discuss when a new Debian distribution is about to be release.

For your second question, about minified js, we can't use some cdn since the Debian policy disallow network access during builds, and also for privacy purposes, so we can take a look about your idea (it's pretty the same thing that repack the tarball to include [sources js files](https://lintian.debian.org/tags/source-is-missing.html)); about hiredis, since the sources files copyright are correctly attributed (in debian/copyright file), I don't see any problem, but I'll ask to ftp-masters.

Let me know if my answer is not clear (I'm not a native english speaker), or if you want to clarify some others points!

Thanks in advance!

Cheers,

PR #1644, opened 2017-05-18
[Minor] Relax CTYPE_MISSING_DISPOSITION rule

fatalbanana:
Not sure if this breaks the intention of the rule.


smfreegard:
Have you got a sample message that hit this incorrectly? 

My only comment would be that this looks like it's trying to address a message with inline images, but the images have the wrong ctype header?  e.g. application/octet-stream instead of image/*

fatalbanana:
I do; it bears an inline image with a bad mime type indeed.

smfreegard:
Ok - seems like a good workaround.

PR #1661, opened 2017-06-02
[Minor] Metadata exporter: round up symbol scores

moisseev:


moisseev:
@fatalbanana , On the second thought, maybe symbol scores in fixed format with 2 decimal places would be better, like in extended headers and logs. Unlike total score they are text anyway, right?
What do you think about adding another rounding function?
```lua
exports.round_str = function(x, n)
  return string.format('%.' .. (n or 0) .. 'f', x)
end
```

fatalbanana:
Fixed decimal places would be nice. Symbol scores are numbers but 'flatten' option used for email presents them as strings. Format string seems to be a lot more expensive than math unfortunately.

vstakhov:
I think that `or` is more expensive here...

fatalbanana:
I suppose probably it is better we don't round total score in general metadata and round this and symbol scores using `string.format('%.2f', n)` in "flattened" metadata.

Issue #1662, opened 2017-06-05
RBL monitoring when 127.0.0.1 exists in blacklist


wheldom01:
### Classification (Please choose *one* option):

* [] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [x ] Other bug
* [ ] Feature
* [ ] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
1.5.9

### Operation system, CPU, memory and environment:
Debian Jessie

### Description (Please provide a descriptive summary of the issue):
When a rbl contains the rbl monitoring address of 127.0.0.1 you get error messages in the log:

monitored; rspamd_monitored_dns_cb: DNS reply returned 'no error' for hostkarma.junkemailfilter.com while 'no records with this name' was expected

### Compile errors (if any):
n/a

### Steps to reproduce:
Add the following to local.d/rbl.conf and restart
rbls {

    hostkarma {
        symbol = "RBL_HOSTKAMA";
        rbl = "hostkarma.junkemailfilter.com";
        returncodes {
            RBL_HOSTKAMA_WHITE = "127.0.0.1";
            RBL_HOSTKAMA_BLACK = "127.0.0.2";
            RBL_HOSTKAMA_YELLOW = "127.0.0.3";
            RBL_HOSTKAMA_BROWN = "127.0.0.4";
            RBL_HOSTKAMA_NOBLACK = "127.0.0.5";
            RBL_HOSTKAMA_QUIT = "127.0.1.1";
            RBL_HOSTKAMA_NOQUIT = "127.0.1.2";
            RBL_HOSTKAMA_MIXEDQUIT = "127.0.1.3";
        }
    }
}

### Expected results:
No error message in log

### Actual results:
monitored; rspamd_monitored_dns_cb: DNS reply returned 'no error' for hostkarma.junkemailfilter.com while 'no records with this name' was expected

### Configuration:
See above

### Additional information:
Suggest that the monitoring ip for rbls be changed from 127.0.0.1 to 127.0.1.1

vstakhov:
[...]

PR #1664, opened 2017-06-05
Fix monitoring issue where 127.0.0.1 exists in rbl. hostkarma blacklist

wheldom01:
Fix for issue #1662


vstakhov:
No, I will not accept it: https://tools.ietf.org/html/rfc5782#section-5

Issue #1663, opened 2017-06-05
SURBL monitoring when facebook.com exists in blacklist 


wheldom01:
### Classification (Please choose *one* option):

* [] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [x ] Other bug
* [ ] Feature
* [ ] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
1.5.9

### Operation system, CPU, memory and environment:
Debian Jessie

### Description (Please provide a descriptive summary of the issue):
When a surbl contains the surbl monitoring address of facebook.com you get error messages in the log:

monitored; rspamd_monitored_dns_cb: DNS reply returned 'no error' for hostkarma.junkemailfilter.com while 'no records with this name' was expected

### Compile errors (if any):
n/a

### Steps to reproduce:
Add the following to local.d/rbl.conf and restart
rules {
    "HOSTKAMA_URIBL" {
        suffix = "hostkarma.junkemailfilter.com";
        noip = true;
        enabled = false;
        ips = {
                URIBL_HOSTKAMA_WHITE = "127.0.0.1";
                URIBL_HOSTKAMA_BLACK = "127.0.0.2";
                URIBL_HOSTKAMA_YELLOW = "127.0.0.3";
                URIBL_HOSTKAMA_BROWN = "127.0.0.4";
                URIBL_HOSTKAMA_NOBLACK = "127.0.0.5";
                URIBL_HOSTKAMA_24_48H = "127.0.2.1";
                URIBL_HOSTKAMA_LAST_10D = "127.0.2.2";
                URIBL_HOSTKAMA_OLDER_10D = "127.0.2.3";
        }
    }
}

### Expected results:
No error message in log

### Actual results:
monitored; rspamd_monitored_dns_cb: DNS reply returned 'no error' for hostkarma.junkemailfilter.com while 'no records with this name' was expected

### Configuration:
See above

### Additional information:
Suggest that the monitoring ip for rbls be changed from facebook.com to com

vstakhov:
You are right: it should check for another address: 

~~~
Domain-name-based DNSxLs MUST contain an entry for the [RFC2606]
   reserved domain name "TEST" and MUST NOT contain an entry for the
   reserved domain name "INVALID".
~~~

[...]

PR #1665, opened 2017-06-05
Fix monitoring issue where facebook.com exists in rbl. hostkarma blac…

wheldom01:
Fix for #1663


vstakhov:
I don't like this approach. I'd better add a dedicated option for surbl rules to check for some specific host in monitored.

PR #1675, opened 2017-06-11
[Conf] Do away with systemd specifics completely

fatalbanana:


vstakhov:
Hum, are we going with this change in 1.6? If yes, then we need to add the appropriate migration entry.

PR #1688, opened 2017-06-16
[Feature] milter_headers: add `extended_headers_rcpt` option

moisseev:
Milter_headers module adds extended headers to messages if EVERY envelope recipient matches `extended_headers_rcpt` list.

```
extended_headers_rcpt = ["user1", "@example1.com", "user2@example2.com"];
```


vstakhov:
I like the idea indeed!

moisseev:
What idea do you mean? The `extended_headers_rcpt` option was in Rmilter.

PR #1705, opened 2017-06-24
Fix Debian package build.

sebastianlipponer:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+rspamd (1.6.1) unstable; urgency=medium
+
+  * New release.
+
+ -- Sebastian Lipponer <mail@sebastianlipponer.de>  Sat, 24 Jun 2017 22:12:07 +0200
```
vstakhov: That will break my build scripts badly. I cannot accept this change.

sebastianlipponer: Which script in particular are you referring to? Nevertheless, then just apply the other commit to add the missing build-dependencies.

vstakhov: https://github.com/vstakhov/rspamd-build-scripts/blob/master/rspamd_build.sh#L551

PR #1734, opened 2017-07-07
[Feature] Reload file maps more frequently

fatalbanana:


src/libserver/cfg_utils.c
```diff
@@ -36,6 +36,7 @@
 #define DEFAULT_RLIMIT_NOFILE 2048
 #define DEFAULT_RLIMIT_MAXCORE 0
 #define DEFAULT_MAP_TIMEOUT 60.0 * 5
+#define DEFAULT_MAP_FILE_WATCH_MULTIPLIER 1
```
vstakhov: I'll set it to 0.1 I think.

PR #1753, opened 2017-07-15
IP netmask support for Redis multimap

mkuron:
[Multimap](https://rspamd.com/doc/modules/multimap.html)s of type `ip` currently only support IP addresses (e.g. 192.0.2.1) and not IP networks in CIDR notation (e.g. 192.0.2.0/24). Other places, like [user settings](https://www.rspamd.com/doc/configuration/settings.html), do support networks.

For consistency and to make it possible to specify a range of hosts without needing to create dozens of individual entries, this pull request adds support for networks to the multimap module. As one cannot search Redis for keys by matching a CIDR specification, this means we need to loop over all possible network masks.

Originally reported in https://github.com/mailcow/mailcow-dockerized/issues/412.

### DIFF
--- a/src/plugins/lua/multimap.lua
+++ b/src/plugins/lua/multimap.lua
@@ -35,7 +35,12 @@ local urls = {}
 
 local value_types = {
   ip = {
-    get_value = function(ip) return ip:to_string() end,
+    get_value = function(ip)
+      if type(ip) == 'string' then
+        return ip
+      end
+      return ip:to_string()
+    end,
   },
   from = {
     get_value = function(val) return val end,
@@ -373,14 +378,14 @@ local function multimap_callback(task, rule)
     if r['cdb'] then
       local srch = value
       if r['type'] == 'ip' then
-        srch = value:to_string()
+        srch = value_types['ip'].get_value(value)
       end
       ret = r['cdb']:lookup(srch)
     elseif r['redis_key'] then
       local srch = value
       if r['type'] == 'ip' or (r['type'] == 'received' and
         (r['filter'] == 'real_ip' or r['filter'] == 'from_ip' or not r['filter'])) then
-        srch = value:to_string()
+        srch = value_types['ip'].get_value(value)
       end
       ret = rspamd_redis_make_request(task,
         redis_params, -- connect params
@@ -636,6 +641,14 @@ local function multimap_callback(task, rule)
       if ip:is_valid() then
         if rt == 'ip' then
           match_rule(rule, ip)
+          local bits = 128
+          if ip:get_version() == 4 then
+            bits = 32
+          end
+          for i=bits,1,-1 do
+            local nip = ip:apply_mask(i):to_string() .. "/" .. i
+            match_rule(rule, nip)
+          end
         else
           local cb = function (_, to_resolve, results, err)
             task:inc_dns_req()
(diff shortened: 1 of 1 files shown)


vstakhov:
Multimaps support CIDR notation. Are you talking about Redis maps?

mkuron:
Multimaps configured with Redis, like this:
```
WHITELISTED_FWD_HOST {
  type = "ip";
  map = "redis://WHITELISTED_FWD_HOST"
}
```

If the redis hash contains an entry "192.0.2.1", it is matched for 192.0.2.1. If it only contains an entry "192.0.2.0/24", it is not matched. This PR fixes that.

vstakhov:
I don't think it is a good idea to send like 128 requests for querying of a single IP address. I would suggest the following procedure:

1. Query redis for existing masks (e.g. stored in a Redis list)
2. For each v4 or v6 mask perform the appropriate request

In this scheme, if you have just /32 and /24 masks you would need to to just 3 requests instead of 32. For IPv6 this scheme seems to be the only possible one.

mkuron:
Would it be better to query all 128 entries in a single `HMGET`? Then it's just one request.

mkuron:
Seems like `HMGET` is very efficient. In a simple benchmark, looking up 32 IPv4 networks via `HMGET` takes about 20% more time than looking up a single IPv4 address via `HGET`. Looking up 128 IPv6 networks via `HMGET` takes about 60% more time than looking up a single IPv6 address via `HGET`.

I have updated the pull request accordingly. This solution is much more efficient than doing multiple requests as you had suggested. Compared to not supporting netmasks in Redis multimaps, its overhead is less than one extra Redis request.

vstakhov:
Well, that looks good. However, for IPv4 we can reduce range of lookup to 8..29, since everything else is not used in practice. For IPv6 there is still no good solution due to that fact that IPv6 address can have multiple string representations (and even here there is no need in 128 requests since IPv6 address must have at least /64 for network part, so we can easily reduce number of requested keys by half).

mkuron:
> However, for IPv4 we can reduce range of lookup to 8..29, since everything else is not used in practice.

/8 is the largest allocation one was able to get from IANA back in the days, so that's a good upper bound. Regarding the lower bound, I'd stick to 32 as a list of hosts need not necessarily coincide with a subnet or IANA allocation -- someone might only want to add a certain subset of a network to an IP list. Furthermore, 192.0.2.1 is in principle the same as 192.0.2.1/32, and one might expect both to work. So I suggest 8..32.

> For IPv6 there is still no good solution due to that fact that IPv6 address can have multiple string representations

Right now, we assume the compressed form (no leading zeros, `::` for the largest block of zeros).

> and even here there is no need in 128 requests since IPv6 address must have at least /64 for network part, so we can easily reduce number of requested keys by half

One might not want to add an entire subnet (/64) to the list, but only a certain subset, as above for IPv4. It seems like the largest BGP announcements out there are for /19 networks and the largest companies (non-ISPs) get /32. So 32..128 would be a range that would cover all reasonable requirements.

I've updated the PR with 8..32 and 32..128. If you don't agree with the above and want me to shrink the ranges further, let me know.

mkuron:
Any chance this can be cherry-picked into an upcoming 1.6.3 release? It'd be great not to have to wait until 1.7.0 is released in a few months.

Since this fixed an inconsistent behavior between Redis and non-Redis maps, it's a bugfix and should thus be worth backporting.

PR #1756, opened 2017-07-17
[Fix] Don't use whitelist/greylist maps as regexp, but as map

dupondje:
In whitelist & greylist module, the domain files (like spf_dkim_whitelist.inc, etc) were read as regexp map.
This caused the following false whitelists:
WHITELIST_SPF_DKIM(-3)[onmicrosoft.com]
this while only 'microsoft.com' was in the list!

Reading the maps as 'map'/'hash' instead of 'regexp' fixes the issue.


vstakhov:
You are right, thank you!

PR #1781, opened 2017-07-31
[Minor] Fann_redis: Fix iterator & some warnings

fatalbanana:
- Also number of layers


vstakhov:
Oh, wait, the number of layers part is not correct.

PR #1786, opened 2017-08-04
MX_MISSING is RFC-compliant and should not increment score

mkuron:
When a domain has no MX record, an implicit MX record pointing to the domain name is assumed and the symbol _MX_MISSING_ is added. This comes with a score of 1.0. However, such implicit MX records are perfectly standards-compliant and found with surprisingly many legitimate mail servers. For details on implicit MX records, see [RFC 5321](https://tools.ietf.org/html/rfc5321) section 5 or [Wikipedia](https://en.wikipedia.org/wiki/MX_record#History_of_fallback_to_address_record).

This pull request sets the score of the symbol to zero so that this RFC-compliant and somewhat widespread behavior is not punished.

### DIFF
--- a/src/plugins/lua/mx_check.lua
+++ b/src/plugins/lua/mx_check.lua
@@ -186,7 +186,7 @@ local function mx_check(task)
         task = task,
         forced = true
       })
-      task:insert_result(settings.symbol_no_mx, 1.0, err)
+      task:insert_result(settings.symbol_no_mx, 0.0, err)
     else
       table.sort(results, function(r1, r2)
         return r1['priority'] < r2['priority']


vstakhov:
I do not care about RFC I do care about spam filtering. I don't see any point in your suggested changes from the perspective of spam fighting.

mkuron:
It seems arbitrary to consider _MX_MISSING_ as an indication for the likelihood of a sender being a spammer. I guess if something is permitted and somewhat commonly used by legitimate senders, it's not a good indicator for spam. In my (admittedly very limited) data set, _MX_MISSING_ seems to even be a lot more widespread on legitimate email (mostly coming through small web hosting companies that clearly haven't touched their DNS setups in years/decades) than on spam. In essence, this is like giving a +1.0 score to a randomly chosen subset of messages.

> I don't see any point in your suggested changes from the perspective of spam fighting.

It reduces false positives without significantly increasing false negatives. At least it does for me; please check your own history or training sets to see whether _MX_MISSING_ actually helps you discriminate between ham and spam or if it's about equally common on ham and spam.

smfreegard:
We're talking about a single point of score... MX_MISSING will hit upon hacked web servers that have been re-purposed to send out e-mail spam etc., so whilst it might be legal by the RFC - it is an anomaly which for things like spam filtering are important.   It might score 1.0 by itself, but when combined with things like rules which look for Wordpress URL signatures which point to plugin directories or admin functions, then it can obviously be scored higher via a composite.

If you want to report false-positives, then please provide the full symbol list of messages that you have seen that are false-positives along with the message headers to go with each, because for a rule like this to cause an FP - it has to have hit other rules.  We would then either reduce the score for this or other rules, or, create a composite rule to negate it in specific cases.

The more information we have regarding FPs, the better - but we can't go changing rule scores without some firm evidence, because as you can probably appreciate - there are specific individuals which would try and make us do this for reasons other than reducing FPs for ham...

PR #1805, opened 2017-08-22
Various changes to CentOS RPM

LorbusChris:
Update License to be the same as repo (ASL 2.0)
Remove hardcoded user and group
Add %{rspamd_pluginsdir}/lib/*.lua


LorbusChris:
wow that was quick :smile: thanks!
should I also PR these changes against the other repo?

PR #1829, opened 2017-09-07
Restore task:set_metric_subject() function

fatalbanana:


src/lua/lua_task.c
```diff
@@ -4069,6 +4076,25 @@ lua_task_set_metric_action (lua_State *L)
...
 static gint
+lua_task_set_metric_subject (lua_State *L)
+{
+	struct rspamd_task *task = lua_check_task (L, 1);
+	const gchar *subject;
+
+	subject = luaL_checkstring (L, 2);
+
+	if (task && subject) {
+		rspamd_mempool_set_variable (task->task_pool, "metric_subject", subject, NULL);
```
vstakhov: That won't work: `subject` is allocated on Lua side and have uncertain time of live. `rspamd_mempool_strdup` is required here I suppose.

vstakhov:
LGTM, merging

PR #1838, opened 2017-09-18
[Feature] Generic key name expansion for Redis keys

fatalbanana:


lualib/lua_redis.lua
```diff
@@ -146,6 +156,89 @@ end
...
+          return v
+        else
+          return
+        end
+      end
+      if gen_meta[k] then
+        v = gen_meta[k]() or '__empty__'
+        rawset(self, k, v)
+      end
+      if v ~= '__empty__' then
```
fatalbanana: Maybe I had better add some special type to use in place of '__empty__'; or not bother with this.

lualib/lua_redis.lua
```diff
@@ -146,6 +156,89 @@ end
...
+  }
+  process_cmd.get = process_cmd.set
+  process_cmd.hset = process_cmd.set
+  process_cmd.hget = process_cmd.set
+  process_cmd.hgetall = process_cmd.set
+  process_cmd.hdel = process_cmd.set
+  process_cmd.hincrby = process_cmd.set
+  process_cmd.hincrbyfloat = process_cmd.set
+  process_cmd.hscan = process_cmd.set
+  process_cmd.evalsha = process_cmd.eval
```
fatalbanana: I plan to review the rest of commands and add them as well.

lualib/lua_redis.lua
```diff
@@ -146,6 +156,89 @@ end
...
+      end
+      return a
+    end,
+    principal_recipient_domain = function()
+      return (task:get_principal_recipient() or E)['domain']
+    end,
+    ip = function()
+      local i = task:get_ip()
+      if i and i:is_valid() then return i:to_string() end
+    end
```
fatalbanana: Will want to add a couple more functions here I suppose: `smtp_from`, `asn`, whatever else is likely to be useful (unsure).

vstakhov:
Can you also provide some example usages of this feature?

fatalbanana:
The idea is to set `expand_keys = true` in `redis` configuration (so globally or in module configuration) & use some special values where-ever redis keys would be specified. Example using ratelimit configuration:

~~~
servers = "127.0.0.1";
expand_keys = true;
dynamic_rates {
  ip = "redis:rl-ip-specific:${ip}";
}
~~~

vstakhov:
Cool, I like it!

PR #1846, opened 2017-09-24
Fix recent regressions in force_actions

mkuron:
Some recent changes cause problems when using the force_actions module in rspamd 1.7 on the master branch.

The first problem is that using symbols with zero score inside the force_action expression no longer works after score matching was introduced in 64c4be01551d970c281c0e79c54ea16b972348d7. The code cannot differentiate between "expression did not match" and "expression matches but returned a zero score", so one cannot work around by setting the new `limit` setting to something less than zero. This PR, as a workaround, pretends that all symbols with zero or negative score have a score of 1 (as was the case before 64c4be01551d970c281c0e79c54ea16b972348d7).

The second problem is that once a force_action has been applied, all idempotent postfilters are skipped. Idempotence was introduced in 7a4c7de7035499eee2d68d525dc003feb18a749b and 7b902c00bce62688d9154ef76017135ad014c33c. In my case, that meant that messages that matched a force_action did not have any headers added (because MILTER_HEADERS was skipped) and did not show up in the history (because HISTORY_SAVE was skipped). Other postfilters may also be affected, but those are the ones where I saw the problem.

Please note that this PR does not solve the underlying problems (the lack of differentiation of "expression did not match" and "expression matches but returned a zero score" and the fact that idempotent filters may still have side effects like MILTER_HEADERS and HISTORY_SAVE do). So it should probably not be merged as-is but only considered as a starting point in tracking down the design issues in the two new features that caused the problems.


vstakhov:
No, HISTORY_SAVE and MILTER_HEADERS **MUST** remain idempotent.

mkuron:
@vstakhov, so history and headers are purposely not written when the force_actions module is used? That‘s going to cause a lot of trouble (e.g. it renders all force_actions with action _add_header_ useless). Or do you have another suggestion how to deal with symbols that are idempotent but have side effects?

vstakhov:
No, it seems to be some issue. I will investigate.

vstakhov:
And well, there are no side effects of history. Idempotent property means that they cannot change metric score, symbols or result in any way.

vstakhov:
Can you try this patch?

mkuron:
I‘ll try it tomorrow, though I  need to set up a build environment first. So far I‘ve only run rspamd from the apt package.

mkuron:
Unfortunately, the patch does not help. I‘m happy to try other suggestions you might have.

vstakhov:
Ok, I cannot reproduce your issue. Idempotent filters are correctly executed if force_action sets a result. How do you reproduce this issue?

vstakhov:
JFYI: I cannot reproduce this issue even without the mentioned patch.

mkuron:
Hmm, then it must be due to an interaction with some of my other settings. My configuration is the one from https://github.com/mailcow/mailcow-dockerized/tree/master/data/conf/rspamd . Since you couldn‘t reproduce the issue, I guess I‘ll have to come up with a minimal configuration that exhibits the idempotence problem.

Were you able to reproduce the other issue (the one with the zero scores)? That one looks pretty obvious in the Lua code.

mkuron:
The zero score issue is fixed (9f8bcbd508f081c79d75bc32b81cfb8ac09c4b1c). Thanks!

The issue of idempotent symbols not being executed when a force_action matches remains in the current master, however. I'm still in the process of tracking it down to a specific configuration parameter so I can provide you with a minimal example that exhibits the issue. The one thing I can already tell is that it's not milter-related as it happens when scanning a message via curl too.

mkuron:
I've isolated the problem with idempotent symbols not getting executed: it's caused by `autolearn = true` in statistic.conf. I've [attached a minimal configuration](https://github.com/vstakhov/rspamd/files/1383281/conf.zip) that exhibts the problem. Scan the [attached test message](https://github.com/vstakhov/rspamd/files/1383282/test.txt) using the following command to see whether the header would be added or not:
```
curl -s -H 'IP: 192.0.2.1' --data-binary @- http://127.0.0.1:11333/check < test.txt | json_pp | grep X-Spamd-Result
```
If you disable autolearn, the header is added; if you enable it, it is missing. Can you try to reproduce the problem like this, @vstakhov? I have updated the pull request with something that seems to fix it.

vstakhov:
Hey, that's a good catch indeed! I have added my own version of the fix that seems to be a more appropriate to my opinion. Could you please test it?

mkuron:
Thanks, that does appear to fix it.

One question: what does the `RSPAMD_TASK_FLAG_LEARN_AUTO` flag mean? If it is unset, idempotent symbols will still get skipped as per your change. I see that there are some places calling `rspamd_task_new` that don't set this flag, but all the ones that appear relevant (worker.c's `accept_socket`and rspamd_proxy.c's `rspamd_proxy_self_scan`) do.

Also, could you please trigger a re-build of the [nightly APT packages](https://rspamd.com/apt/dists/stretch/Release)? It looks like they're behind master by about 30 commits.

vstakhov:
This flag is set when learn is triggered during scan. Otherwise, when we learn manually, e.g. via `learn_spam` command this flag is not set and learn error is thus fatal. 

I've triggered a fresh packages build.

PR #1857, opened 2017-10-09
Fix display of symbols at least in chrome

gaetronik:
In chrome at least sym.name value is defined to "undefined" string and thus all symbol get displayed as undefined

### DIFF
--- a/interface/js/app/history.js
+++ b/interface/js/app/history.js
@@ -117,7 +117,7 @@ function($, _, Humanize) {
             preprocess_item(item);
             Object.keys(item.symbols).map(function(key) {
                 var sym = item.symbols[key];
-                if (!sym.name) {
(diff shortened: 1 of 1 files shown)


interface/js/app/history.js
```diff
@@ -117,7 +117,7 @@ function($, _, Humanize) {
             preprocess_item(item);
             Object.keys(item.symbols).map(function(key) {
                 var sym = item.symbols[key];
-                if (!sym.name) {
+                if (!sym.name || sym.name == "undefined") {
```
vstakhov: But why is it undefined? Maybe there is a better check for a missing hash field than this ugly comparison?

gaetronik: You're right, I think I understood where it's going replaced by "undefined" string. I will close this PR and open another one

PR #1865, opened 2017-10-17
[Fix] Fix empty threshold check in greylisting module

moisseev:


fatalbanana:
It seems it shouldn't matter either way but this code looks strange indeed.

moisseev:
@fatalbanana I believe `a == a` is always true in Lua. Am I wrong?

fatalbanana:
You're not wrong and there's no point in such a comparison, the effects of this are just not harmful.

~~~lua
if settings['greylist_min_score'] then
  -- we set greylist_min_score from config
else
  -- we try get it from metric
  local greylist_threshold = rspamd_config:get_metric_action('greylist')
  if true then
    -- maybe greylist_threshold is nil but doing the assignment doesn't hurt
    settings['greylist_min_score'] = greylist_threshold
  end
end
~~~

vstakhov:
`a == a` is not true in case of NaN. That's the point of this check.

fatalbanana:
So... I should not have merged this I suppose. ;) I wonder why would it be NaN though.

vstakhov:
It is NaN if action threshold is not set.

PR #1868, opened 2017-10-18
Revert "[Fix] Fix empty threshold check in greylisting module"

moisseev:
Reverts vstakhov/rspamd#1865


vstakhov:
Crap, in fact, `lua_config_get_metric_action` returns `nil` on NaN so your initial fix was right:

```C
if (!isnan (metric->actions[act].score)) {
  lua_pushnumber (L, metric->actions[act].score);
}
else {
  lua_pushnil (L);
}
```

PR #1876, opened 2017-10-21
DKIM header canonicalization: correctly handle more than two Received headers

mkuron:
When rspamd verifies the DKIM signature of an email that had more than two _Received_ headers before the signature was applied, it always finds that the signature is invalid -- even though e.g. Perl's Mail::DKIM finds the signature valid.

Let's assume that we have N _Received_ headers in the signed part of the message. When looking into `rspamd_dkim_canonize_header_relaxed`, one can now see that that first the first _Received_ header is canonicalized, then the second one, and then the first one is canonicalized N-2 more times. This is because `rspamd_dkim_parse_hdrlist_common` appears to not have been designed to handle more than two instances of the same header.

This pull request keeps track of the total header count inside the `htb` hash table to fix the issue.

### DIFF
--- a/src/libserver/dkim.c
+++ b/src/libserver/dkim.c
@@ -165,6 +165,7 @@ struct rspamd_dkim_sign_key_s {
 struct rspamd_dkim_header {
 	const gchar *name;
 	guint count;
+	guint count_total;
 };
 
 /* Parser of dkim params */
@@ -452,6 +453,7 @@ rspamd_dkim_parse_hdrlist_common (struct rspamd_dkim_common_ctx *ctx,
 					sizeof (struct rspamd_dkim_header));
 			new->name = h;
 			new->count = 0;
+			new->count_total = 1;
 
 			/* Check mandatory from */
 			if (!from_found && g_ascii_strcasecmp (h, "from") == 0) {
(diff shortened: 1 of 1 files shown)


vstakhov:
Do you have a particular sample for this issue?

mkuron:
Yes. I just sent it to you by private email. That message has four _Received_ headers before the signature. Its signature fails to verify with current master, but successfully verifies with this pull request.

vstakhov:
Ok, I see the problem. I'm just not convinced that we need to add another field to this structure now.

mkuron:
The `count` field is the sequential numbering for multiple instances of the same header. We can't reuse that as it needs to be different for each _Received_ header. Instead of putting another field into the struct (and only using it in the first header, as that is the one added to the `htb` hash table), we could create another hash table that keeps track of how many instances there are of each header.

vstakhov:
No, we can store next counter value in the hash table.

vstakhov:
Could you please test this patch?

vstakhov:
Grr + one more commit :(

mkuron:
Yes, that appears to work too. Thanks!

PR #1881, opened 2017-10-26
[Minor] Add check_authed and check_local

croessner:
Skip authenticated users or local networks


vstakhov:
This changes the default behaviour. Can we leave it as is?

PR #1895, opened 2017-10-30
[Minor] Multimap: support matching Received header flags

fatalbanana:
`flags` specifies one or more flags which *must* be present
`nflags` specifies one or more flags which must *not* be present


vstakhov:
Can you also update docs? Preferably with samples.

PR #1898, opened 2017-10-31
HTTP map: periodic callback after failed request

mkuron:
If the webserver responds with e.g. a temporary error, like a 502, to an HTTP map request, one sees an error message like
```
2017-10-30 14:23:05 #8(controller) <9fydgy>; map; http_map_finish: cannot load map http://172.22.1.251:8081/settings.php from 172.22.1.251: HTTP error 502
```
and rspamd will not ever update the map again until it is manually restarted. This is because the branch that deals with status codes other than 200 and 304 does not set the periodic callback. This pull request adds that, though I have not yet been able to confirm whether this completely fixes the issue.

### DIFF
--- a/src/libutil/map.c
+++ b/src/libutil/map.c
@@ -755,6 +755,7 @@ read_data:
 	else {
 		msg_info_map ("cannot load map %s from %s: HTTP error %d",
 				bk->uri, cbd->data->host, msg->code);
+		rspamd_map_periodic_callback (-1, EV_TIMEOUT, cbd->periodic);
 	}
 
 	MAP_RELEASE (cbd, "http_callback_data");


vstakhov:
I think this fix is not right but the problem really exists and should definitely be fixed.

vstakhov:
Could you please check the proposed fix? Thank you for hunting this long living bug indeed...

mkuron:
Thanks, that looks right. The difference between your fix and mine is that you set `cbd->periodic->errored = 1`, which of course makes sense.

vstakhov:
Yes, because without errored flag being set it might cause the combined maps failure. In fact, this particular bug has likely appeared just when I've added combined maps support (I mean maps with multiple backends here).

PR #1899, opened 2017-11-02
[Minor] Correct a typo

moisseev:


moisseev:
BTW Why is it called `*_ANY` if in the composite rule `AND` operation is actually used ?
```
    RBL_SPAMHAUS_XBL_ANY {
        expression = "RBL_SPAMHAUS_XBL & RECEIVED_SPAMHAUS_XBL";
    }
```

PR #1932, opened 2017-11-27
[Minor] Add whitelist symbol to mx_check plugin

AlexeySa:


src/plugins/lua/mx_check.lua
```diff
@@ -305,6 +311,14 @@ if opts then
     one_shot = true,
     one_param = true,
   })
+  rspamd_config:set_metric_symbol({
+    name = settings.symbol_white_mx,
+    score = -0.00,
+    description = 'Domain has whitelist MX',
```
moisseev: That's not true. We are whitelisting the domain itself, not MX.
What about 'Domain is whitelisted from MX check' ?

src/plugins/lua/mx_check.lua
```diff
@@ -305,6 +311,14 @@ if opts then
     one_shot = true,
     one_param = true,
   })
+  rspamd_config:set_metric_symbol({
+    name = settings.symbol_white_mx,
+    score = -0.00,
```
moisseev: Sign and precision do not make sense with zero score. I'd suggest to change the score to `0.0`.

src/plugins/lua/mx_check.lua
AlexeySa: Ok, I hurried )

vstakhov:
LGTM, merging.

PR #1938, opened 2017-11-30
get authenticated user from client certificate subject

philfry:
Hi,

in my postfix setup I have two different methods of authenticating users: sasl with passwords and client certificates.
rspamd only recognizes sasl authentication by evaluating the `{auth_authen}` macro. This tiny little PR also checks for `{cert_subject}`.

I've tested it against rspamd 1.6.5, works fine for me.
Any thoughts?

Kind regards

Phil


vstakhov:
Looks good to me.

PR #1939, opened 2017-12-01
[Minor] Fix & extend redis key expansion library

fatalbanana:
- Some functions did not work, fixed this
- Added new functions & variants (is there something else that's likely to be useful?)
- Moved table out of function
- I will add some applicable tests ...


vstakhov:
Hum, why [Minor] tag then? It looks like [Feature] to my opinion :)

fatalbanana:
That's `[Feature]` and/or `[Fix]` indeed but it falls under previous `[Feature]` tag somehow. ;)

PR #1947, opened 2017-12-05
[Fix] Use the right boolean operator on error check

mmcco:
It seems that someone used the wrong boolean operator here. The current condition means `if (True)`, while the new version makes perfect sense.

Thanks for your time,
Mike


vstakhov:
Another DeMorgan failure, thank you.

Issue #1981, opened 2018-01-10
`ip` matching is broken in user settings


moisseev:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [x] Other bug
* [ ] Feature
* [ ] Enhancement

### Reproducibility (Please choose *one* option):

* [x] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
1.7.0-git.e0d4da6

### Description (Please provide a descriptive summary of the issue):

User settings works when a network is specified as `ip` matching (e.g. `ip = "10.10.10.0/24";`), but single ip address (e.g. `ip = "10.10.10.0";`) cause an error on message scan.

### Steps to reproduce:
```
# rspamc -h localhost:11334
a
Results for file: stdin (0.008 seconds)
[Metric: default]
Action: add header
...
```
### Actual results:
```
2018-01-10 17:52:57 #60593(controller) <b06543>; lua; settings.lua:298: check for settings
2018-01-10 17:52:57 #60593(controller) <b06543>; csession; lua_metric_symbol_callback: call to (SETTINGS_CHECK) failed (2): /usr/local/share/rspamd/lua/settings.lua:155: invalid arguments; trace: [1]:{[C]:-1 - to_string [C]}; [2]:{/usr/local/share/rspamd/lua/settings.lua:155 - check_ip_setting [Lua]}; [3]:{/usr/local/share/rspamd/lua/settings.lua:181 - check_specific_setting [Lua]}; [4]:{/usr/local/share/rspamd/lua/settings.lua:323 - <unknown> [Lua]};
```
### Configuration:
```
# rspamadm configdump settings
*** Section settings ***
test {
    apply {
        default {
            HFILTER_HOSTNAME_UNKNOWN = 0;
        }
    }
    priority = "high";
    ip = "10.10.10.10";
}
```

PR #1982, opened 2018-01-10
[Fix] Settings: avoid checking invalid IP (#1981)

fatalbanana:


vstakhov:
I see, it should fix the issue indeed. However, my current `IP` API is way too bad.

Issue #1988, opened 2018-01-16
Header checks for settings map: merge multiple same name "header" fields


andryyy:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [ ] Other bug
* [ ] Feature
* [X] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
1.7.git

### Operation system, CPU, memory and environment:
Debian Stretch

### Description (Please provide a descriptive summary of the issue):
Having a settings map like this...

```
		header = {
			"From" = "/.*@bla/i";
			"From" = "/.*@blub/i";
		}
```

...will fail with...

```
map; lua_map_fin: call to local function failed: /usr/share/rspamd/lua/settings.lua:519: bad argument #1 to 'get_cached' (string expected, got table)
```

It should be less resource intensive to test each value itself than matching (1|2|3) with a huge list.

### Expected results (from aboves example):
Try any "From" value until match.

### Actual results:
Multiple "From" values are merged to a table and not matched.

PR #1989, opened 2018-01-16
[Fix] Settings: header: deal with multiple settings (#1988)

fatalbanana:
Support syntax like:

```
header = {"From" = "/bla/i"};
header = {"From" = "/ble/i"};
```

**OR**

```
header = {
   "From" = "/bla/i";
   "From" = "/ble/i";
}
```

### DIFF
--- a/src/plugins/lua/settings.lua
+++ b/src/plugins/lua/settings.lua
@@ -262,9 +262,19 @@ local function check_settings(task)
     end
 
     if rule['header'] then
-      for k, v in pairs(rule['header']) do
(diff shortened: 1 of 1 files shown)


fatalbanana:
Hmm, we seem to miss ability to create map content from Lua currently; I'm not sure I understand the suggestion.

vstakhov:
`rspamd_map_add` from maps.lua should do that I suppose

vstakhov:
Anyway, I'll merge this as is, let's think about improvements later.

PR #2002, opened 2018-01-27
[Fix] Virus infection string for F-PROT Antivirus

macgeneral:
Fixes the error `(rspamd_proxy) ; lua; antivirus.lua:396: Unhandled response: 1 <contains infected objects: VIRUSNAME>` and therefore no handled detection

--
F-PROT Antivirus response if an infected file was found is
`1 <contains infected objects: VIRUSNAME>` for me, while rspamd only checks for `1 <infected: ...>`.

Unfortunately lua patterns don't allow optional unmatched groups and I don't know if the string differs in older installations or older/other setups which is why I tried to mark contains and objects as optional strings (and infected surrounded by optional whitespace).

System this was tried on: CentOS 7.2; rspamd 1.6.5 and F-PROT Antivirus fpscand version 6.0.3 (latest version from 2012/2013 with the latest signatures from 2018).

### DIFF
--- a/src/plugins/lua/antivirus.lua
+++ b/src/plugins/lua/antivirus.lua
@@ -410,7 +410,9 @@ local function fprot_check(task, rule)
             rspamd_logger.infox(task, '%s [%s]: message is clean', rule['symbol'], rule['type'])
           end
         else
-          local vname = string.match(data, '^1 <.*infected.*: (.-)>')
+          -- returncodes: 1: infected, 2: suspicious, 3: both, 4-255: some error occured
+          -- see http://www.f-prot.com/support/helpfiles/unix/appendix_c.html for more detail
+          local vname = string.match(data, '^[1-3] <[%w%s]-: (.-)>')
           if not vname then
             rspamd_logger.errx(task, 'Unhandled response: %s', data)
           else
--- a/src/plugins/lua/antivirus.lua
+++ b/src/plugins/lua/antivirus.lua
@@ -410,7 +410,7 @@ local function fprot_check(task, rule)
             rspamd_logger.infox(task, '%s [%s]: message is clean', rule['symbol'], rule['type'])
           end
         else
-          local vname = string.match(data, '^1 <infected: (.-)>')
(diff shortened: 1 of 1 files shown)


macgeneral:
ok nevermind, I've got to work on this some more as it seems only infected can be triggered as well and won't get handled by my pattern x.x

> (rspamd_proxy); lua; antivirus.lua:396: Unhandled response: 1 <infected: EICAR_Test_File> 4200C201A964B->eicar.zip->eicar.com

macgeneral:
`^1 <.*infected.*: (.-)>`
works, while my old solution didn't work as expected.

You can trigger the
`^1 <infected: ...>`
message with [emailsecuritycheck.net](http://www.emailsecuritycheck.net/index.html) while you can trigger
`^1 <contains infected objects: ...>`
with [heise.de Emailcheck](https://www.heise.de/security/dienste/emailcheck/virendummies/netsky_p/)

philfry:
How about ```'^1 <%w* ?infected ?%w*: (.-)>'```?

macgeneral:
that works perfect and is easier to read! I will change my pull request in favor of this tomorrow if there are no doubts / objections.

macgeneral:
Maybe there could be an even more general approach working with the return codes 

http://www.f-prot.com/support/windows/fpwin_faq/310.html

http://www.f-prot.com/support/helpfiles/unix/appendix_c.html

macgeneral:
`'^[1-3] <.+: (.-)>'`

would cover infected, suspicious (heuristics) and both, ignoring all potential error codes.

philfry:
just in case: make it as non-greedy as possible `'^[1-3] <[%w%s]-: (.-)>'`

macgeneral:
Thanks for your help! I will change the pull are request in around 10 hours when I'm back at my PC :)
Just tried it on my server and your last version works like a charm.

vstakhov:
Is it ready to merge now?

macgeneral:
Yes, sorry. The string is working and tested on CentOS 7 with rspamd 1.6.5.

vstakhov:
Merged, thank you!

PR #2018, opened 2018-02-13
Allow signing incoming mails with recipient domain-key for ARC

chaosmaster:
ARC should be done on incoming mails using the key for the original recipient domain.
Use together with the following config in _local.d/arc.conf_

> sign_local = false;
> sign_inbound = true;
> allow_hdrfrom_mismatch = true;
> use_domain = recipient;

### DIFF
--- a/lualib/dkim_sign_tools.lua
+++ b/lualib/dkim_sign_tools.lua
@@ -37,6 +37,8 @@ local function prepare_dkim_signing(N, task, settings)
     rspamd_logger.debugm(N, task, 'mail is from address in sign_networks')
   elseif settings.sign_local and is_local then
     rspamd_logger.debugm(N, task, 'mail is from local address')
+  elseif settings.sign_received and not is_local and not auser then
+    rspamd_logger.debugm(N, task, 'mail was sent to us')
   else
     rspamd_logger.debugm(N, task, 'ignoring unauthenticated mail')
     return false,{}
@@ -55,9 +57,12 @@ local function prepare_dkim_signing(N, task, settings)
     return false,{}
   end
 
+  local eto = task:get_recipients(0)
+
   local dkim_domain
   local hdom = ((hfrom or E)[1] or E).domain
   local edom = ((efrom or E)[1] or E).domain
+  local tdom = ((eto or E)[1] or E).domain
   local udom = string.match(auser or '', '.*@(.*)')
 
   local function get_dkim_domain(dtype)
@@ -67,6 +72,8 @@ local function prepare_dkim_signing(N, task, settings)
       return edom
     elseif settings[dtype] == 'auth' then
       return udom
+    elseif settings[dtype] == 'recipient' then
+      return tdom
     end
   end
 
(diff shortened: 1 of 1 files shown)


lualib/dkim_sign_tools.lua
```diff
@@ -37,6 +37,8 @@ local function prepare_dkim_signing(N, task, settings)
     rspamd_logger.debugm(N, task, 'mail is from address in sign_networks')
   elseif settings.sign_local and is_local then
     rspamd_logger.debugm(N, task, 'mail is from local address')
+  elseif settings.sign_received and not is_local and not auser then
```
vstakhov: Probably, it is better to call it `sign_inbound`?

vstakhov:
The main problem with this approach is that we need to sign AAR header in this case, so we need to be extra carefull when dkim/arc/spf checks fail before doing ARC sign. I need to check this carefully...

lualib/dkim_sign_tools.lua
chaosmaster: That sounds better, I have changed it

chaosmaster:
As I understand the Arc-Message-Signature should always include the AAR.
However I wasn't able to get it signed, since the header is only added after the signing is done.

chaosmaster:
This example shows, that ARC seems to be accepted i.e. by google even without AAR being signed:

```
 ARC-Authentication-Results: i=3; mx.google.com;
        dkim=pass header.i=@mydomain.example.com header.s=dkim header.b=cd2a/xCZ;
        arc=pass (i=2 spf=pass spfdomain=openarc.org dkim=pass dkdomain=mydomain.example.com);
        spf=pass (google.com: best guess record for domain of srs0=irju=fi=openarc.org=forward@srs.example.com designates 2001:db8:7000::12 as permitted sender) smtp.mailfrom=SRS0=Irju=FI=openarc.org=forward@srs.example.com
```

- The mail went from _test@mydomain.example.com_ to _forward@openarc.org_ 
- It was then signed by _openarc.org_ and forwarded back to _test@mydomain.example.com_
- _mydomain.example.com_  (rspamd) then also added ARC and forwarded to _test@gmail.com_
- gmail also adds ARC, the results can be seen above

vstakhov:
OK, I have checked the code and it looks fine for this purpose:

```lua
  local arc_seals = task:cache_get('arc-seals')

  local ret,p = dkim_sign_tools.prepare_dkim_signing(N, task, settings)

  if not ret then
    return
  end

  p.arc_cv = 'none'
  p.arc_idx = 1
  p.no_cache = true
  p.sign_type = 'arc-sign'

  if arc_seals then
    p.arc_idx = #arc_seals + 1

    if task:has_symbol(arc_symbols.allow) then
      p.arc_cv = 'pass'
    else
      p.arc_cv = 'fail'
    end
  end
```

So it will set cv to `none` in case if it is the first ARC record, and it will check the previous chain in case of existing ARC headers. `Arc-Authentication-Result` is also added after signing. Hence, I conclude that signing on forwarding should probably work.

PR #2037, opened 2018-02-27
DMARC: apply lua_redis script functions

fatalbanana:
- Also fix lua_redis.exec_script

### DIFF
--- a/lualib/lua_redis.lua
+++ b/lualib/lua_redis.lua
@@ -843,8 +843,8 @@ local function add_redis_script(script, redis_params)
 end
 exports.add_redis_script = add_redis_script
 
-local function exec_redis_script(id, params, callback, args, aargs)
-  local args_modified = false
+local function exec_redis_script(id, params, callback, keys, args)
+  local keys_modified = false
 
   if not redis_scripts[id] then
       logger.errx("cannot find registered script with id %s", id)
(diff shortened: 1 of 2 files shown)


lualib/lua_redis.lua
```diff
@@ -879,7 +879,6 @@ local function exec_redis_script(id, params, callback, args)
     end
 
     if not args_modified then
-      table.insert(args, 1, tostring(#args))
```
vstakhov: What is the reason to do it?

src/plugins/lua/bayes_expiry.lua
```diff
@@ -207,7 +207,7 @@ local function expire_step(cls, ev_base, worker)
   lredis.exec_redis_script(cls.script,
       {ev_base = ev_base, is_write = true},
       redis_step_cb,
-      {'RS*_*', cls.expiry, cur}
+      {3, 'RS*_*', cls.expiry, cur}
```
vstakhov: I don't understand why do you add this complexity. `fann_redis` will also be broken after this commit.

fatalbanana: `fann_redis` does not use `exec_redis_script`; I don't see how it could be broken.

lualib/lua_redis.lua
fatalbanana: The reason is we have key-arguments and non-key arguments: https://redis.io/commands/eval and we can't infer how many of each (as it is we assume they are all keyword arguments which is wrong).

src/plugins/lua/bayes_expiry.lua
vstakhov: It does. I still don't understand why have you changed it by the way.

fatalbanana: the reason for adding this parameter is to deal with the change in `lua_redis.lua`

lualib/lua_redis.lua
vstakhov: We use keyed args only as far as I know. That's why I've added this assumption. It makes this call more errors prone and more simple.

fatalbanana: That's not so in dmarc, ratelimit and url_tags.

src/plugins/lua/bayes_expiry.lua
fatalbanana: I see, I managed to miss it somehow. I can fix this. I could also use key arguments only everywhere.

lualib/lua_redis.lua
vstakhov: I still don't like this idea. Perhaps, we should push args somehow else? Varargs or whatever. Setting keys count explicitly is something I want to avoid at any price.

fatalbanana: Sure, that's a possibility; we could have table argument for keys of each type.

vstakhov: Why not just change its signature to the following one:

```lua
function exec_redis_script(id, params, callback, keys, args)
  if args then ... end
end
```

lualib/lua_redis.lua
```diff
@@ -843,7 +843,7 @@ local function add_redis_script(script, redis_params)
 end
 exports.add_redis_script = add_redis_script
 
-local function exec_redis_script(id, params, callback, args)
+local function exec_redis_script(id, params, callback, args, aargs)
```
vstakhov: Why `aargs` instead of `keys` and `args` as it is used in Redis `eval` command? In fact, even `args` was a misguiding name.

fatalbanana: `args` is keys and `aargs` are argument-arguments. :) we can rename them...

PR #2038, opened 2018-02-27
[Feature] Support sending given header multiple times in lua_http

fatalbanana:
I discovered I have reason to send a given HTTP header multiple times in lua_http. This seems to work.

~~~lua
local rspamd_http = require 'rspamd_http'
rspamd_config.POINTLESS_HTTP = {
  callback = function (task)
    local function http_callback(err_message, code, body, headers)
      return
    end
    rspamd_http.request({
      task=task,
      url='http://127.0.0.1:56789/hello',
      body='hi',
      callback=http_callback,
      headers={
        One='10',
        Two='66',
        Multiple={'AHOY','HI','HELLO'},
      },
    })
  end
}
~~~

### DIFF
--- a/src/lua/lua_http.c
+++ b/src/lua/lua_http.c
@@ -316,16 +316,28 @@ static void
 lua_http_push_headers (lua_State *L, struct rspamd_http_message *msg)
 {
 	const char *name, *value;
+	gint i, sz;
 
 	lua_pushnil (L);
 	while (lua_next (L, -2) != 0) {
 
 		lua_pushvalue (L, -2);
 		name = lua_tostring (L, -1);
-		value = lua_tostring (L, -2);
-
-		if (name != NULL && value != NULL) {
-			rspamd_http_message_add_header (msg, name, value);
+		sz = rspamd_lua_table_size (L, -2);
+		if (sz != 0 && name != NULL) {
+			for (i = 1; i <= sz ; i++) {
+				lua_rawgeti (L, -2, i);
+				value = lua_tostring (L, -1);
+				if (value != NULL) {
+					rspamd_http_message_add_header (msg, name, value);
+				}
+				lua_pop (L, 1);
+			}
+		} else {
+			value = lua_tostring (L, -2);
+			if (name != NULL && value != NULL) {
+				rspamd_http_message_add_header (msg, name, value);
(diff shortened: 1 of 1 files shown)


src/lua/lua_http.c
```diff
@@ -316,16 +316,31 @@ static void
...
 	lua_pushnil (L);
 	while (lua_next (L, -2) != 0) {
 
 		lua_pushvalue (L, -2);
 		name = lua_tostring (L, -1);
-		value = lua_tostring (L, -2);
-
-		if (name != NULL && value != NULL) {
-			rspamd_http_message_add_header (msg, name, value);
+		if (lua_istable (L, -2)) {
```
vstakhov: I'd suggest using of `rspamd_lua_table_size` instead of this heuristic.

fatalbanana: Do you mean instead of `for (i = 1; ; i++) { .. if (lua_isnil (L, -1)) {` story? I'm not sure what to change on this line.

vstakhov: If you check code of the function (rspamd_lua_table_size) it returns 0 for a non-table.

src/lua/lua_http.c
```diff
@@ -316,16 +316,28 @@ static void
...
+			sz = rspamd_lua_table_size (L, -2);
+			for (i = 1; i <= sz ; i++) {
+				lua_rawgeti (L, -2, i);
+				value = lua_tostring (L, -1);
+				if (name != NULL && value != NULL) {
```
vstakhov: Why cannot we check `name` sanity before that loop by the way?

src/lua/lua_http.c
fatalbanana: Yes, I saw this, but I was worried about case of empty table. :)

src/lua/lua_http.c
fatalbanana: Oh. we could. That is copy/paste job...

src/lua/lua_http.c
vstakhov: It should be treated as an error I suppose. `set_header('name', '')` or `set_header('name', {''})` should be absolutely fine on the other hand.

Issue #2010, opened 2018-02-01
filter "reject" on history-page


linuzer:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [X] WebUI/Usability
* [ ] Serious bug
* [ ] Other bug
* [ ] Feature
* [ ] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
1.6.5
### Description (Please provide a descriptive summary of the issue):
If you choose the filter "reject" on the history tab, then all rejected mails (soft and hard rejects) are displayed. I would expect to see only the hard rejected ones. The other filters work like this, so you see only what you have choosen. 

### Steps to reproduce:
- click on "history"
- choose filter "reject"
- observe the result
### Expected results:
only hard rejects
### Actual results:
all rejects - soft and hard ones

moisseev:
https://github.com/vstakhov/rspamd/issues/1604#issuecomment-296455656

PR #2041, opened 2018-02-28
Fix filter reject on history-page #2010

kakshay21:
Fix filter reject on history-page #2010


kakshay21:
@moisseev it's because the `reject -soft` is not present in the `actions` array.
I could have added that to the array itself but felt like UI should be clean.
If that's okay for you then, I can add this to the array itself.

moisseev:
I really don't know which option is better. I just think an empty drop box can confuse users. It can be misread as the filter is not enabled.

kakshay21:
@moisseev  can you check, please.

PR #2043, opened 2018-02-28
[Minor] Remove redundand condition (#2041)

moisseev:
@kakshay21 Sorry, I overlooked it.

### DIFF
--- a/interface/js/app/history.js
+++ b/interface/js/app/history.js
@@ -473,7 +473,7 @@ function($, _, Humanize) {
             if (action instanceof FooTable.Filter) {
                 if(action.query.val() === 'reject -soft'){
                     this.$action.val('reject');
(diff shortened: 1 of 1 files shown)


kakshay21:
@moisseev Have you checked the UI after this change? because as far as I remember changing to `else` instead of `else if` breaks the UI at least for me😅. I know programmatically else should be there. But IDK some strange behavior may be.

moisseev:
@kakshay21 Sure. I checked it before proposing the change. Would you recheck if it really breaks WebUI for you?

kakshay21:
@moisseev My bad! my previous change was creating this bug. I was checking this inside `www` folder and when I built it again to see this commit, It's not breaking anymore. Sorry😅.

Issue #1910, opened 2017-11-10
Greylist form vanished with negative value


timscha:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [x] WebUI/Usability
* [ ] Serious bug
* [ ] Other bug
* [ ] Feature
* [ ] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Description (Please provide a descriptive summary of the issue):
When you enter a negative value (for example "-2" in the Greylist form on Configuration and save the data the form vanished. Only workaround to make it visible again is, to to change the value in var/lib/rspamd/rspamd_dynamic to a positive value.

### Steps to reproduce:
In the web interface go to Configuration. Under actions, set the greylisting to a negative value and safe it.

vstakhov:
Don't set it to negative value.

fatalbanana:
Thresholds for metric actions should typically not be negative, and some order for actions (if defined) is expected to be maintained: `no action <= greylist <= add header <= rewrite subject <= reject`. If you want to disable `greylist` action you can redefine actions in `override.d/metrics.conf`. Note that this won't disable greylisting. To disable greylisting, set `enabled = false` in `local.d/greylist.conf` (or don't give it redis settings).

timscha:
Guys, I know that I don't have to use negative values. But I think the form should catch values likes this.

vstakhov:
Yes, that's correct.

PR #2061, opened 2018-03-05
Fixes greylist form vanish

firedranzer:
Fixes the vanishing of greylist form when a negative value was entered. #1910

### DIFF
--- a/interface/js/app/config.js
+++ b/interface/js/app/config.js
@@ -215,51 +215,35 @@ function($) {
                     $('#actionsFormField').attr('disabled', true);
                 }
 
-                $('#saveActionsClusterBtn').on('click', function() {
+                function saveActions(callback) {
                     var elts = loadActionsFromForm();
                     // String to array for comparison
                     var eltsArray = JSON.parse(loadActionsFromForm());
-                    if(eltsArray[0]>=0 && eltsArray[1]>=0 && eltsArray[2]>=0){
-                        rspamd.queryNeighbours('saveactions', null, null, "POST", {}, {
+                    if(eltsArray[0]<0){
+                        rspamd.alertMessage('alert-modal alert-error', 'Spam can not be negative');
+                    }
+                    else if(eltsArray[1]<0){
+                        rspamd.alertMessage('alert-modal alert-error', 'Probable spam can not be negative');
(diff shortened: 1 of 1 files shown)


vstakhov:
Probably, we should check all thresholds for negative values, not only greylist one.

firedranzer:
@vstakhov is this okay?

interface/js/app/config.js
```diff
@@ -217,17 +217,31 @@ function($) {
...
+                    // String to array for comparison
+                    var eltsArray = JSON.parse(loadActionsFromForm());
+                    if(eltsArray[2]>=0){
+                        rspamd.queryNeighbours('saveactions', null, null, "POST", {}, {
+                            data: elts,
+                            dataType: "json",
+                        });
+                    }
+                    else{
+                        alert("Greylist can not be negative");
```
moisseev: IMO it's better to use `rspamd.alertMessage()` function.

firedranzer:
@vstakhov @moisseev implemented the changes. please review it.

moisseev:
@firedranzer It's a good practice to reuse the code. Both checks for "Save cluster" and "Save actions" are quite similar. Would you refactor them in a function?

firedranzer:
@moisseev We can only refactor the `else` part because the `Save Cluster` and `Save Actions` have different posting methods.

moisseev:
@firedranzer I think even just `else` part is better than nothing as it will grow when you add a threshold order check, but different methods shouldn't be a problem since you can call them as callback functions.

```js
function yourFunction(callback) {
    callback(args);		
}

yourFunction(queryNeighbours);

yourFunction(queryLocal);
```

firedranzer:
@moissev I think it will be good to refactor just the `else` part because introducing a callback function will unnecessarily complicate the code.

moisseev:
@firedranzer I'm not agree with you. In fact, callback will compact the code, make it easier to read and it is very easy to accomplish.

firedranzer:
@vstakhov @moisseev I have implemented all the changes suggested by you.

moisseev:
LGTM
It would be nice if a native speaker could take a look at alert messages texts.

PR #2063, opened 2018-03-06
Allow the multimap header type to be a list of headers as well as a single header

trallen:
I'm writing a multimap set that relies on different headers with similar values set by different MTAs. This would allow you, for example, to merge the freemail_envfrom and freemail_envrcpt multimaps to use one symbol if you wished, or to check both To and Cc headers for an address without writing two maps.

You may already have nixed this idea for good reasons! If not, I have found it a useful option.


vstakhov:
LGTM, thank you!

PR #2072, opened 2018-03-13
[Fix] Fix various modules in case of empty message

fatalbanana:


vstakhov:
That was missing indeed, thank you!

PR #2089, opened 2018-03-19
Corpus/Rescore improvements, new and improved rules

smfreegard:


rules/regexp/headers.lua
```diff
@@ -917,3 +923,36 @@ reconf['HAS_XOIP'] = {
...
+
+reconf['HAS_LIST_UNSUB'] = {
+  re = string.format('%s', 'header_exists(List-Unsubscribe)'),
+  description = 'Has List-Unsubscribe header',
+  score = -0.01,
+  group = 'headers'
+}
+
+reconf['HAS_GUC_PROXY_URI'] = {
+   re = '/\\.googleusercontent\\.com\\/proxy/{raw_mime}i',
```
fatalbanana: Why not a URL regex?

smfreegard: Honestly - I can't remember.

vstakhov: Can we change it?

smfreegard: Of course - once the new version is out, then I'll do a corpus run and check it against the previous results.  I doubt it will make any difference to the result.

smfreegard: Ok - done.

vstakhov: Cool, thank you!

PR #2093, opened 2018-03-19
Subject privacy for webUI

treibholz:
I added an option for "Subject privacy" in the webUI, as the subject may include personal information, which I don't want to see. This is only relevant when redis is used as storage for the history.

It just hashes the subject, so you can still see that a lot of spam-mails have the same subject.
I chose md5 as default, because it results in a shorter string.

### DIFF
--- a/conf/modules.d/history_redis.conf
+++ b/conf/modules.d/history_redis.conf
@@ -18,6 +18,8 @@ history_redis {
   key_prefix = "rs_history"; # Default key name
   nrows = 200; # Default rows limit
   compress = true; # Use zstd compression when storing data in redis
+  subject_privacy = false; # subject privacy is off
+  subject_privacy_alg = 'md5'; # default hash-algorithm to obfuscate subject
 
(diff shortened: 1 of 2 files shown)


vstakhov:
We can cut the blake2 string to size of md5 as well. But the best way is to use HMAC based hash and add some routine to be able to decrypt these strings using a secret key (rspamadm subroutine). However, I can merge this PR as is so far and gradually rework it in future.

treibholz:
You are right, cutting it is a better idea than md5. I guess, I could remove the prefix, too (or change it to something generic like "obfuscated".

Using an obfuscation, that is not reversible, was intentionally. But for other usecases something reversible might be useful, too.

vstakhov:
Well, HMAC is not reversible as well. But you cannot prove that F('blah.com') == '0xdeadbabe' without knowing of some secret key, namely, F('blah.com', 'secret') == '0xdeadbabe'.

treibholz:
hmm, my very limited knowledge of lua (almost none until today) makes it impossible for me, to do this now.
So I vote for "gradually rework it in future". :-)

PR #2103, opened 2018-03-22
Fix and clean up Dockerfile

TimWolla:
This pull request updates the Dockerfile as follows:

1. Make it compatible with Debian Stretch: gpg needs to be installed explicitely as of stretch.
2. Pin to `debian:stretch` to not encounter surprises once `buster` is gold.
3. <strike>Update the usage of `curl` to the [official guidelines](https://github.com/docker-library/official-images). In the long term the GPG key should be published to keyservers and retrieved by it's ID.</strike> see: https://github.com/vstakhov/rspamd/pull/2103#issuecomment-375282793
4. Declare `/var/lib/rspamd` as a volume, as it stores persistent information.
5. Expose the controller worker outside of the Container (localhost just does not make sense with Docker).
6. Replace overriding of the full `rspamd.conf` with files in `override.d`.

Note that I'm currently evaluating the use of rspamd for myself and thus some things (especially bullet point 6) might not be optimal. I wanted to give back my changes anyway, as running rspamd inside of Docker is crucial for me (and a working Dockerfile is strictly better than a broken one). I'm open for feedback!


andryyy:
Looks good to me, but since you refer to official guidelines: use `	&& bla` instead of `&&	bla` :-)

TimWolla:
> Looks good to me, but since you refer to official guidelines: use _TAB_&& bla instead of &&_TAB_bla :-)

@andryyy In fact I maintain two official images myself where I use the original style of this pull request. But sure: Updated :-)

I also updated the Dockerfile to make bullet point 3 obsolete: The key is now fetched from the keyservers (I must have missed it yesterday).

I made my changes in two new separate commits. I can squash if you want.

andryyy:
I am just repeating what's in their guides. But who cares about the indent before `& ` really? 🤷‍♂️

croessner:
What about an optional volume for the config?

TimWolla:
> What about an optional volume for the config?

Can you create volumes for the config just fine, the `VOLUME` declaration just states what by default is a volume. It makes sense for `/var/lib/rspamd/`, as it stores non-volatile information that is regularly written and thus needs the better performance of a volume.

For the config you can provide the volume or bind mounts at runtime. Personally I just bake the config into the container by creating a descendent image.

PR #2111, opened 2018-03-25
Improve readability of AR-header

bjoe2k4:
Concat `hdr_parts` with a newline instead of a space, similar to OpenDKIM and the implementation of many big mail providers (i.e. gmail).

### DIFF
--- a/lualib/lua_auth_results.lua
+++ b/lualib/lua_auth_results.lua
@@ -191,8 +191,10 @@ local function gen_auth_results(task, settings)
     table.insert(hdr_parts, hdr)
   end
 
+  local fold_symbol = task:has_flag("milter") and "\n" or "\r\n"
+
   if #hdr_parts > 0 then
-    return table.concat(hdr_parts, '; ')
+    return table.concat(hdr_parts, ';' .. fold_symbol)
   end
 
   return nil


vstakhov:
It is not a correct approach to fold headers (but there is a routine `lua_util.fold_header`). Furthermore, it needs to be different in milter/non-milter modes but that should be somehow handled by Rspamd itself in the same matter as it is done with DKIM signatures.

bjoe2k4:
I saw `lua_util.fold_header` but its folding depends on the length of the header. While this is technically correct, the folding might not happen after the semicolon, and the readability is not much improved.

vstakhov:
Well, I'm still not with you. What is the point in 'readability of AAR'? Are you expecting some users to actually read email headers? Your approach will potentially break a lot of messages as it doesn't take line endings (+milter specific) into account. That's why I cannot accept it.

bjoe2k4:
I am actually reading mail headers from time to time. Also other headers like `Received` are most of the time formatted for readability.
Anyways, i have updated the commit taking into account line-endings for milter/non-milter mode.

vstakhov:
That won't work as well as there could be different line endings in non-milter mode. I have dig into this issue and provided a more straightforward solution to this issue.

bjoe2k4:
Thank you very much for looking into this. I've just recompiled rspamd with the current master 5b73f19bc26253b7c80a783ee100a96a644610df. I don't think this is working as expected. As far as i can understand a709da98c462abc9f6b08e4288525ad92698685f, it should be possible to fold on semicolons now, which is not the case.
```
Authentication-Results: mail.mydomain.com; dkim=pass
	header.d=mydomain.com; dmarc=pass (policy=none)
	header.from=mydomain.com; spf=pass smtp.mailfrom=_me_@gmail.com
```
Furthermore i am seeing the following issues:
- `X-Spamd-Result` has lost its readability.
- `ARC-Seal` `b=` is not folded at all and shown on a single line.
- Any DKIM signature's `h=` field is not folded at all.
- If there is an `arc=pass` present, it only appears in `ARC-Authentication-Results` and is missing from `Authentication-Results`

If needed, I can open seperate issues.

vstakhov:
That should be fixed in the master branch.

bjoe2k4:
Yes it is, thank you. Maybe the folding function can be further improved by skipping all other folding rules if a stopcharacter is provided. This would avoid folding like this:
```
	 IP_SCORE(-1.78)[ip: (-4.85), ipnet: 94.237.32.0/21(-2.43), asn:
	202053(-1.58), country: FI(-0.04)];
```

bjoe2k4:
I'd consider this fixed now, however something still prevents the arc-seal from being folded. but this is a different issue.

PR #2153, opened 2018-04-05
[WebUI] Add symbols order selector to history

moisseev:


interface/js/app/history.js
```diff
@@ -605,6 +615,9 @@ function($, _, Humanize) {
             e.preventDefault();
             interface.getHistory(rspamd, tables, neighbours, checked_server);
         });
+        $("#selSymOrder").change(function() {
+            interface.getHistory(rspamd, tables, neighbours, checked_server);
```
spikerwork: `interface.getHistory(rspamd, tables, neighbours, checked_server);`
`e.preventDefault();` // Avoid recursion problem

PR #2220, opened 2018-05-04
[Feature] support for CommuniGate Pro self-generated messages

andy-igoshin:


fatalbanana:
Why add those settings at all? `arc` and `dkim_signing` already sign messages from 127.0.0.0/8 by default and `sign_networks` does not exist in `dkim` module.

andy-igoshin:
`arc` and `dkim_signing` already sign messages from 127.0.0.0/8 by default only if you use `sign_local = true` and `local_addrs` contains 127.0.0.0/8.

about dkim you right.

PR #2240, opened 2018-05-11
Small libucl fixes

flowerysong:
626c7a1 attempted to fix vertical tab handling, but '\v' is not a valid JSON escape so most parsers will reject it.

libucl has two places that escape strings, 626c7a1 didn't modify the one that occurs during parsing, so I have done so.

As far as I can tell there's no reason for lua_ucl to use UCL_STRING_ESCAPE; strings are escaped again during JSON/UCL generation, producing excessively escaped output. Example:
```
"description": "To header display name is \\\"Recipients\\\""
```

### DIFF
--- a/contrib/libucl/lua_ucl.c
+++ b/contrib/libucl/lua_ucl.c
@@ -550,10 +550,10 @@ ucl_object_lua_import_escape (lua_State *L, int idx)
 	t = lua_type (L, idx);
 	switch (t) {
 	case LUA_TTABLE:
-		obj = ucl_object_lua_fromtable (L, idx, UCL_STRING_ESCAPE);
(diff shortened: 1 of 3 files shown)


vstakhov:
eea6b5ea92584c588e5b65fce86aaf9f1441eb38 looks quite slashing. I need to check if it breaks anything.

flowerysong:
I don't think it will break anything, but it might. We've been using a similar patch in production since mid-February and haven't encountered any issues, but that build is based on 1.6.6 and there are significant portions of code that our setup doesn't exercise.

vstakhov:
Can you please apply that to vstakhov/libucl?

PR #2283, opened 2018-06-11
[Fix] Fix for CommuniGate Pro maillist

andy-igoshin:


vstakhov:
BTW, it would be good to have a corresponding test in `test/functional/` for this symbol (and for every other symbol in fact).

PR #2350, opened 2018-07-18
Implement memory efficient stat convert

oxpa:
I had an old statistics database with 11 million keys on a fairly old host.
`redis.call('HGETALL', k)` caused OOM condition on that host: redis would use more memory than it is allowed to use.

Current implementation is memory efficient, but less replication friendly: even though, replication is not broken, redis may suffer from high load for some time.

Conversion time is more or less same for the old and current approach (at leas, it takes seconds to convert my 11M keys DB)


vstakhov:
Cool, that's very useful, thank you!

PR #2352, opened 2018-07-18
Add mos.ru to whitelists

oxpa:
Mos.ru is more or less popular website in Moscow region of Russia. They send notifications about housing and utilities payments, and related information.

They send emails with various problems (wrong msg ID, URLs from RBL, etc), but with valid DKIM and matching SPF. They also have 'reject' policy enable for DMARC.

I believe, it's worth adding mos.ru into whitelists to improve delivery of their letters


vstakhov:
Could you please open PR against https://github.com/rspamd/maps repo?

PR #2375, opened 2018-07-26
Clickhouse retention

negram:
Added optional retention for ClickHouse tables

### DIFF
--- a/src/plugins/lua/clickhouse.lua
+++ b/src/plugins/lua/clickhouse.lua
@@ -18,6 +18,9 @@ local rspamd_logger = require 'rspamd_logger'
 local rspamd_http = require "rspamd_http"
 local rspamd_lua_utils = require "lua_util"
 local upstream_list = require "rspamd_upstream_list"
+local lua_util = require "lua_util"
+local ucl = require "ucl"
+
 local N = "clickhouse"
 
 if confighelp then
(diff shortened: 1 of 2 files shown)


src/plugins/lua/clickhouse.lua
```diff
@@ -741,6 +750,174 @@ local function clickhouse_collect(task)
...
+    return nil
+  end
+
+  local current_ts = os.time()
+
+  if last_ts == nil or (last_ts + settings.retention.period) <= current_ts then
+    local res
+    res, err = write_file:write(tostring(current_ts))
+    if err then
+      rspamd_logger.errx(ev_base, 'Failed to write %s, will not perform retention: %s', ts_file, err)
```
vstakhov: ev_base seems to be an invalid userdata here. I'd recommend to use `rspamd_config`: https://github.com/vstakhov/rspamd/blob/master/src/lua/lua_logger.c#L503 (or fix it here)

src/plugins/lua/clickhouse.lua
```diff
@@ -741,6 +750,174 @@ local function clickhouse_collect(task)
...
+  <int> - how many seconds wait until next run
+]]
+local function get_last_removal_ago(ev_base)
+  local ts_file = string.format('%s/%s', rspamd_paths['DBDIR'], 'clickhouse_retention_run')
+  local f, err = io.open(ts_file, 'r')
+  local write_file
+  local last_ts
+
+  if err then
+    rspamd_logger.debugm(N, ev_base, 'Failed to open %s: %s', ts_file, err)
```
vstakhov: Same about `ev_base`

PR #2403, opened 2018-08-07
Add negative weight on DNSWL entries

heraklit256:
This supersedes https://github.com/vstakhov/rspamd/pull/2281 and adds some negative score to DNSWL entries. However, if a message is not properly DKIM signed, it is only granted -0.5 points.

In case `R_DKIM_ALLOW` symbol is present and it has medium or high trust at DNSWL, it is granted -2.0 / -4.0 points in summary. This prevents some unconditional positive effect just because of a listing there.


conf/scores.d/rbl_group.conf
```diff
@@ -29,15 +29,15 @@ symbols = {
         description = "Sender listed at https://www.dnswl.org, no trust";
     }
     "RCVD_IN_DNSWL_LOW" {
-        weight = 0.0;
+        weight = -0.5;
```
vstakhov: I still think these scores shouldn't be changed.

heraklit256: If you don't mind, I disagree.

To be listed at DNSWL with some trust (if an IP has no trust, it won't get any negative score for it) takes some effort indicating that a mail server at this IP is actively maintained and monitored - in case of spamming, an IP loses trust very quickly.

Second, spammers usually are not listed here since it makes them tracable and is a time-consuming procedure. Besides, if they are known to spam, it won't help them.

-0.5 points is not very much so it is unlikely to push a spam mail into ham category. Negative scores with more weigth are only granted if a mail is DKIM signed, which is what you mentioned in first place. Besides of this limitation, I consider it safe to "honour" a positive DNSWL listing with the proposed score.

vstakhov: Then it should be probably -0.1 for low, -0.2 for med and -0.5 for high.

heraklit256: Okay, I am fine with this. Should the scores in the composite rules be adjusted, so we get -2.0 and -4.0 in the end? Or do you just want to keep it this way?

vstakhov: The current weights are fine.

heraklit256: Could you adjust them and merge the pull request or should I send in a new one?

vstakhov: Just add a commit with a fix.

PR #2404, opened 2018-08-07
add NiX spam RBL

heraklit256:
NiX spam is an RBL run by German publisher Heise Zeitschriften Verlag. Their listing policies are quite conservative, with an automatic delisting after 24 hours last seen, so the amount of false positives is low.

Since it is widely used in the D-A-CH area, I consider it being useful for rspamd.

This pull request also fixes some small typos in other RBL descriptions and unitises them.


conf/scores.d/rbl_group.conf
```diff
@@ -142,16 +142,21 @@ symbols = {
...
     }
 
     "RBL_VIRUSFREE_BOTNET" {
         weight = 2.0;
-        description = "Source IP is listed in virusfree.cz BL";
+        description = "From address is listed in virusfree.cz BL";
+    }
+
+    "RBL_NIXSPAM" {
+        weight = 4.0;
```
vstakhov: This score seems to be too high. Can we start with a lower default weight for it?

heraklit256: In the pull request, the score is the same as for Spamhaus XBL. Personally, I consider NiX spam to be similar (very low FP, reported IPs are highly unwanted, etc.) so I propose this score.

What score would you prefer?

PHPGangsta:
I would really love to see NIX Spam supported by rspamd.

They mostly list spam trap hits (plus some hash matches).
Here you can find some statistics: Good accuracy while having 0 false positives (which is very important to be sure it's spam).
https://www.intra2net.com/en/support/antispam/index.php_sort=accuracy_order=desc.html
The details page of NIX spam shows it's constantly at 0% false positives:
https://www.intra2net.com/en/support/antispam/blacklist.php_dnsbl=rcvd_in_nix_spam.html

They list single IP addresses for 12 hours (not 24 hours as written above) + you can manually delist: http://www.dnsbl.manitu.net/stats.php?language=en

From the "false positives" point of view, it's comparable to Spamhaus ZEN or XBL. So a weight of 4.0 would be OK.
It's definitly better than Spamhaus SBL (weight of 2.0 in rspamd), which has a false positive rate >0%:
https://www.intra2net.com/en/support/antispam/blacklist.php_dnsbl=uribl_sbl.html

PR #2411, opened 2018-08-10
[Minor] Fixed parsing of faked mx-record:

negram:
priority goes first and then the value


vstakhov:
Hum, why do we need that? So we would now write `10:some_mx` instead of `some_mx:10` but what is the purpose of such a change?

PR #2412, opened 2018-08-10
[Minor] Add record type into log when set up fake records

negram:


contrib/librdns/rdns.h
```diff
@@ -391,6 +391,14 @@ const char *rdns_strtype (enum rdns_request_type type);
  */
 enum rdns_request_type rdns_type_fromstr (const char *str);
 
+/**
+ * Returns string representing request type
+ * @param rcode
+ * @return
+ */
```
vstakhov: Could we have `rdns_` prefix for this function please?

PR #2441, opened 2018-08-24
Coroutines to finish/periodic/onload

negram:


src/lua/lua_config.c
```diff
@@ -2871,11 +2874,8 @@ lua_periodic_callback (gint unused_fd, short what, gpointer ud)
 			plan_more = timeout > 0 ? TRUE : FALSE;
 		}
 
-		lua_pop (L, 2); /* Return value + error function */
+		lua_pop (L, 1); /* Return value + error function */
```
vstakhov: The comment now becomes misleading.

PR #2443, opened 2018-08-25
[Fix] Properly remove missing dkim/spf whitelist files.

aj-gh:
Unfortunately the changes in commit 85d3aa18edb5c5efba5561361f97670884b2443d seem to have broken loading of the following local spf / dkim whitelist include files, at least with rspamd 1.7.9-1~stretch/amd64.
This PR moves the removed URLs outside the "domains" part which fixes this issue.

Previously:
```
$ rspamadm configdump whitelist
(..)
        WHITELIST_DKIM {
            score = -1;
            domains [
                "                \"/var/lib/rspamd/dkim_whitelist.inc.local\"",
                "fallback+file:///etc/rspamd/dkim_whitelist.inc",
            ]
            valid_dkim = true;
            description = "Mail comes from the whitelisted domain and has a valid DKIM signature";
        }
        WHITELIST_SPF {
            description = "Mail comes from the whitelisted domain and has a valid SPF policy";
            domains [
                "                \"/var/lib/rspamd/spf_whitelist.inc.local\"",
                "fallback+file:///etc/rspamd/spf_whitelist.inc",
            ]
            valid_spf = true;
            score = -1;
        }
    }
(..)
```

Afterwards:
```
$ rspamadm configdump whitelist
(..)
    WHITELIST_DKIM {
        score = -1;
        domains [
            "/var/lib/rspamd/dkim_whitelist.inc.local",
            "fallback+file:///etc/rspamd/dkim_whitelist.inc",
        ]
        valid_dkim = true;
        description = "Mail comes from the whitelisted domain and has a valid DKIM signature";
    }
    WHITELIST_SPF {
        description = "Mail comes from the whitelisted domain and has a valid SPF policy";
        domains [
            "/var/lib/rspamd/spf_whitelist.inc.local",
            "fallback+file:///etc/rspamd/spf_whitelist.inc",
        ]
        valid_spf = true;
        score = -1;
    }
(..)
```

rspamd then correctly reads the local files.

A better fix might be to address the underlying issue that these comments don't seem to get properly parsed. But I'll leave that one to the experts.

PR #2447, opened 2018-08-28
[Conf] Fix whitelist.conf

moisseev:
Reported by @bjo81 via Telegram group


vstakhov:
Accepted, but I need to fix bug in ucl at some day.

aj-gh:
Fixed 3 days before this was opened in #2443 - will close that one then.

PR #2451, opened 2018-08-30
[wip] Coroutines tcp

negram:


lualib/lua_tcp_sync.lua
```diff
@@ -0,0 +1,213 @@
...
+
+  self._conn = connection
+
+  setmetatable(self, metatable)
+
+  return self
+end
+
+--[[[
+-- @function tcp_sync.read_once()
```
vstakhov: It looks like a `@method`

src/lua/lua_tcp.c
```diff
@@ -92,6 +92,61 @@ rspamd_config:register_symbol({
 
 LUA_FUNCTION_DEF (tcp, request);
 
+/***
+ * @method connect_sync()
```
vstakhov: `lua_tcp.connect_sync` - it is a function and it should reference the module as well.

src/lua/lua_tcp.c
```diff
@@ -185,6 +303,23 @@ struct lua_tcp_dtor {
 #define LUA_TCP_FLAG_SHUTDOWN (1 << 2)
 #define LUA_TCP_FLAG_CONNECTED (1 << 3)
 #define LUA_TCP_FLAG_FINISHED (1 << 4)
+#define LUA_TCP_FLAG_SYNC (1 << 5)
+
+#undef TCP_DEBUG_REFS
+#ifdef TCP_DEBUG_REFS
+#define TCP_RETAIN(x) do { \
+	msg_err ("retain ref %p, refcount: %d", (x), (x)->ref.refcount); \
```
vstakhov: Perhaps, `info` is better here.

src/lua/lua_tcp.c
```diff
@@ -342,7 +515,17 @@ lua_tcp_push_error (struct lua_tcp_cbdata *cbd, gboolean is_fatal,
 	struct lua_callback_state cbs;
 	lua_State *L;
 
-	lua_thread_pool_prepare_callback (cbd->task->cfg->lua_thread_pool, &cbs);
+	if (cbd->thread) {
+		va_start (ap, err);
```
vstakhov: Why do we need a copy here counting that we don't use the original va_list?

src/lua/lua_tcp.c
```diff
@@ -1320,6 +1626,175 @@ lua_tcp_request (lua_State *L)
...
+
+	if (0 > port || port > 65535) {
+		return luaL_error (L, "invalid port given (correct values: 1..65535)");
+	}
+
+	if (task == NULL && (cfg == NULL || ev_base == NULL || session == NULL)) {
+		return luaL_error (L, "invalid arguments: either task or config+ev_base+session should be set");
+	}
+
+	if (timeout < 0.00001) {
```
vstakhov: It looks weird. I'd suggest to set timeout to NaN, and switch `timeout=N` to `timeout=D` and then check for `isnan` here and set default timeout if it is NaN.

src/lua/lua_tcp.c
```diff
@@ -1320,6 +1626,175 @@ lua_tcp_request (lua_State *L)
...
+	cbd->cfg = cfg;
+	cbd->thread = lua_thread_pool_get_running_entry (cfg->lua_thread_pool);
+
+	h = rspamd_random_uint64_fast ();
+	rspamd_snprintf (cbd->tag, sizeof (cbd->tag), "%uxL", h);
+	cbd->handlers = g_queue_new ();
+
+	cbd->ev_base = ev_base;
+	cbd->flags |= LUA_TCP_FLAG_SYNC;
+	msec_to_tv (timeout, &cbd->tv);
```
vstakhov: Why `msec_to_tv` and *= by 1000 previously? Perhaps, it is better to use `double_to_tv` and seconds here.

src/lua/lua_tcp.c
```diff
@@ -1494,6 +1969,226 @@ lua_tcp_shift_callback (lua_State *L)
...
+	gdouble ms = lua_tonumber (L, 2);
+
+	if (cbd == NULL) {
+		return luaL_error (L, "invalid arguments: self is not rspamd{tcp_sync}");
+	}
+	if (lua_type (L, 2) != LUA_TNUMBER) {
+		return luaL_error (L, "invalid arguments: second parameter is expected to be number");
+	}
+
+	ms *= 1000.0;
```
vstakhov: It looks that you multiply seconds by 1000. Or milliseconds by 1000000.

src/lua/lua_tcp.c
```diff
@@ -1494,6 +1969,226 @@ lua_tcp_shift_callback (lua_State *L)
...
+static gint
+lua_tcp_sync_shutdown (lua_State *L)
+{
+	LUA_TRACE_POINT;
+	struct lua_tcp_cbdata *cbd = lua_check_sync_tcp (L, 1);
+	if (cbd == NULL) {
+		return luaL_error (L, "invalid arguments [self is not rspamd{tcp_sync}]");
+	}
+
+	shutdown (cbd->fd, SHUT_WR);
```
vstakhov: It differs from the standard unix shutdown semantics. I don't know, maybe leave it as is for now but consider adding `SHUT_RD` as an optional argument.

src/lua/lua_tcp.c
negram: copy-paste from lua_tcp_set_timeout () ¯\_(ツ)_/¯

src/lua/lua_tcp.c
negram: fixed

PR #2470, opened 2018-09-08
add composite rules against private health insurance spam

heraklit256:
Trying to get rid of a Krankenhaus (private health insurance) spam wave.

Most of them are generated by (abused?) PHP scripts, and contain subjects with questions or exclaims.

Sore is low to prevent false positives.


heraklit256:
Sorry for being a bit chaotic today, added some more rules containing this topic.

PR #2471, opened 2018-09-08
Add blocklist.de RBL

heraklit256:
Blocklist.de (https://www.blocklist.de/en/index.html) is a reporting service for brute force detection tools such as `fail2ban`, which can be queried like any other DNSBL.

Being listed indicates an IP address of (massively) having attacked other computers in the past, usually because it is infected, or operated by professional cyber criminals. Since it is useful to know if a message passed such systems on its way, this RBL will be used for `Received`-IPs, too.

I am using this service for years, experiencing nearly no false positives. Since it consults popular white lists before an IP address is added, I consider it very reliable. Because of this (and its nature of listing infected systems), the scores are equal to those of Spamhaus XBL (4.0 and 3.0).


heraklit256:
I disagree.

Of course, parsing the `Received` headers reveal a lot of information to an RBL, but since they are "public" metadata, I consider the user to care about them. For example, they may use a proxy for delivering mails to the submission server, hiding their DSL IP address.

Personally, RBL checks against `Received` IPs is one of the features of which I choose rspamd: It is extremly helpful to know if a mail has been relayed through an IP with bad reputation. Of course, there are false positives (ISP with bad hygiene, ...), but they are not too bad for me.

heraklit256:
Thank you for merging it. I hope I wasn't too rude in my answer.

PR #2492, opened 2018-09-18
Sa body regex

negram:


src/libmime/message.c
```diff
@@ -315,7 +319,7 @@ rspamd_strip_newlines_parse (const gchar *begin, const gchar *pe,
 			case seen_cr:
 				/* \r\n */
 				if (!crlf_added) {
-					if (IS_PART_HTML (part) || g_ascii_ispunct (last_c)) {
+					if (IS_PART_HTML (part) || !url_open_bracket) {
```
vstakhov: We should leave `ispunct` case here. Otherwise, in the following case:

```
hello.
again
```

we would have `hello.again` that might break (?) some regular expressions (those, that rely on `\s` and not on `\b`).

PR #2502, opened 2018-09-19
[Test] Few "basic" test cases for selectors

negram:


negram:
2500 PR & 5000 build got!

PR #2511, opened 2018-09-20
[Test] Nicely print table difference when they do not match

negram:


negram:
failure is irrelevant

PR #2537, opened 2018-09-27
Add Spamcop to RBL List

robgnu:
This PR ist for using Spamcop for RBL checking. I have quite good results for international users.


robgnu:
I think it is a good way to start with a low score for Spamcop. Maybe +1 or +1.5? A FP would not beak everything, but it would confirm other symbols...

PR #2543, opened 2018-09-27
leaving original symbols if DNSWL composite rules are triggered

heraklit256:
Without policy = "leave", rspamd replaces the original symbols
in log and message header (if enabled), which makes debugging more
hard and is not used in this case.

### DIFF
--- a/conf/composites.conf
+++ b/conf/composites.conf
@@ -85,16 +85,19 @@ composites {
         expression = "RECEIVED_PBL & -RCVD_VIA_SMTP_AUTH";
         description = "Relayed through ZEN PBL IP without sufficient authentication (possible indicating an open relay)";
         score = 2.0;
+        policy = "leave";
     }
     RCVD_DKIM_ARC_DNSWL_MED {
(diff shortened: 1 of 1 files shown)


moisseev:
Probably default policy should be `remove_weight`. Or do you want to leave symbols' weights as well?

heraklit256:
Both of them: I would prefer to leave the original symbol names (`leave`) and their weight - at least for these three composite rules.

Personally, I find it puzzling if symbol list and/or weight changes because of composite rules. However, there are definitely use cases for doing so.

Thereof, I would rather not use `remove_weight` here. :-)

heraklit256:
@moisseev : Thank you for the hint. I lowered the scores on the DNSWL composite rules and hope everything is fine now.

Please let me know if it isn't.

moisseev:
Looks good to me. But I am not sure about `RCVD_DKIM_ARC_DNSWL_HI` score. What do you think about  -0.5 ... -1.0 ?

heraklit256:
Here you go...

PR #2544, opened 2018-09-27
Fix description for composite rule RBL_SPAMHAUS_XBL_ANY

heraklit256:


vstakhov:
I think the current description is more precise then the one you are suggesting.

vstakhov:
Or wait, there is `and` not `or`, then you are right but the composite name is unfortunate and confusing.

PR #2547, opened 2018-09-28
Change dcc / antivirus plugins

c-rosenberg:
- Antivirus: Next upstream should be tried on any error. Current Upstream should be set to fail in that case
- Antivirus: While scanning mime_parts only is great for caching scanned attachment digests, sometimes it's useful to scan the textparts (e.g. ClamAV Unofficial Sanesecurity ...).
- DCC: separate socket and server configuration to make use of the upstreams feature.

### DIFF
--- a/conf/modules.d/antivirus.conf
+++ b/conf/modules.d/antivirus.conf
@@ -18,8 +18,12 @@ antivirus {
   clamav {
     # If set force this action if any virus is found (default unset: no action is forced)
     # action = "reject";
-    # if `true` only messages with non-image attachments will be checked (default true)
-    attachments_only = true;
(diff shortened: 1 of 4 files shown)


vstakhov:
I'm sorry but I cannot accept the antivirus module changes: you've broken the backward compatibility by removing `attachments_only` option. Is it possible to preserve this option and its meaning?

c-rosenberg:
Sorry, forgot to push the workaround for attachments_only

vstakhov:
Could you also fix luacheck warnings please?

vstakhov:
Looks good to me, thank you very much!

PR #2557, opened 2018-10-02
[Fix] ip_score - respect check_authed and check_local settings

c-rosenberg:
check_authed  and check_local wasn't overwritten by options set in the config file.


src/plugins/lua/ip_score.lua
```diff
@@ -355,7 +359,7 @@ end
 
 -- Configuration options
 local configure_ip_score_module = function()
-  local opts = rspamd_config:get_all_opt('options')
```
vstakhov: Good catch!

PR #2566, opened 2018-10-04
Minor Composite rule cleanup

heraklit256:
Following pull request #2543 , some other rules are updated so original symbols will be preserved. In addition, some missing descriptions were added.

Sorry for becoming so sloppy here. :-)


heraklit256:
`- ping -`

:-)

conf/composites.conf
```diff
@@ -103,16 +103,19 @@ composites {
         expression = "(HAS_X_POS | HAS_PHPMAILER_SIG | HAS_X_PHP_SCRIPT) & (SUBJECT_ENDS_QUESTION | SUBJECT_ENDS_EXCLAIM | MANY_INVISIBLE_PARTS)";
         description = "Message was generated by PHP script and contains some spam indicators";
         score = 1.0;
+        policy = "leave";
     }
     PHISH_EMOTION {
         expression = "(PHISHING | DBL_PHISH | PHISHED_OPENPHISH | PHISHED_PHISHTANK) & (SUBJECT_ENDS_QUESTION | SUBJECT_ENDS_EXCLAIM)";
         description = "Phish message with subject trying to address users emotion";
         score = 2.0;
```
moisseev: Probably the score should be adjusted as you leave symbols scores. `PHISHING` score is dynamic and in many cases can be close to zero. Adding significant score to it can lead to false positives.

conf/composites.conf
```diff
@@ -103,16 +103,19 @@ composites {
...
+        policy = "leave";
     }
     HAS_ANON_DOMAIN {
         expression = "HAS_GUC_PROXY_URI | URIBL_RED | DBL_ABUSE_REDIR";
         description = "Contains one or more domains trying to disguise owner/destination";
         score = 0.5;
+        policy = "leave";
```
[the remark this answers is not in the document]
heraklit256: The rule became useful in environments where mails containing URLs with anonymised domains should be wrapped or rejected because of some policy guidelines. Since it was difficult to find out which symbols indicate this behaviour, `HAS_ANON_DOMAIN` was a good summary of it. Perhaps some other users have this problem, too.

Would lowering the score (0.1 ?) or setting it to 0.0 suit you?

conf/composites.conf
heraklit256: Thanks for the comment; I am thinking about adding other phishy symbols to it to avoid FPs.

Is 1.0 low enough for now?

conf/composites.conf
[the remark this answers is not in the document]
moisseev: ok, I got it. I would use 0.1 to make it positive.

conf/composites.conf
moisseev: Yes, I think so.

heraklit256:
Here you go...

moisseev:
Thank you.

PR #2587, opened 2018-10-12
cyrptobox: work with OpenSSL-1.1

rsmarples:
ERR_load_EC_strings and friends don't need to be called with
OpenSSL-1.1. Infact, on some platforms it causes a cyclic init
which stops rspamd from working entirely.

Luckily OpenSSL-1.1 should load these string automatically so we
can avoid calling them.


rsmarples:
I don't understand the drone failure above. Exit code 3 is not clear as to why it's failing....

vstakhov:
Ignore drone.ci for now. I'm with you about the deprecation for these functions, however, I'm not completely sure about the proposed fix. I will check that later, thanks for report!

PR #2589, opened 2018-10-13
add rule for .onion URIs

heraklit256:
Recently, I observed some unwanted messages containing `.onion` URIs (mostly for paying ransom, or "unlock instructions"). Since they are not necessarily malicious, the `HAS_ONION_URI` does not get a positive score.

However, it is included to the `HAS_ANON_DOMAIN` composite group, which may or may not be filtered by force actions.


heraklit256:
Closing due to merge conflict. :-\

PR #2606, opened 2018-10-23
[Dmarc] Changed message id domain

jpbede:
Changed static message-id domain to report_settings.helo


jpbede:
I thought mistakenly it had to do with a problem i had. After further research it was fixed due to a other change.

PR #2607, opened 2018-10-23
Run GDB on cores dumped by functional test

citrin:
Run GDB on cores dumped by functional test and other small changes.


citrin:
GDB prints warning 'warning: core file may not match specified executable file'. Is it OK? Can it be caused by changing process title in rsapmd?

vstakhov:
Yes, that's the reason.

PR #2616, opened 2018-10-30
also query resolved IPs against Spamhaus ZEN

heraklit256:
Currently, resolved IP addresses of a domain are "only" checked against Spamhaus SBL. Since it might be useful to know if an IP address is listed in PBL, XBL, or even DROP, these are added here.

To gather full results, `zen.spamhaus.org` must be used instead of `sbl.spamhaus.org`.

### DIFF
--- a/conf/modules.d/surbl.conf
+++ b/conf/modules.d/surbl.conf
@@ -87,12 +87,15 @@ EOD;
             }
         }
 
-        "SBL_URIBL" {
-            suffix = "sbl.spamhaus.org";
+        "SPAMHAUS_ZEN_URIBL" {
+            suffix = "zen.spamhaus.org";
             resolve_ip = true;
             ips {
                 URIBL_SBL = "127.0.0.2";
                 URIBL_SBL_CSS = "127.0.0.3";
+                URIBL_XBL = ["127.0.0.4", "127.0.0.5", "127.0.0.6", "127.0.0.7"];
+                URIBL_PBL = ["127.0.0.10", "127.0.0.11"];
+                URIBL_DROP = "127.0.0.9";
             }
         }
 
--- a/conf/scores.d/surbl_group.conf
+++ b/conf/scores.d/surbl_group.conf
@@ -163,7 +163,7 @@ symbols = {
         description = "A domain listed in the mail is listed in Spamhaus XBL URIBL";
     }
     "URIBL_PBL" {
-        weight = 0.1;
+        weight = 0.01;
         description = "A domain listed in the mail is listed in Spamhaus PBL URIBL";
     }
     "URIBL_DROP" {
--- a/conf/scores.d/surbl_group.conf
+++ b/conf/scores.d/surbl_group.conf
@@ -152,26 +152,26 @@ symbols = {
     }
     "URIBL_SBL" {
         weight = 6.5;
-        description = "A domain in the mail resolves to an IP listed in Spamhaus SBL";
+        description = "A domain in the message body resolves to an IP listed in Spamhaus SBL";
     }
     "URIBL_SBL_CSS" {
         weight = 6.5;
-        description = "A domain in the mail resolves to an IP listed in Spamhaus SBL CSS";
+        description = "A domain in the message body resolves to an IP listed in Spamhaus SBL CSS";
     }
     "URIBL_XBL" {
         weight = 6.5;
-        description = "A domain in the mail resolves to an IP listed in Spamhaus XBL";
+        description = "A domain in the message body resolves to an IP listed in Spamhaus XBL";
     }
     "URIBL_PBL" {
         weight = 0.01;
-        description = "A domain in the mail resolves to an IP listed in Spamhaus PBL";
+        description = "A domain in the message body resolves to an IP listed in Spamhaus PBL";
     }
     "URIBL_DROP" {
         weight = 8.0;
-        description = "A domain in the mail resolves to an IP listed in Spamhaus DROP";
+        description = "A domain in the message body resolves to an IP listed in Spamhaus DROP";
     }
     "RBL_SARBL_BAD" {
         weight = 2.5;
-        description = "A domain in the mail is blacklisted in SARBL";
+        description = "A domain in the message body is blacklisted in SARBL";
    }
 }
--- a/conf/scores.d/surbl_group.conf
+++ b/conf/scores.d/surbl_group.conf
@@ -152,26 +152,26 @@ symbols = {
     }
     "URIBL_SBL" {
         weight = 6.5;
-        description = "A domain listed in the mail is listed in Spamhaus SBL URIBL";
+        description = "A domain in the mail resolves to an IP listed in Spamhaus SBL";
     }
     "URIBL_SBL_CSS" {
         weight = 6.5;
-        description = "A domain listed in the mail is listed in Spamhaus SBL CSS URIBL";
+        description = "A domain in the mail resolves to an IP listed in Spamhaus SBL CSS";
     }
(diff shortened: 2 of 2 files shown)


heraklit256:
`- ping -`

:-)

smfreegard:
Quoted from Spamhaus' website on Zen:

> Do not use ZEN in filters that do any ‘deep parsing’ of Received headers, or for anything other than checking IP addresses that hand off to your mailservers.

Whilst it might yield sone results checking IPs of hosts against the CBL (XBL), that would need testing first and shouldn't form part of the base rspamd ruleset until it's proven.

heraklit256:
Well, the [complete quote](https://www.spamhaus.org/zen/) says something else:

> Caution: Because ZEN includes the XBL and PBL lists, do not use ZEN on smarthosts or SMTP AUTH outbound servers for your own customers (or you risk blocking your own customers). Do not use ZEN in filters that do any ‘deep parsing’ of Received headers, or for anything other than checking IP addresses that hand off to your mailservers.

In my opinion, Spamhaus assumes their RBLs as being a hard reject criteria - in `rspamd`, they are not. A message gets some additional points if something (Received IP, direct IP, or domain) is listed there, but it is not rejected just because of one symbol (use `force_actions.conf` if this behaviour is desired).

Of course, blocking your own customers does not make sense, but if a message was relayed through an IP address known for spam (SBL) or malicious (XBL, CBL) behaviour, you usually want to know about it. Same goes for domains.

Since we also look domains up in URIBLs, I do not see why we should not do so with the IP addresses behind them.

PHPGangsta:
ZEN contains PBL.
https://www.spamhaus.org/pbl/
"The PBL lists both dynamic and static IPs, any IP which by policy (whether the block owner's or -interim in its absence- Spamhaus' policy) should not be sending email directly to the MX servers of third parties."

In other words: PBL contains all dynamic IPs from private DSL and cable end users. Nearly all mails contain IPs from these dynamic ranges if the user provider does not remove it via "IP stripping". The very first Received-header should contain a dynamic IP in nearly all cases where a mail client is used (Outlook, Thunderbird, ...).

I would say, ZEN should not be applied to all Received headers of a mail because of PBL, especially not the very first Received-header. XBL and DROP might make sense.

heraklit256:
@PHPGangsta : Yes, but mails are not penalised because Spamhaus PBL triggers:
```
    "RECEIVED_SPAMHAUS_PBL" {
        weight = 0.0;
        description = "Received address is listed in ZEN PBL (ISP list)";
        one_shot = true;
}
```
([snippet source](https://github.com/rspamd/rspamd/blob/master/conf/scores.d/rbl_group.conf#L113-L116))

Talking about [`RCVD_UNAUTH_PBL`](https://github.com/rspamd/rspamd/blob/master/conf/composites.conf#L89-L94), looking up Received IPs against PBL allows us to detect open relays: If a message was received from a dial-up IP, but was not authenticated (no `ESMTP(S)A`), this might indicate a system that accepts mails from any destination.

Sorry for any incompetence, but I do not see why the current procedure of looking up Received IPs against RBL is problematic. Besides, this pull request is about something different. :-)

heraklit256:
Sorry, it must say source, not destination. :-\

heraklit256:
Please let me know if there are more concerns against this... :-) Thanks!

smfreegard:
There have been multiple people saying that scanning the Received headers with something that includes the PBL is a terrible idea and is completely against it's intended purpose.   You haven't been listening.

I've had this argument several times over the last few days (I recently took over running a new DNSBL) - the ultimate goal of the Policy list is to list *every IP that isn't a mail server*.   Think about that when you argue the case for checking every Received hop against it.

Checking non-policy lists against each Received header - I have absolutely no problem with, provided the DNSBL operator only lists based on *recent* trap hits.   However if you use the public mirrors of Spamhaus, the SBL contains what they call *escalation* listings.   Which means they list network CIDRs to try and strongarm bad ISPs to clean-up their act because they've seen a lot of listings over a long period.   It's admirable, but futile and *could* cause false-positives if applied to all the intermediate hops in the Received headers.

So - If you changed this to use sbl-xbl.spamhaus.org instead of zen, and we did a load of testing on that then it *might* get accepted, but in it's present form, it's never going to get accepted.

heraklit256:
@smfreegard: This is about looking up IP addresses resolved from _domains_ in a message, _NOT_ Received IPs. You did not look at the commit.

I understand looking up received IP addresses against an RBL might be problematic, and for most points, I agree. However, this is a different topic here, and I do not get why my commit causes more trouble than we have had before.

Number of DNS queries will not increase, as no additional RBL/URIBL is called. All I want is to improve detecting a domain resolving to IP addresses known for abusive behaviour (XBL) or belonging to a dial-up range (PBL).

In case of RBL issues, please refer to https://github.com/rspamd/rspamd/blob/master/conf/modules.d/rbl.conf and https://github.com/rspamd/rspamd/blob/master/conf/scores.d/rbl_group.conf . Thank you.

smfreegard:
> This is about looking up IP addresses resolved from _domains_ in a message

Sorry - I had a particularly late night last night and missed that point.

> However, this is a different topic here, and I do not get why my commit causes more trouble than we have had before.

In some ways this is worse - you're taking what was designed as an Email Policy List and now try to apply it to IPs running a web server.

If we started seeing botnets hosting their own web servers for the purposes of hosting spam sites, then this might be useful, but this doesn't happen because they wouldn't be able to do it reliably enough (because of NAT), so they don't currently bother.   That could always change though.

The point here is that URIBL_SBL is already a complicated rule for people to understand, and several people on the IRC channel wanted to get rid of it recently.   I can see the argument for adding URIBL_XBL, but not PBL.

heraklit256:
@smfreegard: No problem, I am glad things now cleared up. :-)

You are right, some of my comments are confusing as they refer to the "Received IPs RBL lookup" problematic. Sorry about that.

According to [Spamhaus](https://www.spamhaus.org/xbl/), XBL also lists IP addresses of compromised devices:

> The Spamhaus Exploits Block List (XBL) is a realtime database of IP addresses of hijacked PCs infected by illegal 3rd party exploits, including open proxies (HTTP, socks, AnalogX, wingate, etc), worms/viruses with built-in spam engines, and other types of trojan-horse exploits.

To conclude, my intention of the commit is as follows:
- If a domain resolves to an IP listed at SBL, it might be used for spamming. Visiting such a domain is probably not desired by users and admins.
- If a domain resolves to an IP listed at XBL, it might be compromised and even spread malware. (Botnet controllers are also listed here, but they will mostly host C&C domains.) Visiting such a domain is probably not desired by users and admins, too.
- If a domain resolves to an IP listed at DROP, it is likely to be operated by cryber criminals (Don't Route Or Peer) and there is little chance for being legitimate.
- Domains resolving to an IP listed at PBL may be more common for private usage, but are kind of suspicious otherwise (a company will host its stuff there).

What do you think about:
1. Lowering the score for `URIBL_PBL` to 0.01, e.g. be less penalising here?
2. Change rule description in order to make clear what these are about (suggestions welcome)?
3. Lower the score for `URIBL_SBL` & co. in a dedicated issue? (I was unaware of user complaints here, as I do not participate in IRC.)

Is there anything I forgot?

heraklit256:
Could anybody have a look at the updated rule descriptions?

I think they are more clear than the version before, but double-checking seems to be better here.

Thank you.

moisseev:
I would write "message body" instead of "mail" as it is even more clear.

heraklit256:
Here you go...

vstakhov:
LGTM, thanks for the useful addition!

PR #2632, opened 2018-11-14
fix return code of rspamc

apreiml:
It'll set the return code to EXIT_FAILURE if one of the parties (rspamd side or sub process) or both are failing.


vstakhov:
Yes, that looks better, thanks.

PR #2644, opened 2018-11-23
[Minor] Fix bayes_expiry disabling

moisseev:


src/plugins/lua/bayes_expiry.lua
```diff
@@ -72,7 +72,7 @@ local function check_redis_classifier(cls, cfg)
       end
     end
 
-    if not symbol_spam or not symbol_ham or not type(expiry) == 'number' then
```
vstakhov: What's different here?

moisseev: `not` has precedence over `==`, so `not type(expiry)` is always boolean `false`

vstakhov: Oh, blame me then!

PR #2709, opened 2019-01-13
add HAS_X_TICKET_ID rule

heraklit256:
Add a Lua rule for detecting the presence of a `X-Ticket-Id` header.

This header frequently appears in autogenerated ticket messages, as well as some private health insurance spam.


vstakhov:
I think that should be a regexp rule, not Lua rule.

heraklit256:
Closing this due to lack of interest.

PR #2710, opened 2019-01-13
add HAS_X_REPORT_ABUSE rule

heraklit256:
Add Lua rule for detecting `X-Report-Abuse` header, wich is necessary to build the composite rule.

The `SPAMMY_SNOWSHOE_PRIVHEALTH` composite rule matches on private health insurance spam, usually in German, which I am observing for some weeks. These messages always contain a `List-Unsubscribe` header, a mentioned `X-Report-Abuse` header, a `Reply-To` header, have a distinguished name for the sender, but not for the recipient.

Sometimes, they are relayed through GMail.

In order not to cause False Positives, the rule has only assigned a score of `2.0`.


andryyy:
I don’t think that’s a composite that should be added by default.

andryyy:
Too complicated?

I can see this being triggered too often: "HAS_X_REPORT_ABUSE & HAS_LIST_UNSUB & HAS_REPLYTO & FROM_HAS_DN & TO_DN_NONE";

Another idea: score it 0 and let the individual admin decide.

heraklit256:
Adjusted pull request title...

vstakhov:
Same here: it should be a regexp rule. I also don't see any use in adding of 0 weight symbols to the vanilla config.

heraklit256:
Here you go... :-)

heraklit256:
Closing this due to lack of interest.

PR #2730, opened 2019-01-24
Produce DMARC report email body with CRLF

analogic:
We are using CRLF everywhere but not in template itself, this PR should fix it. See https://cr.yp.to/docs/smtplf.html


vstakhov:
Good catch, thank you!

Issue #2147, opened 2018-04-04
Add ed25519-sha256 DKIM alogrithm


bjoe2k4:
### Classification (Please choose *one* option):

* [x] Enhancement

### Additional information:
- should have low priority as not yet final rfc.
- Current progress: https://datatracker.ietf.org/doc/draft-ietf-dcrup-dkim-crypto/
- rspamd already has some ed25519 implementation, not sure if it can be reused?
- if implemented, there should be an option to include rsa signature as well (double signing).

vstakhov:
They use slightly different ed25519 AFAIR but I might be wrong.

andreasschulze:
I'm running a ed25519 capable signer and verifier if you're interested in interop tests one day..

vstakhov:
It is added now.

bjoe2k4:
Is it possible to sign with both rsa-sha256 and ed25519-sha256? Afaict, there is option to advise rspamd to do double signing.

vstakhov:
Signing supports RSA only so far.

bjoe2k4:
I think there is something wrong with the current implementation (or just the tests are wrong). See my comment in 15a5ad7fa3b877472b5ee7e14883623763eb2bed.

vstakhov:
It contradicts with 4.1 in the same document then.

andreasschulze:
@bjoe2k4 seeb to be correct
https://tools.ietf.org/html/draft-ietf-dcrup-dkim-crypto-13#section-4.1 define sig-a-tag-k =/ "ed25519"
https://tools.ietf.org/html/rfc6376#section-3.5 define sig-a-tag-alg  = sig-a-tag-k "-" sig-a-tag-h
so the a=ed25519-sha256, not only a=ed25519

bjoe2k4:
Now that RFC 8463 is out of the door, there should be an easy way to sign with both ed25519 and rsa. :->

PR #2739, opened 2019-01-31
Add ed25519 and multiple signature support

McKayJT:
This resolves issue #2147

This set of patches adds support for ed25519 signatures, multiple signatures on a single message, and adds some tests for the new functionality. I don't expect it to get merged without more changes, but I'm opening this request to start the conversation.

Limitations:
* multiple signatures only supported when using keys from files
* multiple signatures only supported using rspamc/http, not milter

Flaws:
* configuration still a bit ugly
* exposes a low-level ed25519 primative
* end-users need to expect a scalar or array from dkim-signature in the json

### DIFF
--- a/conf/modules.d/dkim_signing.conf
+++ b/conf/modules.d/dkim_signing.conf
@@ -58,10 +58,17 @@ dkim_signing {
   # Domain specific settings
   #domain {
   #  example.com {
-  #    # Private key path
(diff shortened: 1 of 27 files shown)


McKayJT:
I can rename the debug function. I didn't expect to keep it, but it got
useful.

For the selectors, right now there are also some global options that are
returned as part of that as well. Would you just want to flatten everything?

PR #2771, opened 2019-02-28
[Feature] Export url visible part to lua and add new url flag

miecio45:
Hello everyone!
I crafted some code allowing me to export visible part of each link to lua modules for further processing. I have no idea if there is other way to get such information there.
I added also new flag for url structure which inform whether visible part is something looking similar to url address. Difference between my flag and PHISHED is that, my flag does not require visible url to be in different domain.
I'll be very happy to see Your comments.

Regards!

### DIFF
--- a/src/libserver/html.c
+++ b/src/libserver/html.c
@@ -2373,6 +2373,12 @@ rspamd_html_check_displayed_url (rspamd_mempool_t *pool,
 		return;
 	}
 
+	gint visible_part_len = dest->len - href_offset;
+	url->visible_part = rspamd_mempool_alloc0(pool, visible_part_len +1);
+	url->visible_partlen = visible_part_len;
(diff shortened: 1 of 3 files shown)


vstakhov:
There are multiple issues with this PR.
1) There are clearly couple of the memory leaks there
2) It won't actually save the visible part but everything pretended to be visible - e.g. `<a href=foo><span><p>bar</span></p></a>` will come to `\r\nbar\r\n` so whitespaces cleanup is at least needed.

miecio45:
Hello! Thanks for your remarks. I fixed memory leaks (at least I hope so) and add some whitespace processing. I will be glad to see Your opinion

Regards!

vstakhov:
Merged, thanks! Please use `[Minor]` tag for minor changes and features in the commit messages as it simplifies ChangeLog generation in further.

PR #2788, opened 2019-03-13
[Minor] Add some lua string utils

miecio45:
Hello,
I created two functions useful for string processing in my use case, have no idea If you are interested in adding them to upstream. In any case I will be very glad to see Your comments.
 
Those two functions are:
* `util.is_utf_outside_range` for detecting if selected string contains only characters
  from specyfic UTF range
* `util.get_string_stats` for counting number of digits and letters in string. This function may be extended in the future


src/lua/lua_util.c
```diff
@@ -2470,6 +2489,83 @@ lua_util_is_utf_spoofed (lua_State *L)
...
+	const gchar *string_to_check = lua_tolstring (L, 1, &len_of_string);
+	guint32 range_start = lua_tointeger (L, 2);
+	guint32 range_end = lua_tointeger (L, 3);
+
+	USpoofChecker *spc_sgl;
+	USet * allowed_chars;
+	UErrorCode uc_err = U_ZERO_ERROR;
+
+	if (string_to_check) {
+		spc_sgl = uspoof_open (&uc_err);
```
vstakhov: This can be quite expensive operation. Have you tested performance?

src/lua/lua_util.c
```diff
@@ -2470,6 +2489,83 @@ lua_util_is_utf_spoofed (lua_State *L)
...
+			uspoof_close(spc_sgl);
+			return 1;
+		}
+
+		allowed_chars = uset_openEmpty();
+		uset_addRange(allowed_chars, range_start, range_end);
+		uspoof_setAllowedChars(spc_sgl, allowed_chars, &uc_err);
+
+		uspoof_setChecks (spc_sgl,
+			USPOOF_CHAR_LIMIT | USPOOF_ANY_CASE, &uc_err);
```
vstakhov: Don't we need to check `uc_err`?

miecio45: Sure thing, I'll fix it.

src/lua/lua_util.c
miecio45: I have checked performance and You are right It's 4times slower than similar `util.is_utf_spoofed`. For 1M calls I have 1.3s runtime for `util.is_utf_outside_range` vs 0.3 for `util.is_utf_spoofed`. It's still enough for my purpose.

Problem is that I have no clue how to skip this object initialization on every call and allow customer to use different call parameters at the same time. I think that I cannot use approach from `util.is_utf_spoofed` (static variable). Maybe I should create lua module for this purpose? It will limit the scope of initialization to once per task or maybe You have better idea?

vstakhov: Probably, create a static hash table (or `rspamd_lru_hash_t`) that would store some certain amount of entries to speed up generic calls. I suppose that the parameters set would be strictly defined anyway.

miecio45:
So I used `rspamd_lru_hash_t` as a cache for my validators and its really much more efficient. I added more error handling and created basic tests. As usual any comments are welcome, I hope that I did not create any leak.
Regards!

src/lua/lua_util.c
```diff
@@ -2470,6 +2496,107 @@ lua_util_is_utf_spoofed (lua_State *L)
...
+	gsize len_of_string;
+	gint ret;
+	const gchar *string_to_check = lua_tolstring (L, 1, &len_of_string);
+	guint32 range_start = lua_tointeger (L, 2);
+	guint32 range_end = lua_tointeger (L, 3);
+
+	static rspamd_lru_hash_t *validators;
+
+	if (validators == NULL) {
+		validators = rspamd_lru_hash_new(16, g_free, (GDestroyNotify)uspoof_close);
```
vstakhov: You need to use `rspamd_lru_hash_new_full` with `g_int64_hash` and `g_int64_equal` here as you are using 64 bit integers and not strings as it is implied in a simplified version.

src/lua/lua_util.c
```diff
@@ -2470,6 +2496,107 @@ lua_util_is_utf_spoofed (lua_State *L)
...
+	}
+
+	if (string_to_check) {
+		guint64 hash_key = (guint64)range_end << 32 || range_start;
+
+		USpoofChecker *validator = rspamd_lru_hash_lookup(validators, &hash_key, time(NULL));
```
vstakhov: You don't have a TTL there, so you can save a syscall (even VDSO one) by replacing call to `time(NULL)` to a plain `0` here.

src/lua/lua_util.c
```diff
@@ -2470,6 +2496,107 @@ lua_util_is_utf_spoofed (lua_State *L)
...
+
+			uspoof_setChecks (validator,
+				USPOOF_CHAR_LIMIT | USPOOF_ANY_CASE, &uc_err);
+
+			uset_close(allowed_chars);
+
+			if (uc_err != U_ZERO_ERROR) {
+				msg_err ("Cannot configure uspoof: %s", u_errorName (uc_err));
+				lua_pushboolean (L, false);
+				uspoof_close(validator);
```
vstakhov: Hash key is leaked here.

src/lua/lua_util.c
```diff
@@ -2470,6 +2496,107 @@ lua_util_is_utf_spoofed (lua_State *L)
...
+		UErrorCode uc_err = U_ZERO_ERROR;
+
+		if (validator == NULL) {
+			USet * allowed_chars;
+			guint64 * creation_hash_key = g_malloc(sizeof(guint64));
+			*creation_hash_key = hash_key;
+
+			validator = uspoof_open (&uc_err);
+			if (uc_err != U_ZERO_ERROR) {
+				msg_err ("cannot init spoof checker: %s", u_errorName (uc_err));
```
vstakhov: creation_hash_key is leaked here

miecio45:
Thanks for all Your feedback! I hope I fixed everything, changed hash creation function, free up memory before return, and remove `time` syscalls. Please let me know if this failing build is my fault or just some CI troubles.

vstakhov:
It looks great now, thank you!

PR #2813, opened 2019-03-21
[Minor] Add util.if_utf_mixed_script to lua

miecio45:
Last time I encounter some troubles with using `util.is_utf_spoofed` with only one string because behavior varies between different versions of libicu. Seems that between version 57 and 58 of ICU support for `USPOOF_MIXED_SCRIPT_CONFUSABLE` in single string context was removed. According to [ICU-12823](https://unicode-org.atlassian.net/browse/ICU-12823) and [BUG 12594](http://bugs.icu-project.org/trac/ticket/12549) and [this MR](https://codereview.appspot.com/307020043/diff/1/i18n/uspoof_impl.cpp?context=10&column_width=80)
I personally find this option very useful, especially for processing part of my messages in lua, therefore I implement something similar as a PoC. If You think that it might be useful, I'll create additional tests or maybe we should think about adding this to `lua_util_is_utf_spoofed` ?

### DIFF
--- a/src/lua/lua_util.c
+++ b/src/lua/lua_util.c
@@ -2513,15 +2513,18 @@ lua_util_is_utf_mixed_script(lua_State *L)
 {
 	LUA_TRACE_POINT;
 	gsize len_of_string;
-	const gchar *end, *string_to_check = lua_tolstring (L, 1, &len_of_string);
+	const gchar *string_to_check = lua_tolstring (L, 1, &len_of_string);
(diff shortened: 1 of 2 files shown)


vstakhov:
Hum, can we do that using libicu macros instead? I prefer not to use glib utf8 library as it has proven to be quite slow.

miecio45:
Sure, I can use U8_NEXT from libicu and I've added some small tests.

vstakhov:
Great, thank you!

PR #2845, opened 2019-04-10
add support ES>=6.7 for elasctic module

gauthierc:
Since version 6.7.0 of ElasticSearch, the ingest-geoip plugin is a module integrated into the solution.

Instead of detecting the version of ElasticSearch I added the configuration parameter ingest_module which is false by default for ElasticSearch versions < 6.7.0 and true for versions >= 6.7.0


vstakhov:
That looks good thank you! I would appreciate if you also could send an update to the module's documentation: https://github.com/rspamd/rspamd.com/blob/master/doc/modules/elastic.md

PR #2852, opened 2019-04-16
[Minor] Improve mime_types plugin

spacefreak86:
Hackers sometimes try to hide the filename of an attachment by encoding some characters in HEX (e.g. "attached%2E%62at"). The most e-mail clients would decode this characters and the final filename would be "attached.bat". The mime_types plugin now decodes those characters before determining the file extension.

Additionally, the plugin now uses the higher weighted MIME type, if the detected type differs from the one specified in the Content-Type header, so we are always on the safe side.

### DIFF
--- a/src/plugins/lua/mime_types.lua
+++ b/src/plugins/lua/mime_types.lua
@@ -851,6 +851,9 @@ local function check_mime_type(task)
               fname:sub(1, ch_pos)))
     end
 
+    -- Decode hex encoded characters
+    fname = string.gsub(fname, '%%(%x%x)', function (hex) return string.char(tonumber(hex,16)) end )
+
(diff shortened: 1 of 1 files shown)


vstakhov:
Do you have any samples of such an obfuscation?

spacefreak86:
Yes, I have seen such obfuscations in the wild.
Additionally, you can get some examples from here: https://www.emailsecuritycheck.net/

This is an example of the mentioned header:

Content-Transfer-Encoding: 7bit
Content-Type: application/x-msdownload;
  name*0*="''attached%2E";
  name*1*="%62";
  name*2=at
Content-Disposition: attachment;
  filename*0*="''attached%2E";
  filename*1*="%62";
  filename*2=at

PR #2886, opened 2019-05-04
[Fix] allow PKCS7 signatures to be text/plain, too

heraklit256:
PKCS7 signatures, as inserted by MS Outlook for example, have their own MIME type. However, they are detected as `text/plain` in most cases, so `MIME_BAD_ATTACHMENT` was insered here.

This allows `text/plain` for PKCS7 signatures (`.p7s`), too, thus avoiding this false positive.


vstakhov:
Hum, why they are detected as `text/plain`? Do you have an example?

heraklit256:
Sorry for the late reply.

Example is as follows:
```
--=-_SEM_GW_P_0002_684412DF.5D80BC1A
Content-Type: application/x-pkcs7-signature; name="smime.p7s"
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="smime.p7s"

MIAGCSqGSIb3DQEHAqCAMIACAQExDzANBglghkgBZQMEAgEFADCABgkqhkiG9w0BBwEAAKCAMIIE
[... 28 lines redacted ...]
RGhFSxw0LeSRx7goRC6YIFVHYVJAblqQeLjBDsOlbu6v0ce1Ib1zE/pWCfRk4CdODYd5cfXvNT93
N7dNdjPoZqyOw0BMuQgXYgAAAAAAAA==

--=-_SEM_GW_P_0002_684412DF.5D80BC1A--
```
Since the signature is declared as `application/x-pkcs7-signature`, but detected as `text/plain`, rspamd inserts `MIME_BAD_ATTACHMENT` here.

PR #2887, opened 2019-05-04
small DKIM/ARC signature housekeeping

heraklit256:
This introduces three minor patches for which individual pull requests seemed to be an over-kill:

1. Sign `List-Help` header according to [RFC 6376, section 5.4.1](https://tools.ietf.org/html/rfc6376#section-5.4.1).
2. Do not oversign `Sender` header, as mailing lists might introduce a new one (see [RFC 6376, section B.2.3](https://tools.ietf.org/html/rfc6376#appendix-B.2.3) for details). This improves interoperability, but might open up a security vulnerability in case MUAs confuse `Sender` and `From` - which would be a more general flaw DKIM/ARC signing cannot help against.
3. Sort the list of headers to be signed to improve readability.


vstakhov:
For oversigning I think I've found somehow better way to do oversigning control. It is now possible to oversign **existing** only headers and ignore it if a header is missing. It is done by changing `(o)` to `(x)`. Probably, we should consider switching some of the headers to this mode.

heraklit256:
That's great.

`Sender` might be a good candidate for this - do you have some other headers in mind?

heraklit256:
@vstakhov: Some of the changes commit e0d31efc8573ad2b6c5b262c22801b5709945893 seem to be risky to me:
1. Each message must have a `Date`, `Subject` and a `Message-ID` header. I think we should oversign them in any case.
2. `MIME-Version`, `Content-Type` and `Content-Transfer-Encoding` seem to be dangerous: If they are not present, and we will not sign them, an attacker might [insert faked content displayed by the MUA](https://noxxi.de/research/breaking-dkim-on-purpose-and-by-chance.html).

What do you think about these? I am fine with any other headers the patch changes.

Thank you.

vstakhov:
1. I'm not quite sure about `Message-Id` - it is often inserted by MTA. Subject might, again, be missing intentionally. I have the same concerns about `Date`

2. You have misread the original article: this attack is possible when there are more than zero `Content-Type`/CTE headers and they are not oversigned. Rspamd will oversign them if they are presented destroying the attack surface.

In general, DKIM standard strongly suggests that signer should not sign/oversign **missing** headers.

heraklit256:
Hi @vstakhov , sorry for being so sloppy on relpy.

My objection on 2. was: An attacker might insert a fake MIME content which he points the MUA to by using `Content-Type` header. In case that header was not present before (and thus not oversigned), DKIM signature might still be valid, if the body signature is only applied to the original body part.

But this is more like a field research question. I am fine with your changes.

vstakhov:
Nonsense: there is a `bh` tag in DKIM signature so altering body literally means that you need to find a collision on sha256 which is impossible so far. Body signature has nothing about MIME btw, so I cannot imagine any attack on that.

PR #2915, opened 2019-06-04
libutil: Fix parsing of address "IPv6:::1"

gollux:
Dropping of leading colon must be performed after the "IPv6:" prefix
is stripped.

Also fixed a couple of wrong comments.


vstakhov:
Thanks! Could you please add the test case here: https://github.com/rspamd/rspamd/blob/master/test/lua/unit/addr.lua?

PR #2916, opened 2019-06-05
[Minor] Comment ragel bug workaround

citrin:
It is a good idea to comment code added for no obvious reason to fix a very rare case (which in some time will be not actual anymore).


vstakhov:
I think the original patch is not correct. I will add some unit tests...

PR #2947, opened 2019-06-27
Add rspamd hostname to ES fields

ol3k:
Add rspamd hostname field to be able to filter in ES for different rspamd servers


src/plugins/lua/elastic.lua
```diff
@@ -146,6 +146,7 @@ local function get_general_metadata(task)
   r.user = task:get_user() or 'unknown'
   r.qid = task:get_queue_id() or 'unknown'
   r.action = task:get_metric_action('default')
+  r.rspamd_server = util.get_hostname() or 'unknown'
```
vstakhov: It is better to cache hostname in the module scope I suppose to avoid calling this method on each task.

PR #2954, opened 2019-06-28
Add MIME_BASE64_TEXT_BOGUS rule

smfreegard:


rules/headers_checks.lua
```diff
@@ -1064,6 +1064,26 @@ rspamd_config.MIME_BASE64_TEXT = {
...
+      if (check_for_base64_text(mimepart) and not part:has_8bit()) then
+        return true
+      end
+    end
+    return false
+  end,
+  description = 'Has text part encoded in base64 that does not contain any 8bit characters',
+  score = 1.0,
+  group = 'headers',
+  type = 'mime',
```
vstakhov: What is `type` here?

rules/headers_checks.lua
```diff
@@ -1064,6 +1064,26 @@ rspamd_config.MIME_BASE64_TEXT = {
...
+      local mimepart = part:get_mimepart();
+      if (check_for_base64_text(mimepart) and not part:has_8bit()) then
+        return true
+      end
+    end
+    return false
+  end,
+  description = 'Has text part encoded in base64 that does not contain any 8bit characters',
+  score = 1.0,
+  group = 'headers',
```
vstakhov: I suppose that the groups should be `mime`.

rules/headers_checks.lua
smfreegard: Good question - I copied and pasted the definitions from MIME_BASE64_TEXT and then modified the bits I needed.   All of the rules in that file appear to have a type = 'mime' attribute.

Do you want me to remove them all?

vstakhov: Then those rules are also incorrect. However, group `header` makes sense in case of `INVALID_FROM_8BIT` but not `MIME_BASE64_TEXT`. Type `mime` doesn't make any sense for anything.

PR #2961, opened 2019-07-04
Support List-Id as mailing-list attribute (RFC2919) to suppress REPLY…

### LINKED ISSUE #2960: [Feature] identify mailing lists according to rfc2919 via List-Id
## Summary

REPLYTO_EQ_TO_ADDR does not support sufficient mailing-list attributes. List-Id (RFC2919) is missing.

wrosenauer:
…TO_EQ_TO_ADDR (fixes #2960)

### DIFF
--- a/rules/headers_checks.lua
+++ b/rules/headers_checks.lua
@@ -238,6 +238,7 @@ local check_replyto_id = rspamd_config:register_symbol({
               if (not (task:get_header('List-Unsubscribe') or
                   task:get_header('X-To-Get-Off-This-List') or
                   task:get_header('X-List') or
+                  task:get_header('List-Id') or
                   task:get_header('Auto-Submitted')))
               then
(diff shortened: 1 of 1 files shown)


vstakhov:
We should probably depend on `MAILLIST` symbol instead...

wrosenauer:
I somewhat agree but in that case I don't know how to do it.
Also is MAILLIST disabled by default? I do not see that symbol on the mailing list mails I typically receive while I would expect so.

vstakhov:
It is enabled by default and seems to be working. I see lots of that in my messages, e.g. `MAILLIST(-0.10)[mailman];`

wrosenauer:
Opened a bug #2964 (probably not really a bug since mlmmj seems quite flexible what headers to include and therefore it can vary quite a lot what is done outside.

vstakhov:
This is implemented in the master branch in a different way. I'm closing this PR.

PR #2962, opened 2019-07-04
[Fix] Fix fuzzy image score calculation

moisseev:


vstakhov:
I also think about adding some new params:

- `has_text` - has at least one non-empty text part
- `has_significant_text` - has text parts with enough content (e.g. > 20 words)

So if we have no text or at least no significant text, we can penalise image significantly. If we have seen significant amount of text, then probably we should apply the previous scoring.

What do you think?

moisseev:
It's a good idea.
What do you mean by `has_text`? А `text part that has visible text` or just a `non-empty text part`? It can be like this:
```html
--8386306368bc93e03c3970cfbb69b11d82ac
Content-Type: text/html; charset="utf-8"
Content-Transfer-Encoding: quoted-printable

<HTML><HEAD>
<META http-equiv=3D"Content-Type" content=3D"text/html; charset=3Dutf-8">
</HEAD>
<BODY bgColor=3D#ffffff>
<DIV align=3Dleft><FONT size=3D2 face=3DArial><IMG border=3D0 hspace=3D0 =
alt=3D""=20
src=3D"cid:836e348f68a92684a8062d1bedfc7@i30.es"></FONT></DIV></BODY></HT=
ML>

--8386306368bc93e03c3970cfbb69b11d82ac--
```

vstakhov:
The part that has visible text content of course. And the `words` alike fields in `struct rspamd_text_part` are responsible for that. Or, probably, `utf_stripped_content`.

vstakhov:
I think we can close this PR now, cannot we?

PR #2975, opened 2019-07-12
[Minor] fix lua [[...]] is deprecated near '['

ibqn:
This minor fix resolves a deprecation warning which appears when running
```shell
$ rspamadm pw
cannot execute lua script /usr/local/share/rspamd/lualib/rspamadm//rescore.lua: /usr/local/share/rspamd/lualib/rspamadm//rescore.lua:301: nesting
 of [[...]] is deprecated near '['
Enter passphrase:
 ```


vstakhov:
This tool needs to be rewritten with kann :\

PR #3013, opened 2019-08-23
[Fix] Use random string as default monitored address

moisseev:
as facebook.com is listed on dwl.dnswl.org


vstakhov:
As far as I remember, we already skip monitoring checks for whitelists. Probably, black lists should use `facebook` marker and white lists should use some random string indeed.

PR #3016, opened 2019-08-23
Actually set sentinel_masters_pattern.

### LINKED ISSUE #3015: [BUG] sentinel_masters_pattern ignored
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:

**The Redis option sentinel_masters_pattern does not work***

Upon setting the Redis option 'sentinel_masters_pattern' the list of Redis master-servers should be filtered to only contain servers that match the pattern. This does not happen, all servers are always included. It seems that redis_params.sentinel_masters_pattern is never set.

CAPSLOCK2000:
Fixes #3015

### DIFF
--- a/lualib/lua_redis.lua
+++ b/lualib/lua_redis.lua
@@ -317,6 +317,11 @@ local function process_redis_opts(options, redis_params)
   if not redis_params.sentinels and options.sentinels then
     redis_params.sentinels = options.sentinels
   end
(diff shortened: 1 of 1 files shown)


CAPSLOCK2000:
TBH I'm not really sure I understand the difference between params and options, but the patch works for me.

Issue #2703, opened 2019-01-06
Reasonable default settings for ARC


prehor:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [ ] Ordinary bug
* [ ] Feature
* [x] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
```
Rspamd daemon version 1.8.3
```

### Description (Please provide a descriptive summary of the issue):
The `ARC` module inherits its default settings from the `DKIM Signing` module, but they usually has to be overwritten because the `DKIM Signing` module's default settings are reasonable for signing outgoing e-mails while the `ARC` module should sign incoming e-mails. More reasonable default settings for `ARC` module are:
```bash
# Sign only external incoming e-mails
auth_only = false;
sign_local = false;
sign_inbound = true;

# Use envelope recipient's domain for signing
use_domain = "recipient";

# Ignore header/from/user mismatches
allow_hdrfrom_mismatch = true;
allow_hdrfrom_multiple = true;
allow_username_mismatch = true;
```

If pull request #2657 is merged, the server domain (`hostname -d`) could be used as the default value for `use_domain`.

prehor:
Hi @vstakhov,

I know that the ARC should only sign the forwarded emails, but I do not think the existing ARC module configuration options in rspamd allow it.

In a simple set-up with all-in-one postfix and rspamd milter, rspamd is called before envelope recipients expansion, so rspamd is not able to find out whether the email will be delivered locally or forwarded, or even both at once. In this case, however, it is able to recognize that it is an external incoming email to be delivered to a local domain and to create an ARC signature. The ARC signature will not only have forwarded emails, but also emails delivered to local mailboxes, allowing users to forward email sent to their mailbox on their own. The above default values for the ARC module are suitable for this (I think often used) rspamd deployment.

[...]

PR #3019, opened 2019-08-27
[Config] Reasonable default settings for ARC

klausenbusk:
Fix: #2703

---
cc @prehor

I don't think `allow_username_mismatch = true` is needed as we only sign inbound mails. `allow_username_mismatch` only affects authenticated users as far as I understand.


conf/modules.d/arc.conf
```diff
@@ -26,25 +26,27 @@ arc {
   # If false, messages with empty envelope from are not signed
   allow_envfrom_empty = true;
   # If true, envelope/header domain mismatch is ignored
-  allow_hdrfrom_mismatch = false;
+  allow_hdrfrom_mismatch = true;
   # If true, multiple from headers are allowed (but only first is used)
-  allow_hdrfrom_multiple = false;
+  allow_hdrfrom_multiple = true;
```
vstakhov: Why?

klausenbusk:
You correct, that doesn't make sense. I misinterpreted the meaning of that 
option.


I will fix it when I get home.

conf/modules.d/arc.conf
klausenbusk: You correct, that doesn't make sense. I misinterpreted the meaning of that
option.

Fixed.

conf/modules.d/arc.conf
```diff
@@ -26,25 +26,27 @@ arc {
...
   # If true, envelope/header domain mismatch is ignored
-  allow_hdrfrom_mismatch = false;
+  allow_hdrfrom_mismatch = true;
   # If true, multiple from headers are allowed (but only first is used)
   allow_hdrfrom_multiple = false;
   # If true, username does not need to contain matching domain
   allow_username_mismatch = false;
   # If false, messages from authenticated users are not selected for signing
-  auth_only = true;
+  auth_only = false;
```
klausenbusk: I don't think this make any difference as we don't know if the mail is from a authenticated user (ex: `user1@example.com` -> `user2@example.com`),at least that what my testing shows.

klausenbusk: Never mind, it was a configuration error.

whataboutpereira:
These defaults seem effective in getting incoming mail ARC signed. I do hope the config will get reasonable defaults.

prehor:
> I don't think `allow_username_mismatch = true` is needed as we only sign inbound mails. `allow_username_mismatch` only affects authenticated users as far as I understand.

You are right, it is not necessary. I use it in `dkim_signing.conf` and blindly copied it to `arc.conf`.

andryyy:
What do you think, @vstakhov

Issue #968, opened 2016-09-20
Feature request: include p0f fingerprints 


croessner:
Hi,

it would be awesome to have p0f included to detect certain operating systems like Windows XP, 7 and 8. Unfortunately all Microsoft OSes appear as Windows 7 or 8, a decision cam not be done hard. Therefore rspamd would be ideal to just add scores.

p0f could help to fight better against zombie PCs.

vstakhov:
In principle, such a plugin can be written using [lua tcp client](https://rspamd.com/doc/lua/tcp.html) and pack/unpack routines from `rspamd_util`.

BTW, your MTA seems to have broken IPv6 support, so my messages are rejected: 

```
host mx.roessner-net.de[2a05:bec0::134:255:226:247] said:
    550-5.7.25 Client host rejected: cannot find your hostname,
    [2a01:4f8:190:43b5::99] 550 5.7.25 For assistance, please provide the
    following information in your problem report: time (Sep 20 11:35:56),
    client (2a01:4f8:190:43b5::99:55702) and server (mx.roessner-net.de). (in
    reply to RCPT TO command)
```

croessner:
@vstakhov thanks for the MX reject report. Even using a local unbound resolver and a second resolver as backup on a physical machine, it sometimes can not directly resolve stuff. I changed from 550 to 450. Feel free to try sending again please 😊

heraklit256:
I would be interested in p0f support, too.

vstakhov:
That looks very good! I just think this should be a part of the lua_scanners framework to be one of the `external_services` plugins type. I'm also not quite sure about caching in Redis in this case: is it anyhow useful?

denpamusic:
Thank you! I'll thought about making it part of that framework but as I created this plugin for myself at first, I wasn't sure how to do it with "plugins.d" mechanism. If you're interested I can redo this as PR for rspamd using scanners framework.

In case of small to medium volume system, working without cache should fine, but on large volume I'm afraid of overloading p0f with requests, since p0f can't really be scaled, doesn't feature any internal caching and by default has a limit of 20 simultaneous connections. It can be risen with -S flag, but author recommends not going overboard with it. There's possibility to make cache optional though, so users decide if it's needed based on their volume.

vstakhov:
It is now implemented, thanks to @denpamusic !

PR #3037, opened 2019-09-15
[Feature] Add p0f scanner

denpamusic:
As per discussion in #968, this PR adds support for passive OS fingerprinting (p0f) as part of lua_scanners framework. It also includes functional tests and optional header routine in milter headers module. Redis cache will be used by this plugin if available but otherwise not required.


denpamusic:
Sorry forgot to chmod +x dummy_p0f...

denpamusic:
Cleaned up config a bit and added test case for working without redis just in case. This should be final commit. Sorry for the inconvenience.

vstakhov:
That's really great work! Thank you, Denis!

denpamusic:
You're welcome! Always glad to learn something new and help great project in process.

conf/modules.d/p0f.conf
```diff
@@ -0,0 +1,46 @@
...
+# '$LOCAL_CONFDIR/override.d/file.conf' - to override the defaults
+#
+# See https://rspamd.com/doc/tutorials/writing_rules.html for details
+
+p0f {
+  # Disable module by default
+  enabled = false;
+
+  # Path to the unix socket that p0f listens on
+  socket = '/tmp/p0f.sock';
```
moisseev: @denpamusic Would you explain the reason of using `/tmp` for unix sockets? Wouldn't it be better to use `/var/run/p0f.sock` instead?

denpamusic: Sorry, I've come from FreeBSD-land where most sockets created in /tmp by default as /var/run in only writable by root, so it's kinda habit of mine... You're welcome to change it of course, if you think that I'll be better suited for most users. This PR is already merged, so you'll need to create new one.

denpamusic: After doing a bit of research I've decided to change default socket location to `/var/run/p0f.sock` as you've recommended. I've submitted #3043 that includes this change as well as fixes a couple other problems I've noticed during plugin usage.

moisseev: Are you running p0f as an unprivileged user? How? AFAIU it needs permission to read from `BPF`.

denpamusic: You're right, but p0f can deal with it with `-u user` flag. Basically it's opens bpf as root and the drops privileges.

PR #3044, opened 2019-09-19
[Fix] Fix lua_ip_equal logic

wsldankers:
rspamd_inet_address_compare() seems to return a <0 / =0 / >0 valua à la strcmp(). Or am I mistaken?


vstakhov:
Yes, you are right about `rspamd_inet_address_compare`. I will check the lua part...

Issue #3054, opened 2019-09-25
[BUG] *lua_redis_push_data: call to callback failed: Parse Error in file "<string>" on line 1, column 11: unknown or unexpected tag;


CAPSLOCK2000:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**lua_redis_push_data: call to callback failed: Parse Error in file "<string>" on line 1, column 11: unknown or unexpected tag;  **

I got the following error

2019-09-21 15:13:26 #30283(controller) lua_redis_push_data: call to callback failed: Parse Error in file "<string>" on line 1, column 11: unknown or unexpected tag; trace: [1]:{[C]:-1 - error [C]}; [2]:{/usr/share/rspamd/lualib/lupa.lua:69 - <unknown> [Lua]}; [3]:{[C]:-1 - match [C]}; [4]:{/usr/share/rspamd/lualib/lupa.lua:732 - jinja_template [Lua]}; [5]:{/usr/share/rspamd/plugins/dmarc.lua:930 - send_report_via_email [Lua]}; [6]:{/usr/share/rspamd/plugins/dmarc.lua:999 - callback [Lua]}; [7]:{/usr/share/rspamd/lualib/lua_redis.lua:985 - <unknown> [Lua]};

Reformatted for readability:
  2019-09-21 15:13:26 #30283(controller) lua_redis_push_data: call to callback failed:
  Parse Error in file "<string>" on line 1, column 11: unknown or unexpected tag;
  trace:
  [1]:{[C]:-1 - error [C]};
  [2]:{/usr/share/rspamd/lualib/lupa.lua:69 - <unknown> [Lua]};
  [3]:{[C]:-1 - match [C]};
  [4]:{/usr/share/rspamd/lualib/lupa.lua:732 - jinja_template [Lua]};
[...]

PR #3055, opened 2019-09-25
rename report_domain -> reportING_domain

CAPSLOCK2000:
Possible fix for #3054 .
I'm unable to trigger the bug so I can't really test this patch.


CAPSLOCK2000:
This patch does not fix #3054 , but I still think it's a valid patch as the variable 'report_domain' does not occur anywhere else (while reporting_domain does).

PR #3058, opened 2019-09-28
Do not print empty [] in X-Spamd-Result.

whataboutpereira:
Clean up the output from X-Spamd-Result headers, going from:

FROM_EQ_ENVFROM(0.00)[];
to
FROM_EQ_ENVFROM(0.00);

### DIFF
--- a/src/plugins/lua/milter_headers.lua
+++ b/src/plugins/lua/milter_headers.lua
@@ -244,9 +244,9 @@ local function milter_headers(task)
       string.format('%.2f', common['metric_score'][1]), ' / ', string.format('%.2f', common['metric_score'][2]), ']'
     }))
     for _, s in ipairs(common.symbols) do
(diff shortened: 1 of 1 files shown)


vstakhov:
What is the purpose of this change? It breaks compatibility and makes the header more difficult to parse.

whataboutpereira:
In truth I thought it’s an oversight and makes the result a little more difficult to read. Rspamc outputs the same without empty brackets, which is cleaner.

vstakhov:
Rspamd is a CLI tool that faces to a user directly. It can however output raw json for some automated parsing. In case of Email headers we want some trade-off between readability and machine processing capability. Hence, I'd prefer the existing variant as it clearly defines a universal standard.

whataboutpereira:
Understandable. :)

PR #3091, opened 2019-10-16
[Fix] Correct EMA time calculations

realbugi:


vstakhov:
Oh, that is a great catch, thank you!

PR #3147, opened 2019-11-11
Fixing multiple classifiers on redis

### LINKED ISSUE #3129: [BUG] Multiple classifiers on redis working incorrectly
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:

**Describe the bug**
When there are multiple bayes classifiers, if `min_learns` for them are set differently and not all them are fulfilled. Then all the classifiers are not working.

### Steps to Reproduce

Setup two classifiers in `local.d/statistic.conf`:
```
classifier "bayes" {
    name = "bayes_one"
    tokenizer {
        name = "osb";
    }
    backend = "redis";
    new_schema = true;
    store_tokens = false;
    signatures = false;
    min_learns = 1;

    statfile {
        symbol = "BAYES_HAM_ONE";
        spam = false;
    }
    statfile {
        symbol = "BAYES_SPAM_ONE";
        spam = true;
    }
}
classifier "bayes" {
    name = "bayes_two"
    tokenizer {
        name = "osb";
    }
    backend = "redis";
    new_schema = true;
    store_tokens = false;
    signatures = false;
    min_learns = 5000;

    statfile {
        symbol = "BAYES_HAM_TWO";
        spam = false;
    }
    statfile {
        symbol = "BAYES_SPAM_TWO";
        spam = true;
    }
}
```

Start the rspamd and learn some messages for both spam and ham:
```
rspamc learn_ham -c bayes_one /etc/rspamd/scores.d
rspamc learn_spam -c bayes_one /etc/rspamd/modules.d

rspamc learn_ham -c bayes_two /etc/rspamd/scores.d
rspamc learn_spam -c bayes_two /etc/rspamd/modules.d
```

Start any file:
```
rspamc /etc/rspamd/common.conf
```

And you will see no bayes symbols are output.

**Expected behavior**
When one classifier's `min_learns` is fulfilled, it starts working right away.

### Versions

Rspamd daemon version 2.1

### Additional Information

It is because that in 
https://github.com/rspamd/rspamd/blob/7a7791926e0ad43ce69d3ba9cc5c73dc8fa9d3a2/src/libstat/stat_process.c#L427-L431

Any error in pulling the tokens from redis for one statfile, all classifiers are stopped.

resec:
Fix #3129, Things may need further work:

- [ ] `rspamc stat` is not reporting the classifier of statfile, which could be enhanced.
- [ ] learnspam and learnham api should report result for each classifier.
- [ ] functional test cases for multiple classifier should be added.
- [ ] doc about how to define multiple classifiers

@vstakhov, please help to review the current implementation, thanks.

### DIFF
--- a/src/libstat/backends/redis_backend.c
+++ b/src/libstat/backends/redis_backend.c
@@ -1425,29 +1425,13 @@ rspamd_redis_parse_classifier_opts (struct redis_stat_ctx *backend,
 
 	elt = ucl_object_lookup (obj, "prefix");
 	if (elt == NULL || ucl_object_type (elt) != UCL_STRING) {
-		gchar *redis_object;
-
 		/* Default non-users statistics */
 		if (backend->enable_users || backend->cbref_user != -1) {
-			redis_object = REDIS_DEFAULT_USERS_OBJECT;
+			backend->redis_object = REDIS_DEFAULT_USERS_OBJECT;
 		}
 		else {
-			redis_object = REDIS_DEFAULT_OBJECT;
+			backend->redis_object = REDIS_DEFAULT_OBJECT;
 		}
-
-		/* Prepend classifier name if defined */
-		elt = ucl_object_lookup (obj, "name");
-		if (elt != NULL && ucl_object_type (elt) == UCL_STRING) {
-			const gchar *cl_name = ucl_object_tostring (elt);
-			gchar *temp;
-			temp = g_malloc (strlen (cl_name) + strlen (redis_object) + 2);
-			strcpy (temp, cl_name);
-			strcat (temp, "_");
-			strcat (temp, redis_object);
-			redis_object = temp;
-		}
-
-		backend->redis_object = redis_object;
 	}
 	else {
 		/* XXX: sanity check */
--- a/src/libstat/backends/redis_backend.c
+++ b/src/libstat/backends/redis_backend.c
@@ -1425,13 +1425,29 @@ rspamd_redis_parse_classifier_opts (struct redis_stat_ctx *backend,
 
 	elt = ucl_object_lookup (obj, "prefix");
 	if (elt == NULL || ucl_object_type (elt) != UCL_STRING) {
+		gchar *redis_object;
+
 		/* Default non-users statistics */
 		if (backend->enable_users || backend->cbref_user != -1) {
-			backend->redis_object = REDIS_DEFAULT_USERS_OBJECT;
+			redis_object = REDIS_DEFAULT_USERS_OBJECT;
 		}
 		else {
-			backend->redis_object = REDIS_DEFAULT_OBJECT;
+			redis_object = REDIS_DEFAULT_OBJECT;
 		}
+
+		/* Prepend classifier name if defined */
+		elt = ucl_object_lookup (obj, "name");
+		if (elt != NULL && ucl_object_type (elt) == UCL_STRING) {
+			const gchar *cl_name = ucl_object_tostring (elt);
+			gchar *temp;
+			temp = g_malloc (strlen (cl_name) + strlen (redis_object) + 2);
+			strcpy (temp, cl_name);
+			strcat (temp, "_");
+			strcat (temp, redis_object);
+			redis_object = temp;
+		}
+
+		backend->redis_object = redis_object;
 	}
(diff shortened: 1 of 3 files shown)


src/libstat/backends/redis_backend.c
```diff
@@ -1428,13 +1425,29 @@ rspamd_redis_parse_classifier_opts (struct redis_stat_ctx *backend,
...
+
+		/* Prepend classifier name if defined */
+		elt = ucl_object_lookup (obj, "name");
+		if (elt != NULL && ucl_object_type (elt) == UCL_STRING) {
+			const gchar *cl_name = ucl_object_tostring (elt);
+			gchar *temp;
+			temp = g_malloc (strlen (cl_name) + strlen (redis_object) + 2);
+			strcpy (temp, cl_name);
+			strcat (temp, "_");
+			strcat (temp, redis_object);
```
vstakhov: This leaks memory and is terribly ugly. Please use mempool and rspamd_snprintf instead.

src/libstat/backends/redis_backend.c
```diff
@@ -1428,13 +1425,29 @@ rspamd_redis_parse_classifier_opts (struct redis_stat_ctx *backend,
...
 		if (backend->enable_users || backend->cbref_user != -1) {
-			backend->redis_object = REDIS_DEFAULT_USERS_OBJECT;
+			redis_object = REDIS_DEFAULT_USERS_OBJECT;
 		}
 		else {
-			backend->redis_object = REDIS_DEFAULT_OBJECT;
+			redis_object = REDIS_DEFAULT_OBJECT;
 		}
+
+		/* Prepend classifier name if defined */
```
vstakhov: This also might break backward compatibility and finally break all existing stats for users. I don't like it...

src/libstat/learn_cache/redis_cache.c
```diff
@@ -292,7 +292,17 @@ rspamd_stat_cache_redis_init (struct rspamd_stat_ctx *ctx,
...
 	else {
-		cache_ctx->redis_object = DEFAULT_REDIS_KEY;
+		gchar *cl_name = st->classifier->cfg->name;
+		if (cl_name) {
+			gchar *redis_object;
+			redis_object = g_malloc (strlen (cl_name) + strlen (DEFAULT_REDIS_KEY) + 2);
+			strcpy (redis_object, cl_name);
+			strcat (redis_object, "_");
+			strcat (redis_object, DEFAULT_REDIS_KEY);
+			cache_ctx->redis_object = redis_object;
```
vstakhov: Same concerns here.

src/libstat/backends/redis_backend.c
```diff
@@ -1325,13 +1325,10 @@ rspamd_redis_connected (redisAsyncContext *c, gpointer r, gpointer priv)
 				}
 			}
 			else {
-				if (!rt->err) {
```
vstakhov: Why is it needed? `runtime` is different for each classifier, so what is the purpose of this change?

resec: This is to avoid such case: any classifier is skipped, it throws out an error, then the result of the learn is errored.  

Ideally, it should be reporting the result for each classifier.

And "skipped" may not be an error, as it's an expected action defined by config.

src/libstat/backends/redis_backend.c
resec: You are right. 
Then the only way to define separate the redis stat data for different classifier is to properly setup `prefix`.
Personally I would like having a new classifier by unique `name`.
But backward compatibility wins.

src/libstat/learn_cache/redis_cache.c
resec: Same backward compatibility problem here.

I will revert the change. We will have to use the `cache_key`.

src/libstat/backends/redis_backend.c
vstakhov: Ok, I see, I do agree with this point.

vstakhov:
I'm not against the proposed default prefix change. It just should not change the behaviour that is enabled by default. We should also make it clear that on upgrade it won't break other users configuration or clearly state that in the upgrade document. I suppose it is still doable.

PR #3153, opened 2019-11-19
Eliminate redis SMEMBERS usage

resec:
In the case of `per_user` model, redis `SMEMBER` may block the redis server for a long time if there are many users.

So to use `SSCAN` instead.

### DIFF
--- a/src/libstat/backends/redis_backend.c
+++ b/src/libstat/backends/redis_backend.c
@@ -98,7 +98,6 @@ struct rspamd_redis_stat_cbdata {
 	redisAsyncContext *redis;
 	ucl_object_t *cur;
 	GPtrArray *cur_keys;
-	guint 
 	struct upstream *selected;
 	guint inflight;
 	gboolean wanna_die;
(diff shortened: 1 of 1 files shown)


src/libstat/backends/redis_backend.c
```diff
@@ -912,30 +920,50 @@ rspamd_redis_stat_keys (redisAsyncContext *c, gpointer r, gpointer priv)
...
-			rspamd_redis_async_cbdata_cleanup (cbdata);
-			redis_elt->cbdata = NULL;
+		if (more) {
+			/* Cleanup the cbdata->cur_keys and re-allowcate */
+			for (i = 0; i < cbdata->cur_keys->len; i ++) {
+				k = g_ptr_array_index (cbdata->cur_keys, i);
+				g_free (k);
+			}
+
+			g_ptr_array_free (cbdata->cur_keys, TRUE);
```
vstakhov: Why do you free these objects? Reallocating memory seems completely useless for me - you can just set its size to zero. I'm even not quite sure that we need to dup keys in `cur_keys` as redisAsyncCommand already copies all stuff internally. Array reallocation is definitely redundant.

resec: Agree, I myself also wonder why we have to store the key into `cur_keys`.

Then just remove the cleanup code, and let the `g_ptr_array_set_size ` in here to do the job:

https://github.com/rspamd/rspamd/blob/fbd2c0e1e3da12793dbf2681feac1244e9ae5485/src/libstat/backends/redis_backend.c#L863

vstakhov:
It seems that this change has introduced a memory leak.

PR #3166, opened 2019-11-26
cmake: remove unused ENABLE_OPENBLAS

arkamar:
Remove `ENABLE_OPENBLAS` option because it is not used to check if the openblas should be used currently, but `ENABLE_BLAS` from `CMakeLists.txt` is.

Hopefully, I didn't missed anything. I believe this commit may avoid future confusion in configuration phase.


vstakhov:
My suggestion is to move this option to cmake/Openblas.cmake. Perhaps, it is good to support both `ENABLE_BLAS` and `ENABLE_OPENBLAS` options and turn on openblas checks if either of two options matches to `ON`. How do you think?

arkamar:
I would stick with one, because it does not raise any questions what is a difference between those two.

vstakhov:
Ok, that sounds reasonable. But I'm still convinced it should be moved to `cmake/OpenBLAS.cmake`. Top level CMakeLists.txt is already way too much complex.

arkamar:
Yes, that make sense. I force-pushed new version of commit.

arkamar:
Thanks!

PR #3187, opened 2019-12-13
Use CMAKE_CXX_FLAGS_{DEBUG,RELEASE} variable for C++

arkamar:
The `CMAKE_CXX_FLAGS_{DEBUG,RELEASE}` variable should be used with C++
rather than `CMAKE_C_FLAGS_{DEBUG,RELEASE}`.


vstakhov:
Good catch, thank you!

PR #3188, opened 2019-12-16
Send Content-Type header in Vault create request

pschichtel:
I noticed this while building a dummy Vault-like backend.


vstakhov:
Good catch, thank you!

PR #3198, opened 2019-12-26
fix NULL pointer use when JIT compiling a regular expression.

doj:
This is fixing a configtest crash of rspamd on Gentoo Linux.


vstakhov:
Good catch, thank you!

PR #3199, opened 2019-12-26
[WIP] Update documentation

cxmplex:


src/lua/lua_task.c
```diff
@@ -200,6 +200,7 @@ end
  */
 LUA_FUNCTION_DEF (task, append_message);
 /***
+ * Consider using extract_specific_urls depending on use case
  * @method task:get_urls([need_emails|list_protos][, need_images])
```
vstakhov: AFAIR, @method should be first here or it won't be matched by doxydown.

src/lua/lua_util.c
```diff
@@ -620,6 +620,14 @@ LUA_FUNCTION_DEF (util, parse_content_type);
  */
 LUA_FUNCTION_DEF (util, mime_header_encode);
 
+/***
+ *  @function util.nkeys(table)
+ * Returns number of keys in table
+ * @param {table} input table
+ * @return int or nil
+ */
+LUA_FUNCTION_DEF (util, nkeys);
```
vstakhov: This is defined in `lua_util` not `rspamd_util` and is documented there AFAIR.

src/lua/lua_config.c
```diff
@@ -195,8 +195,10 @@ rspamd_config:add_map('http://example.com/map', "settings map", process_map)
 LUA_FUNCTION_DEF (config, get_classifier);
 /***
  * @method rspamd_config:register_symbol(table)
+ * Symbols configuration is static and should not be modified on flight
+ * Registering symbols is only guaranteed to be safe on lua plugin load,
+ * not in on_load, periodic or any other type of callbacks
```
vstakhov: I think this should be moved to the bottom of the  documentation block.

Issue #3208, opened 2020-01-11
[BUG] bayes_expiry.lua, WRONGTYPE when encountering RSIG_* instead of RS_* on expiry run


meyerd:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:

When `bayes_expiry` is run, it iterates through `RS*_*`, which also includes the keys for `RSIG_*` when using only one redis instance for all modules. Since `RSIG_*` is not a hash redis value the call to `local values = redis.call('HMGET', key, 'H', 'S')` fails with `WRONGTYPE`.

### Steps to Reproduce

1. Setup rspamd with only one redis instance.
2. Enable `bayes_expiry`
3. wait.

**Expected behavior**
No error in the embedded redis lua script of `bayes_expiry`

### Versions

tested on 1.8.1, but still there in 2.2

meyerd:
... forgot the error message

`lua; bayes_expiry.lua:332: cannot perform expiry step: ERR Error running script (call to f_7560a8b571233af0c51bc24b982427d7da9075f1): @user_script:58: WRONGTYPE Operation against a key holding the wrong kind of value`

PR #3209, opened 2020-01-11
fix bayes expiry WRONGTYPE error #3208

meyerd:
see bug ticket #3208 for issues

### DIFF
--- a/src/plugins/lua/bayes_expiry.lua
+++ b/src/plugins/lua/bayes_expiry.lua
@@ -456,7 +456,7 @@ local function expire_step(cls, ev_base, worker)
   lredis.exec_redis_script(cls.script,
       {ev_base = ev_base, is_write = true},
       redis_step_cb,
-      {'RS*_*', cls.expiry}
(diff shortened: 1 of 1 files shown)


src/plugins/lua/bayes_expiry.lua
```diff
@@ -456,7 +456,7 @@ local function expire_step(cls, ev_base, worker)
   lredis.exec_redis_script(cls.script,
       {ev_base = ev_base, is_write = true},
       redis_step_cb,
-      {'RS*_*', cls.expiry}
```
vstakhov: It will break all per-user bayes tokens with this change.

vstakhov:
I'm sorry but I cannot accept the proposed fix as it causes more issues that it resolves.

vstakhov:
In Redis 6.0, there is `TYPE` option to `SCAN` command but we cannot require this version to be used I'm afraid.

vstakhov:
A different sollution has been chosen, thanks for the report anyway!

PR #3212, opened 2020-01-14
[Minor] bayes_expiry: Fix type check

moisseev:


src/plugins/lua/bayes_expiry.lua
```diff
@@ -259,7 +259,7 @@ local expiry_script = [[
     0,0,0,0,0,0,0,0,0,0,0
 
   for _,key in ipairs(keys) do
-    local t = redis.call('TYPE', key)
```
vstakhov: hum, what is "ok" here?

vstakhov: 127.0.0.1:6379> eval "return redis.call('set','foo','bar')" 0
OK
127.0.0.1:6379> eval "return redis.call('type','foo')" 0
string

vstakhov: Ah, I see:

127.0.0.1:6379> eval "return redis.call('type','foo')['ok']" 0
"string"

vstakhov: https://github.com/antirez/redis/issues/3231

PR #3214, opened 2020-01-16
[Minor] Dkim_signing: correct is_skip_sign logic

Lekensteyn:
If any of "sign_networks", "auth_only", or "sign_local" are disabled,
then it should not automatically proceed with signing if the enabled
conditions all fail. For example, if only the auth_only setting is
enabled, and is_authed is false, then signing should be skipped.

An earlier check luckily prevents this correctness issue from being
exploitable ("ignoring unauthenticated mail"), but fix the logic anyway.


vstakhov:
Ok, I'm now convinced that your fix is valid, thank you!

PR #3217, opened 2020-01-16
[Minor] remove unused examplesdir

cgzones:


vstakhov:
It was used in the past but it is no longer the case. So I assume this should be merged.

PR #3220, opened 2020-01-16
Compiler warnings

cgzones:


src/lua/lua_task.c
```diff
@@ -4861,13 +4861,16 @@ lua_task_get_date (lua_State *L)
 			lua_pushnumber (L, tim);
 		}
 		else {
-			GTimeVal tv;
 			gchar *out;
+			GDateTime *dt_zero = g_date_time_new_from_unix_utc(0);
+			GDateTime *dt_added;
 
-			double_to_tv (tim, &tv);
-			out = g_time_val_to_iso8601 (&tv);
```
vstakhov: Now this causes memory allocation. I think we should go with a different approach.

vstakhov: Could you please remove this change from your PR as the rest looks good to me?

vstakhov:
Thank you!

PR #3227, opened 2020-01-18
Bring Debian packaging up-to-date

Lekensteyn:
Changes:

* [Minor] Debian: Enable Hyperscan and parallel builds
    Debian 9 (Stretch) ships with Hyperscan 4.4.1, Ubuntu 18.04 ships with
    Hyperscan 4.7.0. Follow rspamd's official recommendation and enable
    support for these.
    The --parallel option has been supported since 2009 and works perfectly
    with CMake, so enable it by default.
* [Minor] Debian: add missing sodium dependency
    Required since rspamd 2.0. Even Debian Jessie ships 1.0.0, so add it
    unconditionally.
* [Minor] Debian: remove unused cmake options
    GD (off by default) and TORCH (on by default) were removed in 2.0, the
    examples option (off by default) was removed in 2.2.
* [Minor] Debian: Expand Build-Depends line and sort it
* [Minor] Debian: use non-deprecated libpcre2-dev package
    See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911935
 * [Minor] Debian: remove unnecessary packages
    dh-systemd is a transitional package and moved into debhelper, so remove
    it. This is supported since Debian Stretch. This change has to be
    reversed if Debian Jessie or Ubuntu Xenial have to be supported.

NOTE:
- The Build-Depends sorting change will break the rspamd_build.sh script. It has to be updated to avoid patching for the common case.
- Adding hyperscan breaks the Debian Jessie and Ubuntu 16.04 Xenial build since they do not have Hyperscan. Xenial is still supported until April 2024, Debian Jessie LTS is supported until June 2020. rspamd_build.sh should be modified to disable external hyperscan for Xenial and Jessie.

Tested by building on Debian Stretch with build-essential pre-installed:
```
sudo tee /etc/apt/sources.list.d/backports.list <<<'deb http://deb.debian.org/debian stretch-backports main'
sudo apt update
sudo apt install -t stretch-backports cmake
sudo apt install $(dpkg-checkbuilddeps |& sed 's/.*dependencies: //;s/ ([^)]*)//g;s/ | [^ ]\+//g') --no-install-recommends
dpkg-buildpackage -b -us -j10
```

Then installed it in a fresh Stretch image, it starts at least and there are no obvious errors in /var/log/rspamd/rspamd.log. After installing lintian, it only complained about:
```
user@stretch:~/rspamd$ lintian ../rspamd_1.0.2_amd64.deb 
W: rspamd: binary-without-manpage usr/bin/rspamd_stats
N: 2 tags overridden (2 warnings)
```
If needed, one could use `pod2man utils/rspamd_stats.pl doc/rspamd_stats.1` to generate a manual page. That is not done here however.


Lekensteyn:
For reference, these are the dependencies installed by this packaging on Debian stretch

    libhyperscan4         4.4.1-1
    libluajit-5.1-common  2.0.4+dfsg-1
    libluajit-5.1-2       2.0.4+dfsg-1+b1
    libsodium18           1.0.11-2
    libunwind8            1.1-4.1

With `sudo apt install -t stretch-backports libhyperscan-dev` on the build machine, I obtain libhyperscan5 5.0.0-3~bpo9+1.

PR #3228, opened 2020-01-21
Fix Debian reproducible builds

Lekensteyn:
These patches change the output and locations of some files, but without functional effect. It does however succeed in producing reproducible builds independent of the build directory. The deb package is reproducible once you patch debian/rules to disable LTO (`-DENABLE_LTO=OFF`).

Test under Debian Buster:
```
rm -rf one two && mkdir one two
git archive --prefix=rspamd/ HEAD | tar xC one
git archive --prefix=rspamd/ HEAD | tar xC two
for i in one two; do (cd "$i/rspamd" && time dpkg-buildpackage -b -us -uc -j8); done
# Should only show changes in the rspamd_1.0.2_amd64.buildinfo file
diff -x '*cache' -x '*internal' -x 'Makefile*' -x '*make' -x '*.txt' -x '*.log' -Nur one two
```

I am not entirely sure why LTO breaks reproducibility. Some other projects seems to have no issues with the randomized section and symbol names in object files.


Lekensteyn:
The failing test is not a regression in this PR, it already exists on master since https://github.com/rspamd/rspamd/commit/c4cdc7a35003eafb1baa2a3bf5f1c48d89465e63
```
------------------------------------------------------------------------------
PDF javascript                                                        | FAIL |
'Results for file: /drone/src/test/functional/messages/pdf_js.eml (0.029 seconds)
[Metric: default]
Action: no action
Spam: false
Score: 2.00 / 20.00
Symbol: DATE_IN_PAST (1.00)
Symbol: FROM_EQ_ENVFROM (0.00)
Symbol: FROM_NO_DN (0.00)
Symbol: HAS_ATTACHMENT (0.00)
Symbol: MID_RHS_MATCH_FROM (0.00)
Symbol: RCPT_COUNT_ONE (0.00)[1]
Symbol: RCVD_COUNT_ZERO (0.00)[0]
Symbol: R_DKIM_NA (1.00)
Symbol: TO_DN_NONE (0.00)
Message-ID: 20190126120458.015328@srv.example.com
' does not contain 'PDF_JAVASCRIPT'
------------------------------------------------------------------------------
Cases.280 Rules                                                       | FAIL |
8 critical tests, 7 passed, 1 failed
```

PR #3239, opened 2020-01-25
Two minor bugfixes

spacefreak86:
Dear maintainers

Today I have found two minor issues, one in the force_actions and another in the dmarc plugin.

**force_actions**
The plugin registers a symbol for every configured rule with its name prefixed with 'FORCE_ACTION_'.
The issue is that the prefix is not printed in the log messages which is misleading.

**dmarc**
It appears that the symbol DMARC_BAD_POLICY has been forgotten to register. This causes rspamd to warn about it if someone creates a forced action based on this symbol:
_cannot find dependency on symbol DMARC_BAD_POLICY for symbol FORCE_ACTION_DMARC_BAD_POLICY_

### DIFF
--- a/src/plugins/lua/dmarc.lua
+++ b/src/plugins/lua/dmarc.lua
@@ -1382,6 +1382,13 @@ rspamd_config:register_symbol({
   type = 'virtual'
 })
 rspamd_config:register_symbol({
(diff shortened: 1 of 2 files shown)


vstakhov:
Nice catch, thank you!

PR #3240, opened 2020-01-26
[Feature] Additional attributes for force_actions rules

spacefreak86:
Two additional attributes for the rules of the force_action plugin. I hope you will find this useful.

### Attribute check_local (boolean)
**true** The rule is only enabled for local IPs.
**false** The rule is only enabled for non-local IPs.
If the option is **absent**, the rule is always enabled.

My use-case is to reject ougoing virus mails with a forced action, but let incoming virus mails pass as they get quarantined by a subsequent milter.

### Attribute group (string)
Set the group of the symbol which is registered for the rule.

### DIFF
--- a/src/plugins/lua/force_actions.lua
+++ b/src/plugins/lua/force_actions.lua
@@ -152,6 +152,7 @@ local function configure_module()
         local raction = lua_util.list_to_hash(sett.require_action)
         local honor = lua_util.list_to_hash(sett.honor_action)
         local check_local = sett.check_local
+        local group = sett.group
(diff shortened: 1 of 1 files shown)


vstakhov:
I'm sorry for a long reply - this PR has just disappeared from my scope somehow. Could you please take a look at the functions `config_check_local_or_authed` and `is_skip_local_or_authed` from `lua_util` module? They should do the job in a more universal way.

spacefreak86:
I'm sorry too, had a lot of other things to do the last months. I will take a look at it and reimplement my changes.

PR #3252, opened 2020-02-05
More german stop words

andryyy:


vstakhov:
Hum, we should probably format this file with jq :) For now, diffs are unreadable.

Issue #3156, opened 2019-11-21
[BUG] R_MIXED_CHARSET false positive with czech language


prehor:
### Describe the bug

All e-mails with correctly encoded Czech diacritics in From, To, Subject, or body are incorrectly marked with `R_MIXED_CHARSET` symbol.

It started after updating rspamd from 1.9.4 to 2.0.

### Steps to Reproduce

```bash
cat <<EOF | rspamc | grep R_MIXED_CHARSET
From: =?utf-8?q?Testovac=C3=AD_odes=C3=ADlatel?= <test@example.com>
To: =?utf-8?q?Testovac=C3=AD_p=C5=99=C3=ADjemce?= <test@example.org>
Subject: =?utf-8?q?P=C5=99i=C5=A1el_nov=C3=BD_mail!?=
Date: Mon, 18 Nov 2019 09:31:51 +0100
MIME-Version: 1.0
Content-Type: multipart/alternative;
 boundary="----=_NextPart_29989030.998171461718"

------=_NextPart_29989030.998171461718
Content-Type: text/plain;
 charset="utf-8"
Content-Transfer-Encoding: quoted-printable

Od: info@test.com, P=C5=99edm=C4=9Bt: Od knihtisku k tisk=C3=A1r=
n=C4=9B.
------=_NextPart_29989030.998171461718--
EOF
Symbol: R_MIXED_CHARSET (8.33)[subject]
```

### Versions

```
Rspamd daemon version 2.0
Linux emx1 4.9.0-11-amd64 #1 SMP Debian 4.9.189-3+deb9u1 (2019-09-20) x86_64 GNU/Linux
```

### Additional Information

Probably releated to 4839327.

vstakhov:
Czech is particularly difficult for this rule. Do you have some more samples?

prehor:
Hi @vstakhov, there is an archive of czech FreeBSD mailing list: [czech-language-eml.tar.gz](https://github.com/rspamd/rspamd/files/3901633/czech-language-eml.tar.gz) with a lot of emails.

Disassembler0:
+1 on this. Any quoted-printable (observed on UTF-8 and iso-8859-2 so far) seems to trigger this. It doesn't seem to happen with base64-encoded data.

An example I had for my report before I found out this one already exists (note that the subject is the same, only differently encoded):
```
cat <<EOF | rspamc | grep R_MIXED_CHARSET
Date: Thu, 03 Dec 2019 08:03:23 +0200
[...]

PR #3253, opened 2020-02-05
Add Czech stop words

Disassembler0:
This PR adds Czech stop words as suggested in #3156.

If the diacritics need to be `\u` escaped, please let me know.


vstakhov:
It looks good but the general rule is to avoid words that can clash with other languages. Some of the words could do it, e.g. `co` or `pro`. Could you please check that? Thanks in advance!

Disassembler0:
OK, will try to clean it up a little and add some other common words expected in e-mails (greetings, farewells etc.) Unfortunately the shortest words are the most used. Half of the single letters of alphabet are either prepositions or conjunctions in Czech. I'm starting to understand why the detection so hard. :)

Disassembler0:
Amended. I've removed the clashing words, added a bunch of most frequently used words according to the Czech National Corpus and sprinkled a few words heavily used in mail communication. There are now 280 Czech words (~100 word stems).

I've run a few tests and I'm getting about 5 hits on short and poorly written mails and between 20 and 100 on half-page long mails. And 0 hits for long English, German and French articles borrowed on Wikipedia. :) I think this is the maximum I can do without delving into some heavy linguistic analysis.

pvagner:
@Disassembler0 Shal I do something like this for slovak? What source have you used e.g. hunspell dictionary or something else?
I don't fully understand the matter but I'd be happy to make it work for me and other people writing emails in slovak.

PR #3263, opened 2020-02-12
[Minor] Fix parsing DMARC records containing upper-case letters

spacefreak86:
According to a post on dmarc.org [Link](https://dmarc.org/2016/07/common-problems-with-dmarc-records/),  the string “DMARC” is the only portion of the DMARC policy record that is case-sensitive, but it is not uncommon for people reading the specification to miss this fact.

It seems that this happened when the DMARC plugin was developed and I hereby fixed this issue.


vstakhov:
It looks good, thank you!

PR #3264, opened 2020-02-14
Update packaging versions

Lekensteyn:
Two more changes to reduce the patching necessary in rspamd_build.sh, and to ensure that `dpkg-buildpackage -b -us -uc` produces a package with a recent version number to make upgrading easier.

CC @sbadia (the current Debian maintainer)
___
[Minor] Update Debian and RPM package versions

* Add script that can be used after tagging to update versions.
* Example (for this commit): `./set-version.sh 2.4`
* Clear Debian changelog, this information is not updated anyway.

[Minor] debian/control: Update Vcs-Git URL and maintainer

* Let Vcs-Git point to a secure transport.
* Use Vsevolod's email address as maintainer to avoid having to patch
  this file in rspamd_build.sh. Mikhail is not the current maintainer of
  the Debian package anyway.

[Minor] use quilt for debian/source/format

* Match the build scripts and main Debian packaging.
* This allows for a separate debian/ directory in debian.tar.gz, see
  https://www.debian.org/doc/manuals/maint-guide/dother.en.html#sourcef

### DIFF
--- a/centos/rspamd.spec
+++ b/centos/rspamd.spec
@@ -8,7 +8,7 @@
 %define rspamd_wwwdir   %{_datadir}/rspamd/www
 
 Name:           rspamd
-Version:        1.1.0
+Version:        2.4
 Release: 1
 Summary:        Rapid spam filtering system
 Group:          System Environment/Daemons
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,40 +1,5 @@
-rspamd (1.0.2) unstable; urgency=low
+rspamd (2.4) unstable; urgency=low
 
   * New release.
 
- -- Mikhail Gusarov <dottedmag@debian.org>  Thu, 17 Sep 2015 15:17:32 +0100
-
-rspamd (0.9.4) unstable; urgency=low
-
-  * New release:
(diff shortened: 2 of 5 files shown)


vstakhov:
Will it break the existing build scripts by the way or they are adopted already?

Lekensteyn:
Updated the build scripts. Sorry for breaking the 2.3 ASAN builds, these have been fixed while at it.

There is more potential for simplifying the build scripts.
To enable ASAN for example, one also pass an env var instead of patching the debian/rules file: https://wiki.debian.org/LTS/Development/Asan

The `chroot ${HOME}/d sh -c "cd X; sed Y < Z > /tmp/.tt; mv /tmp/.tt Z"` could also be simplified to `chroot ${HOME}/d sed Y -i X/Z`. GNU sed does support this. MacOS (presumably all BSD?) require a special `-i ""` option instead of `-i`, but that detail should not matter for these scripts.

vstakhov:
These scripts are intended to run on Linux only. Hence, implying sed == GNU sed is okay. Thank you!

Issue #2035, opened 2018-02-27
Look up systemd sockets by name


argv-minus-one:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [ ] Other bug
* [ ] Feature
* [x] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:

`1.7.0-0~git115~395c1e684~sid`

### Operation system, CPU, memory and environment:

Debian sid, Linux, x86-64, 16GB

### Description:

As I described in issue #2034, I've tried to make a setup where rspamd is handed three different sockets by systemd, and assigns them to the corresponding workers. Because the sockets have different owners, they must be declared in several systemd socket units, rather than just one. They are then attached to the `rspamd.service` with a `Sockets=` directive, and systemd does indeed pass them along.

However, systemd does not guarantee which order the socket FDs from different socket units will be in! Per [`man sd_listen_fds`](https://www.freedesktop.org/software/systemd/man/sd_listen_fds_with_names.html):

>If multiple socket units activate the same service, the order of the file descriptors passed to its main process is undefined.

As a result, with this setup, selecting sockets by FD number doesn't work correctly. The workers will end up listening on the wrong sockets.

Systemd *does* inform rspamd of which sockets are which, through the `LISTEN_FDNAMES` environment variable. (These can be retrieved with the `sd_listen_fds_with_names` function.) But rspamd does not seem to support attaching systemd named sockets. For example, `bind_socket = "systemd:control";` gives a parse error:

	Feb 26 17:35:50 rspamd[6001]: (main) <gayif8>; cfg; rspamd_parse_bind_line: cannot parse bind line: systemd:control
	Feb 26 17:35:50 rspamd[6001]: (main) <gayif8>; cfg; rspamd_config_read: rcl parse error: cannot parse bind line: systemd:control

Please add support for systemd named sockets.

### Configuration:

Same as in issue #2034, except `/etc/rspamd/override.d/worker-controller.inc` instead contains:

	bind_socket = "systemd:control";

### Steps to reproduce:

1. Apply the setup described in issue #2034.
2. Apply the above change to `worker-controller.inc`.
3. `systemctl start rspamd`

### Expected results:

[...]

PR #3265, opened 2020-02-15
[Fix] Support listening on systemd sockets by name

Lekensteyn:
* Add support for looking up sockets by the systemd socket name, e.g.
  `systemd:rspamd-proxy.socket` or the name from `FileDescriptorName`.
  https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html
* Replace EOVERFLOW error code to avoid confusing error messages.

Fixes: #2035
___
Tested with various combinations (and with ASAN/UBSAN):
- `systemd:control`
- `systemd:0`, `systemd:control`
- `systemd:-1` (bad)
- `systemd:1` (bad)
- `systemd:invalid` (bad)
- `systemd:control`, `systemd:proxy`

### DIFF
--- a/src/libserver/cfg_utils.c
+++ b/src/libserver/cfg_utils.c
@@ -98,7 +98,7 @@ rspamd_parse_bind_line (struct rspamd_config *cfg,
 	const gchar *str)
 {
 	struct rspamd_worker_bind_conf *cnf;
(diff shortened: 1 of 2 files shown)


vstakhov:
Looks good to me, thanks again!

PR #3291, opened 2020-03-07
[Minor] --var doesn't work

tech-ozon-io:
We can add varaibels on the command line after argument or using --var parameter.
When --var is used, an assertion is triggered:

```
./rspamd.install/bin/rspamd --var LOGDIR=/tmp/
(rspamd:7171): GLib-CRITICAL **: g_hash_table_insert_internal: assertion 'hash_table != NULL' failed
```
The hash table is not initialized, because the table is initialized after
the argument parser.

This patch initialize the hash table on demand.

**Note**: the patch is tested on 1.9 version and not on master, because I can't compile the master branch. The code is easy, I guess the right behavior.

**Note**: I suggest to backport the patch on 1.9 version

### DIFF
--- a/src/rspamd.c
+++ b/src/rspamd.c
@@ -148,6 +148,11 @@ rspamd_parse_var (const gchar *option_name,
 		v = g_strdup (t + 1);
 		*t = '\0';
 
+		if (ucl_vars == NULL) {
+			ucl_vars = g_hash_table_new_full (rspamd_strcase_hash,
+					rspamd_strcase_equal, g_free, g_free);
+		}
+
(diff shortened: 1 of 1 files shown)


vstakhov:
Merged, thank you! I do not plan to backport anything to 1.9 as this version is not supported for a long time.

tech-ozon-io:
thanks for the advice !
—
Thierry Fournier
Web Performance & Security Expert
m: +33 6 68 69 21 85      | e: thierry.fournier@ozon.io
w: http://www.ozon.io/    | b: http://blog.ozon.io/

PR #3294, opened 2020-03-09
[Minor] Compilation error with __has_attribuute operator

tech-ozon-io:
GCC recommands to add alternative to __has_attribute operator:

https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html

Without this alternative, the project doesn't build with some GCC version.
In other way, the alternative is written i other part of rspamd like mem_pool.h

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -375,8 +375,12 @@ typedef off_t goffset;
 #  endif
 #elif defined(__GNUC__)
 /* GCC based */
-#  if __has_attribute(__no_sanitize_address__)
-#    define RSPAMD_NO_SANITIZE __attribute__((no_sanitize_address))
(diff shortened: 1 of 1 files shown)


vstakhov:
The minimum version of gcc is about 4.7 afair. I'm not going to support old compilers at all: too much hassle for nothing. Soon, Rspamd might require С++17 compatible compiler, so I'd recommend not to bother about old toolchains support.

tech-ozon-io:
ok, you're right. The fail is with gcc 4.8.4.

vstakhov:
Anyway, this change seems to be good, thank you.

tech-ozon-io:
thanks.

PR #3311, opened 2020-03-26
[Feature] metadata_exporter - add custom variables

c-rosenberg:
e.g. grep symbol options like virus names for recipient reports


vstakhov:
Thank you! I would like to postpone this feature unless 2.5 is released. Does it make sense?

c-rosenberg:
That's fine for me. I have just 1-2 installations using it.
I will PR some examples for rspamd.com in the next days

vstakhov:
Well, I can include that to 2.5 if that is guaranteed not to break other things. It looks like it won't from what I see...

src/plugins/lua/metadata_exporter.lua
```diff
@@ -63,6 +63,18 @@ Score: $score
 Symbols: $symbols]],
 }
 
+local variables = {
+  content = function(task)
+   return tostring(task:get_content())
```
vstakhov: Please don't do it! `rspamd_text` is designed precisely to avoid doing tostring interning for large strings.

src/plugins/lua/metadata_exporter.lua
```diff
@@ -138,6 +150,11 @@ local function get_general_metadata(task, flatten, no_content)
     r.header_subject = process_header('subject')
     r.header_date = process_header('date')
     r.message_id = task:get_message_id()
+    if variables then
+      for k, _ in pairs(variables) do
+        r[k] = variables[k](task)
```
vstakhov: This will be very expensive due to string operations here.

src/plugins/lua/metadata_exporter.lua
c-rosenberg: Thx. I will check why I have added tostring here.

c-rosenberg:
sorry for the noise, I will open this PR again.

PR #3313, opened 2020-03-29
Revert "[Fix] Fix hiredis stupidity" and use the hiredis api as intended

arkamar:
This reverts commit b05b9bf904edce75c17e63982d5e0a82dd3e9064.

Async hiredis allows to pass user specific data to the callback function via `data` element of `struct redisAsyncContext` (see https://github.com/rspamd/rspamd/blob/192c6b791d1c2b42861df6526c663f1d0814ee84/contrib/hiredis/async.h#L69-L70). I would suggest to use it as intended and revert your changes. This change makes the rspamd compatible with upstream hiredis.

### DIFF
--- a/contrib/hiredis/async.c
+++ b/contrib/hiredis/async.c
@@ -129,7 +129,6 @@ static redisAsyncContext *redisAsyncInitialize(redisContext *c) {
 
     ac->onConnect = NULL;
     ac->onDisconnect = NULL;
-    ac->disconnectCbdata = NULL;
 
     ac->replies.head = NULL;
     ac->replies.tail = NULL;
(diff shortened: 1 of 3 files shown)


vstakhov:
I don't remember exactly why this change has been done but I believe it fixed some important issue. I'm not quite sure it is still needed though...

arkamar:
You did that because of this commit bc4cf433bbd60416a3d9e4812b7a48009d680418 because you needed to pass your data to the callback but you most probably did not notice the `redisAsyncContext.data`, therefore you tweaked hiredis.

vstakhov:
Probably because `ctx->data` is used by a pool consumer (e.g. lua_redis) and should not be used for the pool data itself.

arkamar:
I did not find any place where `ctx->data` would be used. I hope I did't overlook anything. Anyway, rspamd seems to work for me with this change.

vstakhov:
Ok, I have examined it and it should work indeed. Thank you!

arkamar:
Thanks!

PR #3326, opened 2020-04-07
[Fix] rbl.lua: Helo Check always skipped

membero:
check_helo is missing return true. The check is skipped always.


vstakhov:
Good catch, thank you!

PR #3332, opened 2020-04-11
Resolve various Compiler warnings

cgzones:


contrib/libev/ev.c
```diff
@@ -2146,7 +2146,7 @@ typedef struct
   #include "ev_wrap.h"
 
   static struct ev_loop default_loop_struct;
-  EV_API_DECL struct ev_loop *ev_default_loop_ptr = 0; /* needs to be initialised to make it a definition despite extern */
```
vstakhov, line 2149: This is left for a reason. It is proven that making this var static might cause multiple instantiation of it in `rspamd` binary and `librspamd-server.so`. This usually takes place when IPO is turned on. Hence, I would advice to revert this change please.

vstakhov:
Very cool, thank you very much! The only thing I cannot merge for now is that `extern` default loop in libev...

contrib/libev/ev.c
cgzones: Dropped the related commit

vstakhov: Thank you very much!

Issue #3335, opened 2020-04-14
[Feature] allow variables in force_action return messages


rm-minus-rf:
## Summary
Allow variables in the smtp return message in force_action rules.

## Motivation
We want to allow variables in the return messages to give the peer more and more precise information. Examples for such information can be: queue-id or the filenames of blocked attachments.
We had positive experience with including these kind of information in the smtp return messages:
1) the probability that the peer admin/user will understand the situation and no case at the service-desk is opened increases.
2) If a case is opened most of the time the smtp return message is preserved and the 1st level can handle it based on the included information.
3) in all levels the included queue-id (named support-id in our case) helps a lot tracking the mail in logs.

## Describe alternatives you've considered
No alternatives found.

## Additional context
Our idea was to allow 3 prefixes in the variables:
- 'selector::\<selector-expression\>' => fill with value extracted by a selector
 (selectors are powerful and allow for a lot of different information)
- 'symbol::\<symbol-name\>'   => fill with matching symbol option string(s)
 (we wrote a custom lua plugin to filter attachments which adds unwanted files as symbol option string)
- 'task::\<task-attribute\>'     => fill with matching task attribute (ONLY 'queue_id' and 'uid' ALLOWED!)
 (this prefix is needed for queue_id and uid which cannot be extracted by a selector) 

Let´s look at an example:
```
  DAAS_REJECT_MIME_BLACK {
    expression = "DAAS_MIMEDIVER_BLACKLISTED & !(DAAS_MIME_BLACK_OK)";
    action = "reject";
    honor_action = ["reject"];
    message = "(support-id: ${task::queue_id}) Your mail was rejected because it contains BANNED ATTACHMENTS: ${symbol::DAAS_MIMEDIVER_BLACKLISTED}. Please check https://www.example.com/${selector::languages.first}/allowed-attachments.html for further details!"
  }
```
Will expand to: `5.7.1 (support-id: 491cJ173kpzyT9) Your mail was rejected because it contains BANNED ATTACHMENTS: MExplorer.exe. Please check https://www.example.com/en/allowed-attachments.html for further details!`

Here is a patch that should allow the mentioned prefixes/variables:
```
--- /root/work/force_actions.lua.org	2020-04-09 10:53:01.639165761 +0200
+++ /usr/share/rspamd/plugins/force_actions.lua	2020-04-09 15:45:56.128199622 +0200
@@ -29,6 +29,7 @@
[...]

PR #3336, opened 2020-04-14
[Feature] allow variables in force_actions messages

rm-minus-rf:
Allow variables with the following prefixes in the smtp return message for force_actions rules:
- 'selector::\<selector-expression\>' => fill with value extracted by a selector
 (selectors are powerful and allow for a lot of different information)
- 'symbol::\<symbol-name\>'   => fill with matching symbol option string(s)
 (e.g. custom lua plugin to filter attachments adds unwanted files as symbol option string)
- 'task::\<task-attribute\>'     => fill with matching task attribute (ONLY 'queue_id' and 'uid' ALLOWED!)
 (this prefix is needed for queue_id and uid which cannot be extracted by a selector) 

See https://github.com/rspamd/rspamd/issues/3335 for details.

### DIFF
--- a/lualib/lua_selectors/extractors.lua
+++ b/lualib/lua_selectors/extractors.lua
@@ -419,21 +419,12 @@ The first argument must be header name.]],
   ['symbol'] = {
     ['get_value'] = function(task, args)
       local symbol = task:get_symbol(args[1])
-      if args[2] and symbol then
-	if args[2] == 'options' then
-	  -- concat options tables to avoid table representation strings produced by implicit conversion
-          return fun.map(function(r) return table.concat(r[args[2]], ', ') end, symbol), 'string_list'
-	elseif args[2] == 'score' then
-	  -- only userdata_list seems to work for scores
-          return fun.map(function(r) return r[args[2]] end, symbol), 'userdata_list'
-	else
-          return fun.map(function(r) return r[args[2]] end, symbol), 'string_list'
-	end
+      if symbol then
+        return symbol[1],'table'
       end
-      return symbol,'table_list'
     end,
-    ['description'] = [[Get specific symbol. The first argument must be the symbol name. If no second argument is specified, returns a list of symbol tables. Otherwise the second argument specifies the attribute which is returned as list (`options`, `score` or `group`)]],
-    ['args_schema'] = {ts.string, ts.one_of{'options','score','group'}:is_optional()}
+    ['description'] = [[Get specific symbol. The first argument must be the symbol name. Returns the symbol table. See task:get_symbol()]],
+    ['args_schema'] = {ts.string}
   },
 
 }
--- a/lualib/lua_selectors/extractors.lua
+++ b/lualib/lua_selectors/extractors.lua
@@ -388,6 +388,54 @@ The first argument must be header name.]],
   ]],
     ['args_schema'] = { ts.one_of { 'stem', 'raw', 'norm', 'full' }:is_optional()},
   },
+  -- Get queue ID
+  ['queueid'] = {
+    ['get_value'] = function(task)
+      local queueid = task:get_queue_id()
+      if queueid then return queueid,'string' end
+      return nil
+    end,
+    ['description'] = [[Get queue ID]],
+  },
+  -- Get ID of the task being processed
+  ['uid'] = {
+    ['get_value'] = function(task)
+      local uid = task:get_uid()
+      if uid then return uid,'string' end
+      return nil
+    end,
+    ['description'] = [[Get ID of the task being processed]],
+  },
+  -- Get message ID of the task being processed
+  ['messageid'] = {
+    ['get_value'] = function(task)
+      local mid = task:get_message_id()
+      if mid then return mid,'string' end
+      return nil
+    end,
+    ['description'] = [[Get message ID]],
+  },
+  -- Get specific symbol
+  ['symbol'] = {
+    ['get_value'] = function(task, args)
+      local symbol = task:get_symbol(args[1])
+      if args[2] and symbol then
+	if args[2] == 'options' then
+	  -- concat options tables to avoid table representation strings produced by implicit conversion
+          return fun.map(function(r) return table.concat(r[args[2]], ', ') end, symbol), 'string_list'
+	elseif args[2] == 'score' then
+	  -- only userdata_list seems to work for scores
+          return fun.map(function(r) return r[args[2]] end, symbol), 'userdata_list'
+	else
+          return fun.map(function(r) return r[args[2]] end, symbol), 'string_list'
+	end
+      end
+      return symbol,'table_list'
+    end,
+    ['description'] = [[Get specific symbol. The first argument must be the symbol name. If no second argument is specified, returns a list of symbol tables. Otherwise the second argument specifies the attribute which is returned as list (`options`, `score` or `group`)]],
+    ['args_schema'] = {ts.string, ts.one_of{'options','score','group'}:is_optional()}
+  },
+
 }
 
 return extractors
--- a/src/plugins/lua/force_actions.lua
+++ b/src/plugins/lua/force_actions.lua
@@ -64,65 +64,28 @@ local function gen_cb(expr, act, pool, message, subject, raction, honor, limit, 
 
   return function(task)
 
-    local function fill_vars(repl, var_class, var_payload)
-      -- fill template vars prefixed with 
-      --  'selector::' => fill with value extracted by a selector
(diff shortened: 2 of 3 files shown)


src/plugins/lua/force_actions.lua
```diff
@@ -63,6 +64,66 @@ local function gen_cb(expr, act, pool, message, subject, raction, honor, limit,
...
 
+    local function fill_vars(repl, var_class, var_payload)
+      -- fill template vars prefixed with 
+      --  'selector::' => fill with value extracted by a selector
+      --  'symbol::'   => fill with matching symbol option string(s)
+      --  'task::'     => fill with matching task attribute
+      --                   (only queue_id and uid allowed)
+
+      if var_class == "selector" then
+        local selector = lua_selectors.create_selector_closure(rspamd_config, var_payload, '', true)
```
vstakhov, line 75: Selectors creation must be done once as this is an expensive procedure that can and will leak memory if used more than one time. So I would suggest to use some module-global cache of selectors like `local sel_cache = {}` and index it when a selector is requested by its name.

src/plugins/lua/force_actions.lua
```diff
@@ -63,6 +64,66 @@ local function gen_cb(expr, act, pool, message, subject, raction, honor, limit,
...
+        if not selector then 
+          rspamd_logger.errx(rspamd_config, 'could not create selector [%1]', var_payload)
```
vstakhov, line 77: I would better use `task` log object here.

src/plugins/lua/force_actions.lua
```diff
@@ -63,6 +64,66 @@ local function gen_cb(expr, act, pool, message, subject, raction, honor, limit,
...
+          return "((could not create selector))"
+        end
+        local extracted = selector(task)
+        if extracted then
+          -- replace non ascii chars
+          if type(extracted) == 'table' then
+            extracted = table.concat(extracted, ',')
+          end
+          extracted = string.gsub(extracted, '[\128-\255]', '?')
```
vstakhov, line 86: This will also filter all valid utf8 characters. Why this check is added after all?

src/plugins/lua/force_actions.lua
```diff
@@ -63,6 +64,66 @@ local function gen_cb(expr, act, pool, message, subject, raction, honor, limit,
...
+      if var_class == "symbol" then
+        local symb_opts
+        local symb_strs = {}
+        if task:has_symbol(var_payload) then
+          local symb_tbl = task:get_symbol(var_payload)
+          for _,symb in ipairs(symb_tbl) do
+            local symb_opts = symb.options
+            if symb_opts then
+              -- replace non ascii chars
+              symb_strs[#symb_strs+1] = string.gsub(table.concat(symb_opts, ','), '[\128-\255]', '?')
```
vstakhov, line 103: Same here, why this is needed?

src/plugins/lua/force_actions.lua
```diff
@@ -63,6 +64,66 @@ local function gen_cb(expr, act, pool, message, subject, raction, honor, limit,
...
+            end
+          end
+          symb_str = table.concat(symb_strs, ',')
+        else
+          symb_str = '((symbol not found))'
+        end
+        return symb_str
+      end
+
+      -- NOTE-TO-VSTAKHOV: would it make sense to export task:get_queue_id and task:get_uid as selector data definition?
```
vstakhov, line 113: Yes, I'd suggest to use selectors only here in fact. It will simplify the code and the overall logic. If you need to filter non-ascii characters it would be better to add the transformation to `lua_selectors/transforms.lua`

src/plugins/lua/force_actions.lua
rm-minus-rf: I see. I will try to use a global cache.

src/plugins/lua/force_actions.lua
rm-minus-rf: copy&paste error. Will fix it.

src/plugins/lua/force_actions.lua
rm-minus-rf: Because i do not know if the client has requested utf-8 extensions in the smtp dialog i wanted to make sure that we only return ascii chars.
Do you think it is ok to return utf-8 here and let the client handle the representation of unwanted chars?

src/plugins/lua/force_actions.lua
rm-minus-rf: Same as above. Symbol option strings could contain utf-8 chars as well. (They certainly do for the custom lua plugin we use, that adds blocked filenames to symbol option strings)

src/plugins/lua/force_actions.lua
rm-minus-rf: Ok. Then i will try to add queue-id, uid and symbols as selector data definitions.

(might take a while because i need to learn about github workflow first. Updating/changing the pull request for example ;-) )

rm-minus-rf:
Hello,

i tried to create a clean selectors only variant.

Therefore the prefixes are gone:
```
  DAAS_REJECT_MIME_BLACK {
    expression = "DAAS_MIMEDIVER_BLACKLISTED & !(DAAS_MIME_BLACK_OK)";
    action = "reject";
    honor_action = ["reject"];
    message = "(support-id: ${queueid}) Your mail was rejected because it contains BANNED ATTACHMENTS: ${symbol(DAAS_MIMEDIVER_BLACKLISTED, 'options').to_ascii}. Please check https://www.example.com/${languages.first}/allowed-attachments.html for further details!"
  }
```

It works for my use case and i tried to test some other selector combinations.
But i will admit that i am still fairly new to lua and the table structures returned by task:get_symbol() combined with the selector extraction logic challenged me.
Have a good look if this looks fine for you, please.

vstakhov:
That now looks much better, thank you! I will do a more detailed review but at the first glance it is exactly what I have in my mind.

lualib/lua_selectors/extractors.lua
```diff
@@ -388,6 +388,54 @@ The first argument must be header name.]],
...
+  ['symbol'] = {
+    ['get_value'] = function(task, args)
+      local symbol = task:get_symbol(args[1])
+      if args[2] and symbol then
+	if args[2] == 'options' then
+	  -- concat options tables to avoid table representation strings produced by implicit conversion
+          return fun.map(function(r) return table.concat(r[args[2]], ', ') end, symbol), 'string_list'
+	elseif args[2] == 'score' then
+	  -- only userdata_list seems to work for scores
+          return fun.map(function(r) return r[args[2]] end, symbol), 'userdata_list'
```
vstakhov, line 428: Why not use just `tostring(score)` here?

lualib/lua_selectors/extractors.lua
```diff
@@ -388,6 +388,54 @@ The first argument must be header name.]],
...
+      local mid = task:get_message_id()
+      if mid then return mid,'string' end
+      return nil
+    end,
+    ['description'] = [[Get message ID]],
+  },
+  -- Get specific symbol
+  ['symbol'] = {
+    ['get_value'] = function(task, args)
+      local symbol = task:get_symbol(args[1])
```
vstakhov, line 421: I'd suggest to simplify this code. In fact `get_symbol` always return a table with one element (legacy compatibility). So we can get it and greately simplify the former stuff by just returning a `table`. So `args[2]` will be redundant, since a table can be indexed by a table methods + transforms.

lualib/lua_selectors/extractors.lua
vstakhov: Hum, but probably better follow my comment above and remove this logic completely.

lualib/lua_selectors/extractors.lua
rm-minus-rf: Sounds perfect, thanks! The table_list gave me a hard time.
Returning the first element as a table, will make things a lot easier.
I will try it out as soon as i will be back at a pc (probably tommorrow).

rm-minus-rf:
The last commit would change my example rule to:
```
  DAAS_REJECT_MIME_BLACK {
    expression = "DAAS_MIMEDIVER_BLACKLISTED & !(DAAS_MIME_BLACK_OK)";
    action = "reject";
    honor_action = ["reject"];
    message = "(support-id: ${queueid}) Your mail was rejected because it contains BANNED ATTACHMENTS: ${symbol('DAAS_MIMEDIVER_BLACKLISTED'):options.to_ascii}. Please check https://www.example.com/${languages.first}/allowed-attachments.html for further details!"
  }
``` 
Thanks for your patience :-)

vstakhov:
Great, thank you very much!

PR #3341, opened 2020-04-17
Improve lua api

jendis:
These changes complete the missing parts of documentation and get rid of duplicate methods in the `rspamd_util` module.

### DIFF
--- a/src/lua/lua_task.c
+++ b/src/lua/lua_task.c
@@ -519,6 +519,14 @@ LUA_FUNCTION_DEF (task, has_from);
  * - `addr` - address part of the address
  * - `user` - user part (if present) of the address, e.g. `blah`
  * - `domain` - domain part (if present), e.g. `foo.com`
+ * - `flags` - table with following keys set to true if given condition fulfilled:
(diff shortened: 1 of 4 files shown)


vstakhov:
That all looks very good thank you very much! I would suggest to leave alias for `util.parse_addr` that would just redirect to `util.parse_mail_addr`. Their call semantics is almost similar from what I see. I agree that the current lua API is sometimes spaghetti style: it was written for a long time and without some clean design rules (I'm sorry for that btw). However, I'd prefer to keep API backward compatible where possible. Removing of methods should be avoided at any cost.

jendis:
Initially I planned to make an alias between these two `util.parse_addr` and `util.parse_mail_addr`. I finally decided to remove the `util.parse_addr` because of its ambiguous name but mainly I prefer as minimal API as possible. It took me many hours to get into the API functionality and I had to do many tests before I was sure what should be used where. These duplicates make it more difficult.
I'd suggest to make an alias for the next release but mark it as obsolete and remove it afterwards.

vstakhov:
I agree with this approach, it makes sense. I really want to clean up the API and make it more consistent.

jendis:
Oh, sorry, my mistake.

PR #3342, opened 2020-04-19
[Fix] Do not flag pre-result of virus scanners as least if action is reject

### LINKED ISSUE #3279: [BUG] antivirus "message" is ignored
### Prerequisites

* [ X] Put an X between the brackets on this line if you have done all of the following:

**Describe the bug**

The `message` setting in `antivirus.conf` is ignored. Instead, a generic "Spam message rejected" message is returned to the client.

### Steps to Reproduce
```
# local.d/antivirus.conf snippet
clamav {
  [...]
  action = "reject";
  message = 'Virus found: ${VIRUS}';
  [...]
}
```

**Expected behavior**
milter-reject message "Virus found: Win.Test.EICAR_HDB-1"

**Actual behavior**
milter-reject message "Spam message rejected"

### Versions
Affected: Version 2.4 (precisely: `2.4-2~bionic` from `http://rspamd.com/apt-stable bionic/main`)

Note: exactly this configuration worked correctly (i.e., as expected) in version 2.2.

### Additional Information

This seems to be related, or at least very similar, to #3269 and #3276. However, from what I understand, these were presumably fixed in (or before) version 2.4, whereas this bug occurs with version 2.4-2.

chkpnt:
If the antivirus module says "Reject this mail", this shouldn't be overruled later on.

fixes #3279

### DIFF
--- a/lualib/lua_scanners/common.lua
+++ b/lualib/lua_scanners/common.lua
@@ -68,6 +68,7 @@ local function yield_result(task, rule, vname, dyn_weight, is_fail)
   local symbol
   local threat_table = {}
   local threat_info
+  local flags
 
   -- This should be more generic
   if not is_fail then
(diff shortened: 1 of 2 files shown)


vstakhov:
I do agree with this change. Presumably `least` in conjunction with the most severe threshold action should override threshold action. But this change also looks good. Probably, it will be bad for those who use `discard` action but I'm inclined to merge it now and re-think it at some time in future.

PR #3344, opened 2020-04-20
[Feature] extend lua api

jendis:
add new value accessible by key 'raw' to the table returned by util:parse_mail_address and task:get_from


vstakhov:
Ok, after some discussing I think this change is fine.

vstakhov:
Thank you!

Issue #3249, opened 2020-02-05
[BUG] url_redirector module: rspamd_lua_check_udata_common: lua type error


uvelokk:
### Prerequisites

* [x ] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
URL redirector module logs an error during connection to redis service. One example:
```
#8401(normal) <d313ef>; url_redirector; url_redirector.lua:314: check url https://goo.gl/forms/snip
#8401(normal) <d313ef>; url_redirector; url_redirector.lua:230: select user agent Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; rv:11.0)
#8401(normal) <d313ef>; url_redirector; url_redirector.lua:196: found redirect from https://goo.gl/forms/snip to https://docs.google.com/snip, err code 302
#8401(normal) <d313ef>; url_redirector; url_redirector.lua:204: stop resolving redirects as https://docs.google.com/snip is not a redirector
#8401(normal) rspamd_lua_check_udata_common: lua type error: expected rspamd{text} at position -1, but userdata has table metatable; trace:  [1]:{/usr/share/rspamd/lualib/lua_redis.lua:946 - redis_make_request [Lua]}; [2]:{/usr/share/rspamd/plugins/url_redirector.lua:143 - cache_url [Lua]}; [3]:{/usr/share/rspamd/plugins/url_redirector.lua:206 - <unknown> [Lua]}; stack(5): [1: table] [2: table] [3: table] [4: number] [5: ud=rspamd{url}] 
#8401(normal) lua_http_finish_handler: callback call failed: /usr/share/rspamd/lualib/lua_redis.lua:946: bad argument #5 to 'make_request' ('text' expected)
```

### Steps to Reproduce

[...]

PR #3345, opened 2020-04-21
[Fix] #3249

jendis:


jendis:
Solves problem with exhausting redis connections too.

vstakhov:
Good catch, thank you!

PR #3346, opened 2020-04-21
Add FROM_INVALID and ENVFROM_INVALID rules

jendis:


vstakhov:
Am I understanding correct that this issue depends on #3344 initially?

jendis:
Yes, it does.

vstakhov:
Ok, I have committed it so this can also be safely merged. Thank you for your work, much appreciated!

PR #3355, opened 2020-04-27
[fix] add support ElasticSearch>=7 for elastic module

### LINKED ISSUE #3324: [BUG] Elasticsearch exporter bulk operation fails
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:

**Describe the bug**
Using rspamd 2.5 and elastic 7.6.2 with a fresh install the elastic.lua bulk insert fails with a "strange error":
java.lang.IllegalArgumentException: mapper [rspamd_meta.geoip.location] of different type, current_type [geo_point], merged_type [ObjectMapper].

### Steps to Reproduce

1. Install rspamd
2. Install elastic
3. Configure rspamd
4. See index and pipeline created in elastic but not documents in index
5. See elasticsearch log for the mapper error.

**Expected behavior**
I expected to see data into elastic.

### Versions

Debian buster
Rspamd 2.5
elastic 7.6.2

### Additional Information

After a lot of debug I found that the error reported is not correct or is a sub-problem of the fact that the elastic.lua in the bulk request is sending a deprecated parameter (_type). Maybe is it possibile to check is the elastic is version > 6 (using ingest_module param or adding one?) to change the bulk request? If I use this all is ok.
Maybe I can add a pull request?

Line 74:
from
        '", "_type" : "logs" ,"pipeline": "rspamd-geoip"} }')
to
        '","pipeline": "rspamd-geoip"} }')

Not sure but this can be modified even in the kibana part.

gauthierc:
Since version 7.x of ElasticSearch, logs type is not supported.

Close #3324

### DIFF
--- a/src/plugins/lua/elastic.lua
+++ b/src/plugins/lua/elastic.lua
@@ -71,7 +71,7 @@ local function elastic_send_data(task)
   local tbl = {}
   for _,value in pairs(rows) do
     table.insert(tbl, '{ "index" : { "_index" : "'..es_index..
-        '", "_type" : "logs" ,"pipeline": "rspamd-geoip"} }')
+        '", "_type" : "_doc" ,"pipeline": "rspamd-geoip"} }')
     table.insert(tbl, ucl.to_format(value, 'json-compact'))
   end
 


vstakhov:
It's very hard for me to test, but I assume that 5.6-current ES versions should be safe with this change. So I will merge it, thank you!

PR #3363, opened 2020-05-05
[Fix] lowercase for emails checked on RBLs

membero:
email_tbl.addr was not used in add_dns_request.

### DIFF
--- a/src/plugins/lua/rbl.lua
+++ b/src/plugins/lua/rbl.lua
@@ -589,7 +589,8 @@ local function gen_rbl_callback(rule)
 
   local function check_email_table(task, email_tbl, requests_table, whitelist, what)
     lua_util.remove_email_aliases(email_tbl)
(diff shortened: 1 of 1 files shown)


vstakhov:
The problem is that username part is **case sensitive** by RFC. I doubt this is a correct change (at least from terms of the standard).

membero:
Damn, you're right. I did not think about that.

As this is part of RBL checking, do you think a switch email_tolower (true/false) would be OK? 

My intention was to get rid of all the replyto spam where the big freemailers are used. E.g. Gmail, Outlook etc. do not use case sensitivity. In real world I do not believe that it would come to problems.

vstakhov:
Ok, I agree with this point for now. In general, we need to use specific composition rules as defined by parcicular RBL. But the vast majority of them do lowercase indeed. E.g. that's how EBL is designed: https://msbl.org/ebl-design.html

PR #3369, opened 2020-05-12
[Feature] metadata_exporter - add custom variables - try #2

c-rosenberg:
Thanks for your suggestions - hope I got you right - #3311


vstakhov:
I'm still thinking why don't we use selectors framework here?

c-rosenberg:
Definitely. I like the new implementation in force_actions. I'm a bit short in time currently. Give me some days to add it.

PR #3373, opened 2020-05-15
[Fix] make control.c load configuration to get control_socket value

### LINKED ISSUE #3371: [BUG] rspamadm control command does not use config file for "control_socket"
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:

**Describe the bug**
`rspamadm control` does not use the configuration files to get the default `control_socket` path.

### Steps to Reproduce

1. [First Step]
edit local.d/options.inc and change path of control_socket with `control_socket = /run/rspamd`
2. [Second Step]
reload rspamd
3. [and so on...]
issue an rspamadm control command
`rspamadm control reload`
which gives error
`cannot open connection to /var/lib/rspamd/rspamd.sock: No such file or directory`

**Expected behavior**

rspamadm control would be expected to use the value of `control_socket` from the configuration files, not one hard coded in rspamadm/control.c

### Versions
2.5, arm64, alpine linux edge

You can get this information from copy and pasting the output of `rspamd --version` from the command line. Also, please include the **OS** and what version and architecture including the generation of CPU model (e.g. Haswell) where applicable,  of the OS you're running.

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

a16bitsysop:
Use as default for control commands, based on configtest.c resolves #3371

### DIFF
--- a/src/rspamadm/control.c
+++ b/src/rspamadm/control.c
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 #include "config.h"
+#include "rspamd.h"
 #include "rspamadm.h"
+#include "cfg_file.h"
+#include "cfg_rcl.h"
 #include "cryptobox.h"
 #include "printf.h"
 #include "libserver/http/http_connection.h"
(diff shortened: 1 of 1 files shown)


vstakhov:
The problem with this PR is that config reading is quite an expensive procedure tbh. It is better to use a different method to do partial load (just ucl parsing). I will check later, thank you for the report.

a16bitsysop:
Thanks, I did not know the best way to do it as it is only for the one value.

PR #3384, opened 2020-05-29
Fix analysis table / oletools pattern

ol3k:
I think after rework of oletools with PR #3368 the extended oletools mode and pattern config can't be used any more.

the return analysis_cat_table is empty when using string values in tables and table.concat

### DIFF
--- a/lualib/lua_scanners/oletools.lua
+++ b/lualib/lua_scanners/oletools.lua
@@ -291,7 +291,7 @@ local function oletools_check(task, content, digest, rule)
           lua_util.debugm(N, task, '%s: analysis_keyword_table: %s', rule.log_prefix, analysis_keyword_table)
           lua_util.debugm(N, task, '%s: analysis_cat_table: %s', rule.log_prefix, analysis_cat_table)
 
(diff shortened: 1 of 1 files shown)


vstakhov:
This change looks very ugly. Instead of the table with names we now have some crazy table like `table = {[1] = '_', [2] = '_' ...}`. It should be done in a better way definitely.

vstakhov:
I think this change should be enough

ol3k:
You are right,  this looks much better.

But you have to make sure the return string has the correct order to filter it later in oletools patterns. Therefore I added the "integer table".

1. M=Macros, 
2. A=Auto-executable
3. S=Suspicious keywords
4. I=IOCs
5. H=Hex strings
6. B=Base64 strings
7. D=Dridex strings
8. V=VBA strings

Symbol: OLETOOLS (3.50)[S---AM-I, Auto_Open, Open, ShellExecuteA, Shell32, CALL, URLDownloadToFileA, Hex Strings]

The return "S---AM-I" should be "MASI----".

vstakhov:
Then we need to sort keys and keep them sorted in a description.

c-rosenberg:
Maybe we should just revert to the old variant with concatenated strings?

c-rosenberg:
What about this variant? #3390

Issue #2872, opened 2019-04-25
[BUG] Invalid DMARC Aggregate Reports generated by rspamd


prehor:
I have validated approximately 9600 DMARC Aggregate Reports generated by rspamd versions 1.8.3 and 1.9.1 during March and April on my mail server against [XSD Schema in RFC7489](https://tools.ietf.org/html/rfc7489#appendix-C). I have found three types of errors:

1) 82 reports contained in the `reason` element in `policy_evaluated` string `sampled_out`:
```xml
<policy_evaluated>
  <disposition> quarantine</disposition>
  <dkim>fail</dkim>
  <spf>fail</spf>
  <reason>sampled_out</reason>
</policy_evaluated>
```
The `reason` element is not a string, but a complex type with `type` and `comment` elements and reason must be in the `type` element:
```xml
<policy_evaluated>
  <disposition>quarantine</disposition>
  <dkim>fail</dkim>
  <spf>fail</spf>
  <reason>
    <type>sampled_out</type>
  </reason>
</policy_evaluated>
```

2) 81 reports contained in `disposition` element in `policy_evaluated` string `softfail`:
```xml
<policy_evaluated>
  <disposition>softfail</disposition>
  <dkim>fail</dkim>
  <spf>fail</spf>
</policy_evaluated>
```
 The `disposition` element can contain only strings `none`, `quarantine` and `reject`. I think `softfail` should be changed to `quarantine`:
```xml
<policy_evaluated>
  <disposition>quarantine</disposition>
  <dkim>fail</dkim>
  <spf>fail</spf>
</policy_evaluated>
```

3) 5 reports did not contain any `record` element. These reports should never be sent.

<hr>

Finally, I would like to ask whether it would be possible for mails that have a score above the Probably Spam level to not be included in DMARC reports. In my case, 1900 reports from 9600 were for domains from TLD .icu, from which we only get spam. I don't want to send to spammers information about how successful they are. This new feature was already described in #2745, which was unfortunately closed by stale bot.

vstakhov:
The main problem in DMARC reporting is that the current code quality is very poor so the only way how to make it working is to rewrite this part of the plugin I'm afraid. On the other hand, I don't have any desire in doing that as well as it is not used by me anywhere...

yoshimo:
In that case you should disable the code anf remove the feature.

vstakhov:
I think I'm still able to decide what to do, thanks.

prehor:
[...]

Issue #2745, opened 2019-02-07
Filter emails for rspamd's DMARC reports


analogic:
I've started using native DMARC reports, but there is one problem. 

Imagine you get everyday single email from multiple domains which end up with score 20+. They have proper NS records for everything DMARC included but it has no connectable mx. Then when rspamd generate reports outbound queue get filled with hunderds or tousands of email reports every day waiting to be delivered.

I don't want to destroy DMARC reports purpose for proper domains but there should be a way how to filter either reports itself or emails counted to reports.

michelsup:
Sorry to reopen such an old issue but I think the problem got so worse recently that this issue should be resolved.
As more spammers use DMARC to try to circumvent antispam engines, we end up with hundreds of emails doing a loop.
- Spammer send an email rejected with a very high score
- Next day rspamd send a report. "Best" case is that the mail is queued for days, worst is that the mail is sent
- The next day the spammer send a report he has well received the report we sent him.
- The report is treated as spam and next day we still send a report we received a report.
- Rince and repeat indefinitely. 

This problem arise with all the reports. I send a report to a domain and the next day I receive another report. I guess in a few years most of the emails will be reports sent back and forth to thousands of domains nobody cares about anymore.

I guess a solution would be to check if it's a report we received, and discard it as "to be reported" to sender. And obviously remove any reports sent to spammy domains.

vstakhov:
The whole dmarc reporting part should be rewritten completely :(

PHPGangsta:
@michelsup You describe a loop of "DMARC Report mails"... If you send DMARC reports, you should send them from a subdomain, for which DMARC-Reporting is disabled. Then you break this loop very easily. You should also tell the other domain holder who sends back the reports that he should also send from a subdomain with disabled DMARC reporting. Only one has to do this to break the loop. And it's best if you both do this :-)

That's how you solve this problem. Don't try to "detect dmarc reports" or try to detect the loop in rspamd... The solution is: Don't request DMARC-Reports for the Reports you are sending, that's a lot better.

michelsup:
@PHPGangsta This is a very good advice you gave me. Many thanks.

PR #3387, opened 2020-06-02
Fix some DMARC issues

dasJ:
These issues were reported in #2872 which was closed by the stale bot.
This addresses the first two issues by getting rid of the string
"softfail" and by fixing the reason.

### DIFF
--- a/src/plugins/lua/dmarc.lua
+++ b/src/plugins/lua/dmarc.lua
@@ -218,9 +218,10 @@ local function dmarc_report(task, spf_ok, dkim_ok, disposition,
   local dkim_fail = table.concat(dres.fail or E, '|')
   local dkim_temperror = table.concat(dres.temperror or E, '|')
   local dkim_permerror = table.concat(dres.permerror or E, '|')
+  local disposition_to_return = if disposition == "softfail" then "none" else disposition
   local res = table.concat({
     ip:to_string(), spf_ok, dkim_ok,
-    disposition, (sampled_out and 'sampled_out' or ''), hfromdom,
+    disposition_to_return, (sampled_out and 'sampled_out' or ''), hfromdom,
     dkim_pass, dkim_fail, dkim_temperror, dkim_permerror, spfdom, spf_result}, ',')
 
   return res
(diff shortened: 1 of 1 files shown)


vstakhov:
@fatalbanana looking good?

fatalbanana:
The actual disposition is probably `none` in this case (not `quarantine`).

I have doubts about `<type>...</type>` inside `<reason>`- does the standard really call for it?

Lua doesn't have ternary operator so I suppose this needs to look slightly different.

I suppose it is most fitting to call `dmarc_report()` with `policy.dmarc_policy` rather than `disposition` which gets used in strange ways.

flowerysong:
> I have doubts about `<type>...</type>` inside `<reason>`- does the standard really call for it?

Yes, it does.

```
   <xs:complexType name="PolicyOverrideReason">
     <xs:all>
       <xs:element name="type" type="PolicyOverrideType"/>
       <xs:element name="comment" type="xs:string"
                   minOccurs="0"/>
     </xs:all>
   </xs:complexType>
```

> I suppose it might be most fitting to call `dmarc_report()` with `policy.dmarc_policy` rather than `disposition` which gets used in strange ways (_unless_ we are indeed supposed to be reporting our effective disposition & not the one suggested by policy).

The report is supposed to include both the policy published by the domain (in `policy_published`) and the effective disposition after local policies are applied.

dasJ:
@fatalbanana I rebased the branch to fix the tenary operator (also factored it into a variable to make things clearer).

dasJ:
Also, since I don't know lua that well, does anyone have any hints about the third problem in #2872? The code needs to skip reports without `record` elements.

src/plugins/lua/dmarc.lua
```diff
@@ -218,9 +218,10 @@ local function dmarc_report(task, spf_ok, dkim_ok, disposition,
   local dkim_fail = table.concat(dres.fail or E, '|')
   local dkim_temperror = table.concat(dres.temperror or E, '|')
   local dkim_permerror = table.concat(dres.permerror or E, '|')
+  local disposition_to_return = if disposition == "softfail" then "none" else disposition
```
vstakhov, line 221: That won't work as well I'm afraid.

vstakhov: `disposition == "softfail" and "none" or disposition` should work fine

prehor:
Thanks for fixing #2872. 

Finally, for DMARC reports to be usable, emails with a score higher than Probably Spam must not be included in DMARC reports, as described in #2745, which was unfortunately closed by stale bot.

PR #3390, opened 2020-06-03
[Minor] lua_scanners - Fix analysis table / oletools pattern - varian…

c-rosenberg:
…t #3

Just concatenate strings as done before.


vstakhov:
I'd better sort values. More error prone and neat than your version.

vstakhov:
So what do you think? :)

c-rosenberg:
I'm completely fine with your change.

PR #3396, opened 2020-06-07
[WebUI] Add "Test selectors" tab

moisseev:


vstakhov:
Thank you for doing that!

PR #3401, opened 2020-06-12
[Rules] Add X-CSA-Complaints as bulk header

Manawyrm:
This change adds a new rule to increase the score for the CSA-Complaints header.
This header is used by members of the german CSA: https://certified-senders.org/

You pay them a monthly fee: https://certified-senders.org/wp-content/uploads/2017/07/CSA_Preisliste.pdf and you are then allowed to use this header to signal your compliance. 
This is mostly used by spammers and bulk mail senders. 

This rule is being used in production and seems to work very well for a couple of weeks now.

### DIFF
--- a/rules/regexp/misc.lua
+++ b/rules/regexp/misc.lua
@@ -61,6 +61,13 @@ reconf['HAS_ONION_URI'] = {
     group = 'experimental'
 }
 
+reconf['CSA_COMPLAINTS'] = {
+  re = 'X-CSA-Complaints=/whitelist-complaints@eco.de/Hi',
+  score = 1.0,
+  description = "Sender is part of CSA",
+  group = 'bulk'
+}
+
 local my_victim = [[/(?:victim|prey)/{words}]]
 local your_webcam = [[/webcam/{words}]]
 local your_onan = [[/(?:mast[ur]{2}bati(?:on|ng)|onanism|solitary)/{words}]]


vstakhov:
I don't understand the positive scoring of this symbol. Do you mean that this header is typically forged by spammers?

Manawyrm:
All of the members of this alliance are bulk mail senders and send only newsletters, spam or marketing mails.
I have never seen a legitimate mail with that header, only bulk mails.

vstakhov:
Bulk does not mean spam :( That's why I'm not sure about this severe score.

Manawyrm:
That's true. Maybe this rule should be included with a neutral score and people can then set the score themselves? 

I don't want any unsubscribed bulk mails, most of our mailhosting users seem to think the same (I've never seen a complaint about these mails getting filtered).

andryyy:
I removed it from mailcow a while ago due to too many fp. CSA is indeed a useless, shitty alliance, but some legit senders use their headers.

CSA does not give a damn about their senders content. I once (and only once) reported a malicious mail they decided was fine. They just don't care ($$$).

So... I recommend not to merge it. Perhaps simply add a "CSA" symbol and leave it neutral.

Manawyrm:
Hm, ok. 
We're pretty much fine with the +1 score, because all legit senders I've seen have enough negative scores anyway (through NEURAL_HAM, DKIM, etc.).

I can update this PR to a neutral score by default, should I?

citrin:
@Manawyrm `X-CSA-Complaints` header is added to some important messages like bank statements/notifications. Adding positive score would not be a good idea in such case.

andryyy:
No. It is **not** a header for important messages. Please don't rate it as "good" mail... :P

citrin:
@andryyy This header doesn't indicate neither good neither bad messages. On it's own it is not very useful indicator.

andryyy:
CSA is the biggest useless crap on earth and widely used in spam Mails.

citrin:
It is also used by reputable senders like Bank of America.

hildeb:
Yes, it's quite commonplace. But what prevents me as a spammer from
simply adding this header?


Am Fr., 10. Juli 2020 um 13:58 Uhr schrieb Anton Yuzhaninov <
notifications@github.com>:

citrin:
Nothing, that's why it should not have a negative score either.

GieltjE:
I agree with the bulk != spam statement, a lot of newsletters can be very important to people.
Couldn't we check this list (as far as I read on their site you should have things like DKIM configured and single click unsubscribe is required) and if a sender is on the list and the message has valid DKIM then not mark it as spam?
Because in that case we have verified that it's an mailing the user has asked for and can unsubscribe if he wants too.

Manawyrm:
> we have verified

No, not really. Everybody can just add the Header. 
Spammers can do that, too. (and I often see this happening)

GieltjE:
> No, not really. Everybody can just add the Header.
> Spammers can do that, too. (and I often see this happening)

Like I said, if we can check what domains are on the list.

Manawyrm:
You mean, with a mechanism like DNSBL or something like that? 

That might be possible, but the line between spammers and "legit" CSA senders is pretty slim in any case.

GieltjE:
indeed, something along the lines of

X-CSA-Complaints available? --> "DNSDBL" or similar lookup --> failed? --> add points
X-CSA-Complaints available? --> "DNSDBL" or similar lookup --> succeeded? --> DKIM succeeded? --> subtract points

Manawyrm:
According to the FAQ, the member list is only available after you sign a license agreement with "eco": 
https://certified-senders.org/de/faq/

> How can I use the CSA whitelist and become a CSA partner?
> To become a CSA partner and use the whitelist, all you need to do is sign a license agreement.

That's not ideal. Might still be helpful to offer as an optional plugin, people can then sign this agreement and enable the plugin in Rspamd?

GieltjE:
With a plugin would be ideal, that way no action would be taken when it isn't enabled and configured otherwise the user would have configured it's api key (I guess?) and it would be a potentially valuable addition.

andryyy:
That would mean as long as you are DKIM valid and not on a BL, you can lower your score with Rspamd. Not ideal.

Even "valid" spam crap from CSA is sometimes misleading. I collect spam mails myself and often complained to CSA. They did not agree with my abuse reports and ignored the fact, that some of those "verified senders" send you invoices because you allegedly clicked some shady link and agreed to something. They are not listed on any DNSBL or simply using a relay service like Sendgrid, that will probably never be listed on such lists.

CSA failed.

Manawyrm:
> CSA failed.

I totally agree. Their header is worthless and I personally am doing very well by giving their header a bad score.

PR #3411, opened 2020-06-23
[Test] Don't use deprecated RobotFramework syntax

fatalbanana:
With the contemporary version of Robot Framework it's printing a lot of warnings:
```
[ WARN ] Accessing variable items using '&{kwargs}[CONFIG]' syntax is deprecated. Use '${kwargs}[CONFIG]' instead.
[ WARN ] Accessing variable items using '@{result}[0]' syntax is deprecated. Use '${result}[0]' instead.
```
This patch makes the warnings go away.


vstakhov:
Yes, it has been revealed during our experiments with Python 3. Thank you for your fix.

PR #3414, opened 2020-06-24
☠️ Use Ubuntu 20.04 & Python3 for functional tests ☠️

fatalbanana:
~~There are a couple of failures I need to check still:~~
 - ~~Fuzzy Skip Hash Test~~
 - ~~SSL [Large] TCP request~~ fixed

~~Probably `Follow Log` stuff can die.~~ gone

~~There is some more deprecated Robot syntax to fix, about for loops.~~ gone.

~~I should investigate demjson story.~~ looks ok to me; I restored it.


vstakhov:
Tada! Thank you very much, Andrew!

PR #3418, opened 2020-07-01
[WebUI] Migrate to Bootstrap 4.5

moisseev:


vstakhov:
Oh wow, thanks a lot, @moisseev !

andryyy:
Nice! That's great. :)

PR #3442, opened 2020-07-18
[Minor] Add systemd service sandbox features

tnias:
I need help figuring out if the sandboxing rules interfere with any rspamd features and how the SystemCallFilter could be tightened down.

I am trying to land similar sanboxing options in https://github.com/NixOS/nixpkgs/pull/93293


vstakhov:
Is it supported by older systemd versions, meaning at least that unknown options are ignored instead of causing errors?

tnias:
Yes. Systemd irgnores unknown/newer Options, but logs them to the journald log.
For example CentOS 8 (systemd 239) does not support ProtectClock, ProtectHostname, ProtectKernelLogs.

Mic92:
> Is it supported by older systemd versions, meaning at least that unknown options are ignored instead of causing errors?

Since those option restrict what a service is allowed to do it should not break if those options are not present, it's just less secure.

Mic92:
Why is this marked as WIP? What is missing?

andryyy:
I am sure some of these will cause trouble on _some_ systems.
I am absolutely not a fan of copying from "hardening" guides. If you don't fully know when which option may break something or what it does on individual systems with different backends, you should not copy it. Same with all these HTTP hardening guides people don't even fully understand and still copy so much non sense. 

Some of these options may have nasty side effects you will probably not easily get behind...

Just my two cents.

Mic92:
> I am sure some of these will cause trouble on _some_ systems.
> I am absolutely not a fan of copying from "hardening" guides. If you don't fully know when which option may break something or what it does on individual systems with different backends, you should not copy it. Same with all these HTTP hardening guides people don't even fully understand and still copy so much non sense.
> 
> Some of these options may have nasty side effects you will probably not easily get behind...
> 
> Just my two cents.

We not just blindly coping security options we found on stackoverflow here. We did a review based on our knowledge. We collected some experience with systemd hardening options in other services in nixpkgs. Still we are not expert of all internals of rspamd hence we ask upstream to carefully review them as well. The hope is that if rspamd ever needs more privileges they will update the service file as well.

tnias:
> Why is this marked as WIP? What is missing?

I have only really tested this on my NixOS mailserver. People with other setups should probably test it as well. Otherwise it should be fine.

vstakhov:
> > I am sure some of these will cause trouble on _some_ systems.
> > I am absolutely not a fan of copying from "hardening" guides. If you don't fully know when which option may break something or what it does on individual systems with different backends, you should not copy it. Same with all these HTTP hardening guides people don't even fully understand and still copy so much non sense.
> > Some of these options may have nasty side effects you will probably not easily get behind...
> > Just my two cents.
> 
> We not just blindly coping security options we found on stackoverflow here. We did a review based on our knowledge. We collected some experience with systemd hardening options in other services in nixpkgs. Still we are not expert of all internals of rspamd hence we ask upstream to carefully review them as well. The hope is that if rspamd ever needs more privileges they will update the service file as well.

The problem is that I'm not a systemd expert. I don't understand what these statements do and how could they affect Rspamd...

Mic92:
> The problem is that I'm not a systemd expert. I don't understand what these statements do and how could they affect Rspamd...

We can go over them one by one. The manpages are also on point. All options used here are defined in https://www.freedesktop.org/software/systemd/man/systemd.exec.html

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
 ExecStart=/usr/bin/rspamd -c /etc/rspamd/rspamd.conf -f
 ExecReload=/bin/kill -HUP $MAINPID
 User=_rspamd
+ConfigurationDirectory=rspamd
```
Mic92, line 12: This crates `/etc/rspamd` owned by root:root with 0655.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
 RuntimeDirectory=rspamd
 RuntimeDirectoryMode=0755
+StateDirectory=rspamd
+StateDirectoryMode=0700
```
Mic92, line 16:
```suggestion
StateDirectory=rspamd
StateDirectoryMode=0700
```

This creates `/var/lib/rspamd` owned by the rspamd user with permissions 0700.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+LogsDirectory=rspamd
```
Mic92, line 17: This creates `/var/log/rspamd` owned by the rspamd user.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
 Restart=always
 
+AmbientCapabilities=
+CapabilityBoundingSet=
```
Mic92, line 21: Start the service without any capabilities such as `CAP_SYS_ADMIN`, `CAP_DAC_OVERRIDE`...

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+DevicePolicy=closed
```
Mic92, line 22: Rspamd will be only able to access the following pseudo devices:

/dev/null, /dev/zero, /dev/full, /dev/random, and /dev/urandom.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+LockPersonality=true
```
Mic92, line 23: Rspamd cannot change its [personality(2)](https://man7.org/linux/man-pages/man2/personality.2.html) i.e. execute 32-bit binaries or disable ASLR.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+NoNewPrivileges=true
```
Mic92, line 24: No setuid and friends.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+PrivateDevices=true
```
Mic92, line 25: A new private `/dev` will be created for rspamd.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+PrivateMounts=true
```
Mic92, line 26: Puts rspamd in a private mount namespace.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+PrivateTmp=true
```
Mic92, line 27: Creates a new `/tmp` where only rspamd has access to.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+PrivateUsers=true
```
Mic92, line 28: Creates a new usernamespace where only the rspamd and root exits. Everything else is mapped to "nobody".

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectClock=true
```
Mic92, line 29: Rspamd cannot change the system time.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectControlGroups=true
```
Mic92, line 30: Rspamd cannot create/modify cgroups

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectHome=true
```
Mic92, line 31: `/home` is not accessible to the service.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectHostname=true
```
Mic92, line 32: Rspamd cannot change the host name.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectKernelLogs=true
+ProtectKernelModules=true
```
Mic92, line 34: Cannot load kernel modules.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+NoNewPrivileges=true
+PrivateDevices=true
+PrivateMounts=true
+PrivateTmp=true
+PrivateUsers=true
+ProtectClock=true
+ProtectControlGroups=true
+ProtectHome=true
+ProtectHostname=true
+ProtectKernelLogs=true
```
Mic92, line 33: Disables access to dmesg.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectKernelModules=true
+ProtectKernelTunables=true
```
Mic92, line 35: Rspamd cannot use sysctl.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectSystem=strict
```
Mic92, line 36: Makes everything but `/var/lib/rspamd` readonly.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+RemoveIPC=true
```
Mic92, line 37: Disables System V and POSIX IPC apis.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
```
Mic92, line 38: Only allows ip/ipv6 and unix sockets.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+RestrictNamespaces=true
```
Mic92, line 39: Does not allow rspamd to create additional namespaces.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+RestrictRealtime=true
+RestrictSUIDSGID=true
```
Mic92, line 41: Does not allow rspamd to use setuid/getuid binaries.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+ProtectHome=true
+ProtectHostname=true
+ProtectKernelLogs=true
+ProtectKernelModules=true
+ProtectKernelTunables=true
+ProtectSystem=strict
+RemoveIPC=true
+RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
+RestrictNamespaces=true
+RestrictRealtime=true
```
Mic92, line 40: Rspamd cannot apply realtime policies such as SCHED_FIFO, SCHED_RR or SCHED_DEADLINE

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+RestrictSUIDSGID=true
+SystemCallArchitectures=native
```
Mic92, line 42: Only allow native system calls. For example on x86_64-linux only x86_64 system calls are allowed but no x86 legacy system calls.

andryyy:
I think the problem is not what these option do, but the side effects they can have.......

vstakhov:
Disabling IPC is definitely stupid as it is used by Rspamd. LuaJIT actively uses JIT which is not compatible with W^X models. TBH, this all look very poor as it tries to block bad things instead of allowing things that are really needed. I really miss OpenBSD pledge support in Linux.
Aside that, I see nothing bad but I agree with @andryyy - there could be breaking side effects.

rspamd.service
```diff
@@ -9,9 +9,39 @@ NonBlocking=true
...
+SystemCallFilter=@system-service
```
Mic92, line 43: Full list of all allowed system calls:

- "io_cancel\0"
- "io_destroy\0"
- "io_getevents\0"
- "io_pgetevents\0"
- "io_pgetevents_time64\0"
- "io_setup\0"
- "io_submit\0"
- "io_uring_enter\0"
- "io_uring_register\0"
- "io_uring_setup\0"
- "_llseek\0"
- "close\0"
- "dup\0"
- "dup2\0"
- "dup3\0"
- "lseek\0"
- "pread64\0"
- "preadv\0"
- "preadv2\0"
- "pwrite64\0"
- "pwritev\0"
- "pwritev2\0"
- "read\0"
- "readv\0"
- "write\0"
- "writev\0"
- "chown\0"
- "chown32\0"
- "fchown\0"
- "fchown32\0"
- "fchownat\0"
- "lchown\0"
- "lchown32\0"
- "clock_getres\0"
- "clock_getres_time64\0"
- "clock_gettime\0"
- "clock_gettime64\0"
- "clock_nanosleep\0"
- "clock_nanosleep_time64\0"
- "execve\0"
- "exit\0"
- "exit_group\0"
- "futex\0"
- "futex_time64\0"
- "get_robust_list\0"
- "get_thread_area\0"
- "getegid\0"
- "getegid32\0"
- "geteuid\0"
- "geteuid32\0"
- "getgid\0"
- "getgid32\0"
- "getgroups\0"
- "getgroups32\0"
- "getpgid\0"
- "getpgrp\0"
- "getpid\0"
- "getppid\0"
- "getresgid\0"
- "getresgid32\0"
- "getresuid\0"
- "getresuid32\0"
- "getrlimit\0"      /* make sure processes can query stack size and such */
- "getsid\0"
- "gettid\0"
- "gettimeofday\0"
- "getuid\0"
- "getuid32\0"
- "membarrier\0"
- "nanosleep\0"
- "pause\0"
- "prlimit64\0"
- "restart_syscall\0"
- "rseq\0"
- "rt_sigreturn\0"
- "sched_yield\0"
- "set_robust_list\0"
- "set_thread_area\0"
- "set_tid_address\0"
- "set_tls\0"
- "sigreturn\0"
- "time\0"
- "ugetrlimit\0"
- "access\0"
- "chdir\0"
- "chmod\0"
- "close\0"
- "creat\0"
- "faccessat\0"
- "fallocate\0"
- "fchdir\0"
- "fchmod\0"
- "fchmodat\0"
- "fcntl\0"
- "fcntl64\0"
- "fgetxattr\0"
- "flistxattr\0"
- "fremovexattr\0"
- "fsetxattr\0"
- "fstat\0"
- "fstat64\0"
- "fstatat64\0"
- "fstatfs\0"
- "fstatfs64\0"
- "ftruncate\0"
- "ftruncate64\0"
- "futimesat\0"
- "getcwd\0"
- "getdents\0"
- "getdents64\0"
- "getxattr\0"
- "inotify_add_watch\0"
- "inotify_init\0"
- "inotify_init1\0"
- "inotify_rm_watch\0"
- "lgetxattr\0"
- "link\0"
- "linkat\0"
- "listxattr\0"
- "llistxattr\0"
- "lremovexattr\0"
- "lsetxattr\0"
- "lstat\0"
- "lstat64\0"
- "mkdir\0"
- "mkdirat\0"
- "mknod\0"
- "mknodat\0"
- "mmap\0"
- "mmap2\0"
- "munmap\0"
- "newfstatat\0"
- "oldfstat\0"
- "oldlstat\0"
- "oldstat\0"
- "open\0"
- "openat\0"
- "readlink\0"
- "readlinkat\0"
- "removexattr\0"
- "rename\0"
- "renameat\0"
- "renameat2\0"
- "rmdir\0"
- "setxattr\0"
- "stat\0"
- "stat64\0"
- "statfs\0"
- "statfs64\0"
-#x
- "statx\0"
-#
- "symlink\0"
- "symlinkat\0"
- "truncate\0"
- "truncate64\0"
- "unlink\0"
- "unlinkat\0"
- "utime\0"
- "utimensat\0"
- "utimensat_time64\0"
- "utimes\0"
- "_newselect\0"
- "epoll_create\0"
- "epoll_create1\0"
- "epoll_ctl\0"
- "epoll_ctl_old\0"
- "epoll_pwait\0"
- "epoll_wait\0"
- "epoll_wait_old\0"
- "eventfd\0"
- "eventfd2\0"
- "poll\0"
- "ppoll\0"
- "ppoll_time64\0"
- "pselect6\0"
- "pselect6_time64\0"
- "select\0"
- "ipc\0"
- "memfd_create\0"
- "mq_getsetattr\0"
- "mq_notify\0"
- "mq_open\0"
- "mq_timedreceive\0"
- "mq_timedreceive_time64\0"
- "mq_timedsend\0"
- "mq_timedsend_time64\0"
- "mq_unlink\0"
- "msgctl\0"
- "msgget\0"
- "msgrcv\0"
- "msgsnd\0"
- "pipe\0"
- "pipe2\0"
- "process_vm_readv\0"
- "process_vm_writev\0"
- "semctl\0"
- "semget\0"
- "semop\0"
- "semtimedop\0"
- "semtimedop_time64\0"
- "shmat\0"
- "shmctl\0"
- "shmdt\0"
- "shmget\0"
- "add_key\0"
- "keyctl\0"
- "request_key\0"
- "mlock\0"
- "mlock2\0"
- "mlockall\0"
- "munlock\0"
- "munlockall\0"
- "accept\0"
- "accept4\0"
- "bind\0"
- "connect\0"
- "getpeername\0"
- "getsockname\0"
- "getsockopt\0"
- "listen\0"
- "recv\0"
- "recvfrom\0"
- "recvmmsg\0"
- "recvmmsg_time64\0"
- "recvmsg\0"
- "send\0"
- "sendmmsg\0"
- "sendmsg\0"
- "sendto\0"
- "setsockopt\0"
- "shutdown\0"
- "socket\0"
- "socketcall\0"
- "socketpair\0"
- "arch_prctl\0"
- "capget\0"      /* Able to query arbitrary processes */
- "clone\0"
- "clone3\0"
- "execveat\0"
- "fork\0"
- "getrusage\0"
- "kill\0"
- "pidfd_open\0"
- "pidfd_send_signal\0"
- "prctl\0"
- "rt_sigqueueinfo\0"
- "rt_tgsigqueueinfo\0"
- "setns\0"
- "swapcontext\0" /* Some archs e.g. powerpc32 are using it to do userspace context switches */
- "tgkill\0"
- "times\0"
- "tkill\0"
- "unshare\0"
- "vfork\0"
- "wait4\0"
- "waitid\0"
- "waitpid\0"
- "ioprio_set\0"
- "mbind\0"
- "migrate_pages\0"
- "move_pages\0"
- "nice\0"
- "sched_setaffinity\0"
- "sched_setattr\0"
- "sched_setparam\0"
- "sched_setscheduler\0"
- "set_mempolicy\0"
- "setpriority\0"
- "setrlimit\0"
- "setgid\0"
- "setgid32\0"
- "setgroups\0"
- "setgroups32\0"
- "setregid\0"
- "setregid32\0"
- "setresgid\0"
- "setresgid32\0"
- "setresuid\0"
- "setresuid32\0"
- "setreuid\0"
- "setreuid32\0"
- "setuid\0"
- "setuid32\0"
- "rt_sigaction\0"
- "rt_sigpending\0"
- "rt_sigprocmask\0"
- "rt_sigsuspend\0"
- "rt_sigtimedwait\0"
- "rt_sigtimedwait_time64\0"
- "sigaction\0"
- "sigaltstack\0"
- "signal\0"
- "signalfd\0"
- "signalfd4\0"
- "sigpending\0"
- "sigprocmask\0"
- "sigsuspend\0"
- "fdatasync\0"
- "fsync\0"
- "msync\0"
- "sync\0"
- "sync_file_range\0"
- "sync_file_range2\0"
- "syncfs\0"
- "alarm\0"
- "getitimer\0"
- "setitimer\0"
- "timer_create\0"
- "timer_delete\0"
- "timer_getoverrun\0"
- "timer_gettime\0"
- "timer_gettime64\0"
- "timer_settime\0"
- "timer_settime64\0"
- "timerfd_create\0"
- "timerfd_gettime\0"
- "timerfd_gettime64\0"
- "timerfd_settime\0"
- "timerfd_settime64\0"
- "times\0"
- "brk\0"
- "capget\0"
- "capset\0"
- "copy_file_range\0"
- "fadvise64\0"
- "fadvise64_64\0"
- "flock\0"
- "get_mempolicy\0"
- "getcpu\0"
- "getpriority\0"
- "getrandom\0"
- "ioctl\0"
- "ioprio_get\0"
- "kcmp\0"
- "madvise\0"
- "mprotect\0"
- "mremap\0"
- "name_to_handle_at\0"
- "oldolduname\0"
- "olduname\0"
- "personality\0"
- "readahead\0"
- "readdir\0"
- "remap_file_pages\0"
- "sched_get_priority_max\0"
- "sched_get_priority_min\0"
- "sched_getaffinity\0"
- "sched_getattr\0"
- "sched_getparam\0"
- "sched_getscheduler\0"
- "sched_rr_get_interval\0"
- "sched_rr_get_interval_time64\0"
- "sched_yield\0"
- "sendfile\0"
- "sendfile64\0"
- "setfsgid\0"
- "setfsgid32\0"
- "setfsuid\0"
- "setfsuid32\0"
- "setpgid\0"
- "setsid\0"
- "splice\0"
- "sysinfo\0"
- "tee\0"
- "umask\0"
- "uname\0"
- "userfaultfd\0"
- "vmsplice\0"

Mic92:
> Disabling IPC is definitely stupid as it is used by Rspamd. LuaJIT actively uses JIT which is not compatible with W^X models. TBH, this all look very poor as it tries to block bad things instead of allowing things that are really needed. I really miss OpenBSD pledge support in Linux.
> Aside that, I see nothing bad but I agree with @andryyy - there could be breaking side effects.

Ok. Than we drop the IPC option. Thanks for letting us know. JIT should still work as far as I can see. The point of this pull request to only allow what is needed.

Mic92:
> I think the problem is not what these option do, but the side effects they can have.......

Is there anything anything unclear from the explanation I provided?

Mic92:
I corrected my statement about `RemoveIPC`, the option does something else.

Mic92:
also cc @alucryd as a archlinux maintainer and @dottedmag as a debian maintainer

tnias:
> LuaJIT actively uses JIT which is not compatible with W^X models.

The W^X is explicitly not enabled by this patch. See `MemoryDenyWriteExecute` in systemd.exec (5), which is not used.

> TBH, this all look very poor as it tries to block bad things instead of allowing things that are really needed. I really miss OpenBSD pledge support in Linux.

This PR is functionally somewhat similar to openbsd's unveil as it is mostly concerned about filesystem access and not so much concerned about systemcalls.

The systemcall filter is more similar to pledge. In my entry comment I already  said "the SystemCallFilter could be tightened down", I would like to create a minimal allowlist. But I would probably need support from folks with different setups to cover all usecases as my setup uses only a pretty limited feature set. Also I think this should be done in a different PR.

bjoe2k4:
I've applied this patch to my rspamd@archlinux and could not see any initial problems in the logs. I'll let it run for 24 more hours....

bjoe2k4:
I couldn't observe any error with this patch.

alucryd:
Looks fine on my server as well. You might consider using DynamicUser as well, that way we can get rid of sysusers, and also rename _rspamd to rspamd. Not sure what the reasoning behind the underscore was, system users rarely, if at all, start with an underscore.

Mic92:
On NixOS we also use `rspamd` instead of `_rspamd`. I am not sure if DynamicUser works out here. There is a socket in `/var/lib/rspamd/rspamd.sock`, which will be owned by user `rspamd` and people might need the same user to connect to the service with cli. Using `StateDirectory` should also take care of user changes. I notice however if the directory was previously owned by rspamd already it won't recursivily chown the directory.

Mic92:
So I count at least 4 different systems where this has been tested (2 x Archlinux + 2 x NixOS installations)

tnias:
I just deployed it to a Debian 10 install. On Debian 10 it needs the `mincore` syscall.
And I will test on CentOS 8 soon.

mweinelt:
As this has been tested on different systems, can this please move on? These are quite some useful improvements.

tnias:
Update: TheCentOS 8 setup is now running stable for over a month.

Mic92:
I think @dasJ has a similar sandboxing feature set for rspamd in production on NixOS for a while.

dasJ:
`systemctl cat rspamd`:

```
AppArmorProfile=systemd-rspamd
CapabilityBoundingSet=
ExecStart=/nix/store/5cyasp8h423v4qfdvwa59h2v01gikvlr-rspamd-2.5/bin/rspamd  -c /etc/rspamd/rspamd.conf -f
Group=rspamd
LockPersonality=true
MemoryDenyWriteExecute=false
NoNewPrivileges=true
PrivateDevices=true
PrivateMounts=true
PrivateNetwork=false
PrivateTmp=true
PrivateUsers=true
ProtectClock=true
ProtectControlGroups=true
ProtectHome=true
ProtectHostname=true
ProtectKernelLogs=true
ProtectKernelModules=true
ProtectKernelTunables=true
ProtectSystem=strict
Restart=always
RestrictAddressFamilies=
RestrictRealtime=true
RestrictSUIDSGID=true
RuntimeDirectory=rspamd
StateDirectory=rspamd
SystemCallArchitectures=native
SystemCallFilter=@basic-io @file-system @network-io @system-service
User=rspamd
```

Mic92:
Maybe we could create a new project for systemd and remove the service file from the repo, if rspamd maintainers don't want to maintain systemd service files?

andryyy:
I don't see Vsevolod not wanting to maintain them.

Mic92:
> I don't see Vsevolod not wanting to maintain them.

To quote Vsevolod from https://github.com/rspamd/rspamd/pull/3442#issuecomment-663454011
> The problem is that I'm not a systemd expert. I don't understand what these statements do and how could they affect Rspamd...

In NixOS and many other Linux distributions maintainer have build up this expertise because it is the core of our systems. I see it problematic having not optimal settings upstream because of the default effect.

andryyy:
These files are just filled with every available option and you make it sound like it is bad or insecure not to follow your advice (no offense, it just feels like that for me :)). It may have bad side-effects on some systems. There are defaults in systemd you can set for your services **on your system**.

If you think that's a universal best-practice, you should create a PR @ Debian, Ubuntu etc.

I don't think that's a proper way to go. One should not force a personal preference on all systems. The systemd files provided with a package should never overrule the administrators choice.

What if there's a bug with a random systemd version not interpreting a particular option correctly? There are many "ifs" and I would prefer to stick with the operation systems default configuration and, if I like and made my choice, override or add an option for all or some unit files. :)

I will leave the discussion now. In the end it's up to Vsevo.

Mic92:
> These files are just filled with every available option and you make it sound like it is bad or insecure not to follow your advice (no offense, it just feels like that for me :)). It may have bad side-effects on some systems. There are defaults in systemd you can set for your services **on your system**.
> 
> If you think that's a universal best-practice, you should create a PR @ Debian, Ubuntu etc.
> 
> I don't think that's a proper way to go. One should not force a personal preference on all systems. The systemd files provided with a package should never overrule the administrators choice.
> 
> What if there's a bug with a random systemd version not interpreting a particular option correctly? There are many "ifs" and I would prefer to stick with the operation systems default configuration and, if I like and made my choice, override or add an option for all or some unit files. :)
> 
> I will leave the discussion now. In the end it's up to Vsevo.

Sorry but we did testing on multiple systems and got it in contact with multiple distribution maintainers to make sure it works. What you are bringing up is FUD here that is not backed by any evidence.

Issue #3391, opened 2020-06-04
[BUG] rspamadm vault roll -r example.com does create new keys


croessner:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
If you have create DKIM keys in a vault/consul database and it comes to the first time, where you decide to rollover the keys, the expired keys will be removed but unfortunately the same command creates a new key. Therefor you always have two keys (or more if ed25519, too) per domain.

### Steps to Reproduce

1. rspamadm vault new ... somedomain

2. Wait a month, whatever time period for rollover

3. rspamadm vault roll somedomain
This time a new key is generated and the old keys gets an expire field

4. Wait some days and try to remove _only_ expired keys:

I show this on a existing domain Note the given '-r' flag!:

rspamadm vault roll -r alltagswahnsinn.de
removed expired key for alltagswahnsinn.de (selector rsa-20200512, expire "Fri 15 May 2020 11:10:35 AM CEST"
removed expired key for alltagswahnsinn.de (selector ed25519-20200512, expire "Fri 15 May 2020 11:10:35 AM CEST"
rolled key for: alltagswahnsinn.de, new selector: rsa-20200604
please place the corresponding public key as following:
rsa-20200604._domainkey IN TXT ( "v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAs8vLW/xfEScnOwafL9xc0Qg75E6Iz
[...]

PR #3447, opened 2020-07-20
[Minor] rspamadm roll -r must not generate new keys

croessner:
This patch address this bug: https://github.com/rspamd/rspamd/issues/3391

Please review


vstakhov:
That looks good to me indeed, thank you! Is this patch tested and ready to be merged?

PR #3468, opened 2020-08-10
[Feature] Add BOUNCE rule

citrin:


rules/bounce.lua
```diff
@@ -0,0 +1,117 @@
...
+      return false
+    end
+
+    if bounce_sender and subj_keywords then
+      return true, 0.5, bounce_sender .. '+subj'
+    end
+
+    -- Look for a message/rfc822(-headers) part inside
+    local rfc822_part
+    parts[10] = nil -- limit numbe of parts to check
```
vstakhov, line 97: small misprint - `numbe`

PR #3508, opened 2020-10-06
Url check

lprat:
---
layout: doc
title: URL Check module
---

# URL check module

The URL checking plugin performs several analysis of the urls extracted from your messages:
  - Static analysis:
    - Check if URL uses no standard port
    - Check if URL uses direct IP (without hostname)
    - Check if URL uses suspicious PATH name
    - Check if URL uses suspicious sub domaine name (length and name)
    - Check if URL is using a free download site
    - Check if URL is using a freehosting service
  - Dynamics analysis (data persisted in Redis):
    - Check Resolv IP and if IP is in the blacklist
    - Follow redirect (code 301/302) and check if exceeded the number of redirects (defined)
    - Check mimetype and filename and if they are in the blacklist
    - Check if domain has no MX record
    - Check if hostname resolution is the same as tld resolution
  - Reputation (Threat Intel)
    - RBL

For it to be enabled, the following should be set in `/etc/rspamd/local.d/url_check.conf`:
~~~ucl
enabled = true;
~~~

# Configuration

The following settings can be configured for this module:

~~~ucl
# How long to cache dereferenced links in Redis (default 1 day)
expire = 1d;

# Timeout for HTTP requests (4 seconds by default)
timeout = 4; # warning: adapt "task_timeout" in global option

# check all url, not just redirector_hosts_map (default true)
all_urls = true;

# How many nested redirects to follow (default 2)
nested_limit = 2;

# Prefix for keys in redis (default "urli:")
key_prefix = "urli:";

# Check SSL certificates (default false)
check_ssl = false;

# How many urls to check  (default false)
max_urls = 10;

# Maximum body to process
max_size = 10k;

# no check internal url/ip (default false)
internal_url = false;

# check mx for domain (default false)
mx_check = true;

# check if resolv A host is same than A domain (default false)
a_check = true;

# Follow merely redirectors
redirectors_only = false;

# Insert symbol when the checker has been found
#SYMBOL IP blacklist
ip_blacklist_symbol = "URL_RESOLV_IP_IN_BLACKLIST";
#SYMBOL URL is redirected
redirector_symbol = "URL_REDIRECT";
#SYMBOL URL contains a file to download
download_symbol = "URL_DOWNLOAD";
#SYMBOL URL contains a suspect mimetype
mimetype_symbol = "URL_SUSPECT_MIMETYPE";
#SYMBOL URL contains a suspect filename/extension
filename_symbol = "URL_SUSPECT_FILENAME";
#SYMBOL URL contains a suspect path
path_symbol = "URL_SUSPECT_PATH";
#SYMBOL URL uses no standard port
port_symbol = "URL_WITH_NO_STANDARD_PORT";
#SYMBOL URL uses direct IP without hostname
ip_symbol = "URL_WITH_IP";
#SYMBOL URL contains suspect sub domain
subhost_suspect_symbol = "URL_WITH_SUSPECT_SUBHOST";
#SYMBOL URL use freehosting
freehost_symbol = "URL_USE_FREEHOST";
#SYMBOL URL use domain without MX record
mx_symbol = "URL_WITH_DOMAIN_NO_MX";
#SYMBOL Hostname resolution is the same as tld resolution
a_symbol = "URL_WITH_HOST_RESOLV_DOMAIN";
#SYMBOL sub domain has lot of part
subhost_max_part_symbol = "SUBHOST_HAS_LOTOF_PART"
#SYMBOL URL is using a free download site
free_download_symbol = "URL_DOWNLOAD_ON_FREESITE";
#SYMBOL URL exceeded the number of redirects
too_much_redir_symbol = "URL_TOO_MUCH_REDIRECT";

# Check only those redirectors
#redirector_hosts_map = "${LOCAL_CONFDIR}/local.d/maps.d/redirectors.map";

#Mimetype Check
## check mimetype in url header
suspect_mimetype_in_url_map = "${LOCAL_CONFDIR}/local.d/maps.d/suspect_mimetype_url.map";
## check mimetype map only allowed -- if false then map is checked in mode blacklist
mimetype_url_only_allow = true;
## if suspect_mimetype_in_url_map is map type regexp
mimetype_map_regexp = true;

# check filename (content-disposition) in url -- regexp map
suspect_filename_in_url_map = "${LOCAL_CONFDIR}/local.d/maps.d/suspect_filname_url.map";

# Check suspect path
suspect_path_in_url_map = "${LOCAL_CONFDIR}/local.d/maps.d/suspect_path_url.map";

#free hosted domain
free_hosted_dom_map = "${LOCAL_CONFDIR}/local.d/maps.d/freehosting.map";

#Subhost check
##sub host max part for no suspect
sub_host_number_part_max = 2;
##subhost name suspect: paypal, bank, facebook...
sub_host_suspect_map = "${LOCAL_CONFDIR}/local.d/maps.d/subhost.map";

#host to use get method
host_use_get_map = "${LOCAL_CONFDIR}/local.d/maps.d/host_use_get.map";

#ip black list
url_ip_blacklist_map = "${LOCAL_CONFDIR}/local.d/maps.d/ip_bl.map";

~~~

### DIFF
--- a/conf/modules.d/url_check.conf
+++ b/conf/modules.d/url_check.conf
@@ -10,7 +10,7 @@
 # See https://rspamd.com/doc/faq.html#what-are-the-locald-and-overrided-directories
 # for details
 #
-# Module documentation can be found at  https://rspamd.com/doc/modules/url_check.html
+# Module documentation can be found at https://rspamd.com/doc/modules/url_check.html
 
 url_check {
   expire = 1d; # 1 day by default
--- a/conf/modules.d/url_check.conf
+++ b/conf/modules.d/url_check.conf
@@ -20,7 +20,12 @@ url_check {
   key_prefix = "urli:"; # default hash name
   check_ssl = false; # check ssl certificates
   max_size = 10k; # maximum body to process
-
+  all_urls = true; # check all url
+  internal_url = false; #no check url/ip internal
+  mx_check = false; #no check mx domain
+  a_check = false; # no check A resolv tld & host
+  redirectors_only = false; # Follow merely redirectors
+  
   .include(try=true,priority=5) "${DBDIR}/dynamic/url_check.conf"
   .include(try=true,priority=1,duplicate=merge) "$LOCAL_CONFDIR/local.d/url_check.conf"
   .include(try=true,priority=10) "$LOCAL_CONFDIR/override.d/url_check.conf"
--- a/conf/modules.d/url_check.conf
+++ b/conf/modules.d/url_check.conf
@@ -1,0 +1,27 @@
+# Please don't modify this file as your changes might be overwritten with
+# the next update.
+#
+# You can modify 'local.d/url_check.conf' to add and merge
+# parameters defined inside this section
+#
+# You can modify 'override.d/url_check.conf' to strictly override all
+# parameters defined inside this section
+#
+# See https://rspamd.com/doc/faq.html#what-are-the-locald-and-overrided-directories
+# for details
+#
+# Module documentation can be found at  https://rspamd.com/doc/modules/url_check.html
+
+url_check {
+  expire = 1d; # 1 day by default
+  timeout = 4; # warning: adapt "task_timeout" global option
+  nested_limit = 2; # How many redirects to follow
+  #proxy = "http://example.com:3128"; # Send request through proxy
+  key_prefix = "urli:"; # default hash name
+  check_ssl = false; # check ssl certificates
+  max_size = 10k; # maximum body to process
+
+  .include(try=true,priority=5) "${DBDIR}/dynamic/url_check.conf"
+  .include(try=true,priority=1,duplicate=merge) "$LOCAL_CONFDIR/local.d/url_check.conf"
+  .include(try=true,priority=10) "$LOCAL_CONFDIR/override.d/url_check.conf"
+}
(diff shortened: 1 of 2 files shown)


vstakhov:
The vast majority of the checks could be done by multimap + selectors as far as I see. I don't like the idea of plugins that do too much of different and not very related things and I prefer to have one tool to rule all similar tasks (e.g. multimap or selectors). Maybe this work should be transformed to a set of multimap rules instead?

lprat:
Hi,

Indeed the "static" part, could surely be done by multipmap + selectors.

Initially, I wanted to retrieve information in a "dynamic" way such as content-type, content-disposition, in order to limit malspam campaigns ... 

The basic configuration of RSPAMD mainly allows this protection against spoofed URL (phishing check) and URL known in blacklists (RBL / Threat Intel), as well as URLs hidden behind a public redirector (eg: bit.ly). If the url is not known by its reputation and there is no "spoofing URL" then by default (to my knowledge) rspamd does not have any verification on these URLs. It is this lack, which I wanted to fill with this plugin in order to have a default configuration making possible a better analysis of the risk coming from a URL.

I found that it was better to use the connections potentially made by the url_redirector module. After some thought, finally, the creation of a new module seems clearer.

You say: "I don't like the idea of plugins that do too much of different and not very related things", but it only analyzes all the properties (code, port, path, subdom, content-type, domain mx, ...) that the "object" url has in a message. Like your other modules: MID, MX check, SPF, mime type, ...

I could delete the "static" search part, but I find it more understandable to have modules to check each objet (url, mimetype, attachment, ...), although I am well aware that url_redirector and phishing check are close.

The use of selector + multimap seems to me more appropriate on analyzes without any link as you indicate.

sorry for my English...

PR #3509, opened 2020-10-06
[MID] check TLD hostname

lprat:
Hi,

I observed that on emails coming from gmail the symbol "MID_RHS_MATCH_FROM" was triggered. Gmail uses the hostname "@mail.gmail.com" in MID.
I therefore suggest you add a symbol "MID_RHS_MATCH_FROMTLD",  to compare the TLD of hostnames in FROM and MID.
Thanks.

### DIFF
--- a/rules/mid.lua
+++ b/rules/mid.lua
@@ -14,7 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ]]--
-local util = require "rspamd_util"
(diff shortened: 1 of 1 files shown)


rules/mid.lua
```diff
@@ -14,7 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ]]--
-
+local util = require "rspamd_util"
```
vstakhov, line 17: I would suggest to name variable as `rspamd_util` not `util` to avoid further ambiguity with `lua_util`.

PR #3513, opened 2020-10-12
[Feature] RBL: support use of multiple selectors

fatalbanana:


src/plugins/lua/rbl.lua
```diff
@@ -879,26 +881,42 @@ local function add_rbl(key, rbl, global_opts)
...
-      -- Create a new flattened closure
-      local sel = selectors.create_selector_closure(rspamd_config, rbl.selector, '', true)
 
-      if not sel then
-        rspamd_logger.errx('invalid selector for rbl rule %s: %s', key, rbl.selector)
-        return false
-      end
+    rbl.selectors = {}
+    if type(rbl.selector) ~= 'table' then
+      rbl.selector = {[''] = rbl.selector}
```
vstakhov, line 887: I would prefer to have some meaningful name here, e.g. `_default` to distinguish from user specified names.

src/plugins/lua/rbl.lua
```diff
@@ -879,26 +881,42 @@ local function add_rbl(key, rbl, global_opts)
...
-      rbl.selector = sel
-      known_selectors[rbl.selector] = {
-        selector = sel,
-        id = #lua_util.keys(known_selectors) + 1,
-      }
-      rbl.selector_id = known_selectors[rbl.selector].id
+    for selector_label, selector in pairs(rbl.selector) do
+      if known_selectors[selector] then
+        if selector_label == '' then
+          selector_label = 'sel' .. known_selectors[selector].id
```
vstakhov, line 893: Same: instead of using `sel` as a prefix we can use `_default` in the same way and remove extra `if` branch.

src/plugins/lua/rbl.lua
```diff
@@ -1188,7 +1206,7 @@ local rule_schema_tbl = {
   return_codes = return_codes_schema:is_optional(),
   returnbits = return_bits_schema:is_optional(),
   returncodes = return_codes_schema:is_optional(),
-  selector = ts.string:is_optional(),
+  selector = ts.one_of{ts.string:is_optional(), ts.table:is_optional()},
```
vstakhov, line 1209: It should probably be `ts.one_of{ts.string, ts.table}:is_optional(),`

src/plugins/lua/rbl.lua
fatalbanana: 👍 it's called `selector` now.

I thought about introducing userdata to strongly distinguish it.

Issue #3514, opened 2020-10-13
[BUG] ignored systemd option in logging.inc


jendis:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
`systemd` option in _logging.inc_ file is ignored, so each log message starts with timestamp.

### Steps to Reproduce
Add `systemd = true;` to _/etc/rspamd/local.d/logging.inc_ (or even _/etc/rspamd/logging.inc_), run rspamd.

**Expected behavior**
If _/etc/rspamd/local.d/logging.inc_ contains line `systemd = true;`, timestamps aren’t prepended to log messages.

### Versions

rspamd-2.6.

fatalbanana:
Do you actually use systemd for logging? (it adds its own timestamps)

I'm unable to reproduce it.

jendis:
No, we don't, but regardless systemd or whatever else, from what I understand from documentation, setting `systemd = true;` in _logging.inc_ conf file should do only one thing which is not prepending timestamp to log message, which is actually not happening if you observe raw logs from rspamd printed to the standard output.
I dif quick review of source code and it seems to me that this option is ignored in fact.

vstakhov:
Should be fixed now.

PR #3515, opened 2020-10-13
[Fix] Honour `systemd` setting when logging to console (#3514)

fatalbanana:


vstakhov:
I don't think it is enough to fix the issue. I think it is better to copy this fragment from logger_file:

```C
if (!(rspamd_log->flags & RSPAMD_LOG_FLAG_SYSTEMD)) {
		log_time (now, rspamd_log, timebuf, sizeof (timebuf));
	}

	cptype = rspamd_log->process_type;
	r = 0;

	if (!(rspamd_log->flags & RSPAMD_LOG_FLAG_SYSTEMD)) {
		r += rspamd_snprintf (tmpbuf + r,
				sizeof (tmpbuf) - r,
				"%s #%P(%s) ",
				timebuf,
				rspamd_log->pid,
				cptype);
	}
	else {
		r += rspamd_snprintf (tmpbuf + r,
				sizeof (tmpbuf) - r,
				"(%s) ",
				cptype);
	}
```

PR #3518, opened 2020-10-14
[Fix] RBL: allow for multiple matches of the same label if types are …

fatalbanana:
…different


src/plugins/lua/rbl.lua
```diff
@@ -339,6 +345,9 @@ local function gen_rbl_callback(rule)
       if forced and not nreq.forced then
         nreq.forced = true
       end
+      if not nreq.what[what] then
```
vstakhov, line 348: Can we please avoid that `what[what]` somehow? It looks terrible.

PR #3519, opened 2020-10-15
[Minor] Selectors: avoid returning empty list from extractors

fatalbanana:


lualib/lua_selectors/extractors.lua
```diff
@@ -241,7 +241,8 @@ The optional second argument accepts list of flags:
   ['received'] = {
     ['get_value'] = function(task, args)
       local rh = task:get_received_headers()
-      if args[1] and rh then
+      if not rh[1] then return end
```
vstakhov, line 244: `return nil` is better I think.

PR #3520, opened 2020-10-16
[Test] Map reload: improve test output

fatalbanana:


vstakhov:
I'd suggest to revert `3s` change and leave it as `1s`.

PR #3542, opened 2020-11-04
[Feature] Clickhouse: store metatokens

fatalbanana:


src/plugins/lua/clickhouse.lua
```diff
@@ -93,6 +94,7 @@ local settings = {
     run_every = '7d',
   },
   extra_columns = {},
+  insert_metatokens = false,
```
vstakhov, line 97: Not sure we need that, as the field is already in the schema.

vstakhov:
@citrin could you also take a look please?

PR #3548, opened 2020-11-10
[Minor] Selectors: Add extractor for metatokens

fatalbanana:


vstakhov:
No, I don't like this approach. In general, we should return a list of strings, not a binary string: something like `fun.map(tostring, metatokens)`. This will create a closure of all numbers packed into strings on a final fold operation. Then we can easily apply the backward transition or compression or whatever we'd like - in general, it should be managed by transforms not extractors.

lualib/lua_selectors/transforms.lua
```diff
@@ -441,6 +441,28 @@ Empty string comes the first argument or 'true', non-empty string comes nil]],
...
   },
+  -- Converts list of strings to numbers and returns a packed string
+  ['pack_numbers'] = {
+    ['types'] = {
+      ['string_list'] = true
+    },
+    ['map_type'] = 'string',
+    ['process'] = function(inp, t, args)
+      local fmt = args[1] or 'f'
+      return fun.foldl(
```
vstakhov, line 452: This is not an efficient approach - it it better to create a table and then apply table.concat (or use rspamd_text.fromtable function)

lualib/lua_selectors/transforms.lua
```diff
@@ -441,6 +441,25 @@ Empty string comes the first argument or 'true', non-empty string comes nil]],
...
+  -- Converts list of strings to numbers and returns a packed string
+  ['pack_numbers'] = {
+    ['types'] = {
+      ['string_list'] = true
+    },
+    ['map_type'] = 'string',
+    ['process'] = function(inp, t, args)
+      local fmt = args[1] or 'f'
+      local res = {}
+      fun.each(
```
vstakhov, line 453: One more thing please: could you just use ipairs here instead of `fun.each`. The former has some use cases but it makes the code more difficult here.

PR #3554, opened 2020-11-17
[Feature] `rspamadm clickhouse` command

fatalbanana:


lualib/rspamadm/clickhouse.lua
```diff
@@ -0,0 +1,247 @@
...
+    if v.seen_spam > v.seen_ham then
+      lower_count = v.seen_ham
+      higher_count = v.seen_spam
+    else
+      lower_count = v.seen_spam
+      higher_count = v.seen_ham
+    end
+    if lower_count / higher_count >= 0.95 then
+      -- Weak ham/spam correlation, remove symbol
+      print(string.format("Removing %s - weak ham/spam correlation", k))
```
vstakhov, line 140: I would suggest to use rspamd_logger.messagex here.

lualib/rspamadm/clickhouse.lua
```diff
@@ -0,0 +1,247 @@
...
+
+  local function override_settings(params)
+    for _, which in ipairs(params) do
+      if cmd_opts[which] == nil then
+        cmd_opts[which] = cfg_opts[which]
+      end
+    end
+  end
+
+  override_settings({
```
vstakhov, line 218: why not lua_util.override_defaults?

lualib/rspamadm/clickhouse.lua
```diff
@@ -0,0 +1,247 @@
...
+
+  if not _r then
+    rspamd_logger.errx('cannot parse %s: %s', config_file, err)
+    os.exit(1)
+  end
+end
+
+local function handle_neural_profile(args)
+  local query = string.format(
+      "SELECT Action, Symbols.Names FROM rspamd WHERE SettingsId = '' ORDER BY TS DESC LIMIT %s",
```
vstakhov, line 83: SettingsId and `WHERE` clause should be configurable. Especially we should select a specific partition by Date field.

lualib/rspamadm/clickhouse.lua
```diff
@@ -0,0 +1,247 @@
...
+    for _, sym in ipairs(r['Symbols.Names']) do
+      local t = known_symbols[sym]
+      if not t then
+        local spam_count, ham_count = 0, 0
+        if is_spam then
+          spam_count = spam_count + 1
+        else
+          ham_count = ham_count + 1
+        end
+        known_symbols[sym] = {
```
vstakhov, line 110: Formatting issue here.

lualib/rspamadm/clickhouse.lua
```diff
@@ -0,0 +1,247 @@
...
+          id = symbols_count,
+	  seen = 1,
+	  seen_ham = ham_count,
+	  seen_spam = spam_count,
+        }
+	symbols_count = symbols_count + 1
+      else
+        known_symbols[sym].seen = known_symbols[sym].seen + 1
+	if is_spam then
```
vstakhov, line 119: Again, wrong tabulation.

lualib/rspamadm/clickhouse.lua
```diff
@@ -0,0 +1,247 @@
...
+
+  -- Walk correlation matrix and divide by total counts
+  local remove = {}
+  for sym_id, row in pairs(correlations) do
+    for inner_sym_id, count in pairs(row) do
+      local known = known_symbols_list[sym_id]
+      if known and count == known.seen then
+        print(string.format("Removing %s - overlapped with %s",
+            known.name, known_symbols_list[inner_sym_id].name))
+        remove[known.name] = true
```
vstakhov, line 185: Maybe we should include a reason here, like `remove = { symbol = name, reason = 'correlated with ' .. corr_name}`

lualib/rspamadm/clickhouse.lua
```diff
@@ -0,0 +1,247 @@
...
+      end
+    end
+  end
+
+  for k in pairs(remove) do
```
vstakhov, line 190: Why do we need to remove it from the table? Just keep two tables and when iterating on `know_symbols` check for `remove` and add a corresponding message. We by the way want to have both symbols that are excluded and symbols that are included + json/ucl format to export those directly to a config file.

lualib/rspamadm/clickhouse.lua
fatalbanana: Probably I forgot about it but reason seems to be that these fields don't promise to be set anywhere.

lualib/rspamadm/clickhouse.lua
fatalbanana: It doesn't seem useful in context of rspamadm. The messages are mainly for debugging purposes so I suppose I'll get rid of many of them.

PR #3561, opened 2020-11-25
[Feature] Multimap: support adding map values as extra options

fatalbanana:


vstakhov:
Hum, the PR name does not correspond to what it does. Could you please clarify?

fatalbanana:
The idea is to provide an alternative use for the values in the maps: where-in they're added as extra options, rather than yielded as symbols.

So one could yield something like:
```MAP_RCPT (1.0) [example.com, Cool Description]```
rather than:
```MAP_RCPT_COOL_DESCRIPTION (1.0) [example.com]```

vstakhov:
Maybe we should set a single option like `example.com=Cool Description`? I'm thinking from Clickhouse analytics point.

vstakhov:
Added this but we might need to think about it one more time...

PR #3574, opened 2020-12-09
[Minor] Selectors: support extracting task cache variables

fatalbanana:


vstakhov:
Please bear in mind that cache can also have tables, so it might worth to check type: if it is a table, then return type `string_list`, otherwise `string` should be fine for both text and plain lua strings.

PR #3577, opened 2020-12-14
[Minor] Move http_headers to plugin

fatalbanana:
Maybe it should be default-disabled

Further, we should ensure this works properly with at least `whitelist` module. Generally, it might be a good thing that results should be as similar as possible to those produced by the actual DMARC/DKIM/SPF modules.


src/plugins/lua/http_headers.lua
```diff
@@ -84,30 +92,43 @@ rspamd_config:add_condition("R_DKIM_ALLOW", function(task)
...
+          task:insert_result(dkim_symbols['symbol_tempfail'], 1.0, 'http header')
+          task:insert_result(dkim_symbols['symbol_trace'], 1.0,
+	      string.format('%s:%s', dkim_domain, dkim_trace.temperror))
+        elseif obj['result'] == 'permfail' then
+          task:insert_result(dkim_symbols['symbol_permfail'], 1.0, 'http header')
+          task:insert_result(dkim_symbols['symbol_trace'], 1.0,
+	      string.format('%s:%s', dkim_domain, dkim_trace.permerror))
+        elseif obj['result'] == 'na' then
+          task:insert_result(dkim_symbols['symbol_na'], 1.0, 'http header')
+        end
```
vstakhov, line 122: In the meantime, can we use `table.field` notation for static fields instead of `table['field']`. Just a style note...

TonyGravagno:
Note that there is no documentation for this module yet, but http_headers.conf still says :
> Module documentation can be found at  https://rspamd.com/doc/modules/http_headers.html

If someone provides some basic info on this here, I'll update the docs and we can fill in details later. Thanks.

fatalbanana:
It's rather special-purpose & probably not useful to the majority of the users - was the thinking behind making this a plugin instead of a rule.

If you have a mailer that performs SPF/DKIM/DMARC checks & you'd want to use those results in Rspamd without re-checking that's what this module is intended to be good for. It might be useful to Haraka users in particular but there is no support in Haraka's Rspamd plugin for sending these results.

This module does not accept any configuration.

Headers which could be added to the scan and their values:

~~~
DKIM: allow|reject|softfail|permfail|na
SPF: allow|reject|neutral|softfail|permfail|na
DMARC: allow|reject|quarantine|tempfail|softfail|permfail|na
~~~

fatalbanana:
Updated my previous post to be more truthful about how it should be used.

PR #3587, opened 2020-12-25
[Minor] Remove check for Arch for hyperscan as available for aarch64 now

a16bitsysop:


a16bitsysop:
Didn’t create branch for PR

PR #3588, opened 2020-12-27
[Feature] allow hyperscan for aarch64

a16bitsysop:


cmake/Hyperscan.cmake
```diff
@@ -1,9 +1,6 @@
 option (ENABLE_HYPERSCAN    "Enable hyperscan for fast regexp processing [default: OFF]" OFF)
 
 if (ENABLE_HYPERSCAN MATCHES "ON")
-    if (NOT ("${ARCH}" STREQUAL "x86_64" OR "${ARCH}" STREQUAL "i386"))
-        MESSAGE(FATAL_ERROR "Hyperscan is supported only on x86_64/i386 architectures")
-    endif ()
```
vstakhov, line 6: Shouldn't we limit it to x86_64, i386 and aarch64 only? I'm not quite sure about that, so it might be removed indeed...

src/hs_helper.c
```diff
@@ -254,12 +254,6 @@ static gboolean
 rspamd_rs_compile (struct hs_helper_ctx *ctx, struct rspamd_worker *worker,
 		gboolean forced)
 {
-	if (!(ctx->cfg->libs_ctx->crypto_ctx->cpu_config & CPUID_SSSE3)) {
-		msg_warn ("CPU doesn't have SSSE3 instructions set "
-				"required for hyperscan, disable hyperscan compilation");
-		return FALSE;
-	}
-
```
vstakhov, line 262: We should leave this check for x86/x86_64 platform.

src/libserver/cfg_rcl.c
```diff
@@ -2053,12 +2053,6 @@ rspamd_rcl_config_init (struct rspamd_config *cfg, GHashTable *skip_sections)
 				G_STRUCT_OFFSET (struct rspamd_config, history_rows),
 				RSPAMD_CL_FLAG_UINT,
 				"Number of records in the history file");
-		rspamd_rcl_add_default_handler (sub,
-				"disable_hyperscan",
-				rspamd_rcl_parse_struct_boolean,
-				G_STRUCT_OFFSET (struct rspamd_config, disable_hyperscan),
-				0,
-				"Disable hyperscan optimizations for regular expressions");
```
vstakhov, line 2061: This should not be removed - it is an used feature to disable HS from the configuration.

src/libserver/cfg_file.h
```diff
@@ -364,7 +364,6 @@ struct rspamd_config {
 	gboolean check_text_attachements;               /**< check text attachements as text					*/
 	gboolean check_all_filters;                     /**< check all filters									*/
 	gboolean allow_raw_input;                       /**< scan messages with invalid mime					*/
-	gboolean disable_hyperscan;                     /**< disable hyperscan usage							*/
```
vstakhov, line 367: This should not be removed - it is an used feature to disable HS from the configuration.

src/libserver/cfg_utils.c
```diff
@@ -816,16 +816,6 @@ rspamd_config_post_load (struct rspamd_config *cfg,
...
 
-#ifdef WITH_HYPERSCAN
-	if (!cfg->disable_hyperscan) {
-		if (!(cfg->libs_ctx->crypto_ctx->cpu_config & CPUID_SSSE3)) {
-			msg_warn_config ("CPU doesn't have SSSE3 instructions set "
-					"required for hyperscan, disable it");
-			cfg->disable_hyperscan = TRUE;
-		}
-	}
-#endif
```
vstakhov, line 827: This should not be removed - it is an used feature to disable HS from the configuration. SSSE3 should also be checked for x86/x86_64 architectures.

src/libserver/maps/map_helpers.c
```diff
@@ -1280,12 +1280,6 @@ rspamd_re_map_finalize (struct rspamd_regexp_map_helper *re_map)
 
 	map = re_map->map;
 
-	if (!(map->cfg->libs_ctx->crypto_ctx->cpu_config & CPUID_SSSE3)) {
-		msg_info_map ("disable hyperscan for map %s, ssse3 instructons are not supported by CPU",
-				map->name);
-		return;
-	}
-
```
vstakhov, line 1288: SSSE3 should also be checked for x86/x86_64 architectures.

src/libserver/re_cache.c
```diff
@@ -126,7 +126,6 @@ struct rspamd_re_cache {
 	lua_State *L;
 #ifdef WITH_HYPERSCAN
 	enum rspamd_hyperscan_status hyperscan_loaded;
-	gboolean disable_hyperscan;
```
vstakhov, line 129: This should not be removed - it is an used feature to disable HS from the configuration.

src/libserver/re_cache.c
```diff
@@ -481,7 +480,6 @@ rspamd_re_cache_init (struct rspamd_re_cache *cache, struct rspamd_config *cfg)
 	const gchar *platform = "generic";
 	rspamd_fstring_t *features = rspamd_fstring_new ();
 
-	cache->disable_hyperscan = cfg->disable_hyperscan;
```
vstakhov, line 484: This should not be removed - it is an used feature to disable HS from the configuration.

src/libserver/re_cache.c
```diff
@@ -773,7 +771,7 @@ rspamd_re_cache_process_regexp_data (struct rspamd_re_runtime *rt,
 	cache_elt = g_ptr_array_index (rt->cache->re, re_id);
 	re_class = rspamd_regexp_get_class (re);
 
-	if (rt->cache->disable_hyperscan || cache_elt->match_type == RSPAMD_RE_CACHE_PCRE ||
```
vstakhov, line 776: This should not be removed - it is an used feature to disable HS from the configuration.

vstakhov (changes requested):
Please do not remove `cfg->disable_hyperscan` from the configuration and SSSE3 checks for relevant platforms.

cmake/Hyperscan.cmake
a16bitsysop: I was going to do that, but for compiling locally my SBC it appears as arch `unknown` in cmake instead of `aarch64`.  That is inside an alpine docker container.

src/hs_helper.c
a16bitsysop: Added back the removed sections with checks for __aarch64__

src/libserver/cfg_utils.c
```diff
@@ -816,7 +816,7 @@ rspamd_config_post_load (struct rspamd_config *cfg,
 	rspamd_regexp_library_init (cfg);
 	rspamd_multipattern_library_init (cfg->hs_cache_dir);
 
-#ifdef WITH_HYPERSCAN
```
vstakhov, line 819: That will break `cfg->disable_hyperscan` on aarch64.

src/hs_helper.c
vstakhov: Thanks!

src/libserver/cfg_utils.c
a16bitsysop: If I read it correctly, it reads as:
If hyperscan not disabled and cpu does not have ssse3 then disable hyperscan

what would be the correct logic for aarch64/neon?  I think all aarch64 chips have neon.

PR #3589, opened 2020-12-27
[Minor] Fix some compiler warnings

a16bitsysop:


contrib/libev/ev.c
```diff
@@ -2146,7 +2146,8 @@ typedef struct
   #include "ev_wrap.h"
 
   static struct ev_loop default_loop_struct;
-  EV_API_DECL struct ev_loop *ev_default_loop_ptr = 0; /* needs to be initialised to make it a definition despite extern */
```
vstakhov, line 2149: This cannot be done in this way, I'm afraid. Please revert it as it will break so linking badly. I found no good way to fix it without breaking both static and dynamic linking and there are no ways to do it reliably with LTO, as far as I remember.

src/rspamd.c
```diff
@@ -781,9 +781,6 @@ static void
 mark_old_workers (gpointer key, gpointer value, gpointer unused)
 {
 	struct rspamd_worker *w = value;
-	struct rspamd_main *rspamd_main;
-
-	rspamd_main = w->srv;
```
vstakhov, line 784: `rspamd_main` is initialised for logging here like `msg_*_main`. So I'd not recommend to remove this code.

contrib/libev/ev.c
a16bitsysop: Ok, from what I found looking on the web gcc should not give a warning as they list this type of declaration as an example syntax. But also it says extern does not reserve memory for storing a value?

vstakhov: Yes, but this case is a notable exception. It's all about linking after all. Removing of external will make 2 or more default ev_loops which is potentially extremely dangerous (especialy in the case of inotify usage). Removing of static is also not an option as it should be apparently invisible symbol. Hence, we have a contradiction: symbol must be unique and hidden at the same time. There are some gcc extensions that allows that but I'd better leave it as it is defined in the upstream.

a16bitsysop: Ok, I copied from an example where first they create an int and set a value, then declare it extern.

src/rspamd.c
a16bitsysop: Ok, added an attribute unused and added fix from mumhash upstream as well

src/libserver/re_cache.c
```diff
@@ -553,14 +553,12 @@ rspamd_re_cache_check_lua_condition (struct rspamd_task *task,
...
 	lua_State *L = (lua_State *)task->cfg->lua_state;
 	GError *err = NULL;
-	struct rspamd_lua_text *t;
 	gint text_pos;
 
 	if (G_LIKELY (lua_cbref == -1)) {
 		return TRUE;
 	}
 
-	t = lua_new_text (L, in, len, FALSE);
```
vstakhov, line 563: This breaks this function... `t` is created and pushed to the Lua stack and, subsequently, accessed by varargs function `rspamd_lua_universal_pcall` - so it is a compiler bug that it emits warning here.

vstakhov (changes requested):

src/libserver/re_cache.c
a16bitsysop: Ok, thanks.  Added it back with an unused attribute.

vstakhov:
Thank you!

PR #3595, opened 2021-01-04
[Minor] Magic: try avoid false positive exe detection

fatalbanana:


vstakhov:
We should probably check lengths there (for unpack stuff safety).

lualib/lua_magic/heuristics.lua
```diff
@@ -471,4 +471,20 @@ exports.pdf_format_heuristic = function(input, log_obj, pos, part)
   return 'pdf',weight
 end
 
+exports.pe_part_heuristic = function(input, log_obj, pos, part)
+  local content = part:get_raw_content()
```
vstakhov, line 475: Why cannot we use `input` there?

lualib/lua_magic/heuristics.lua
```diff
@@ -471,4 +471,20 @@ exports.pdf_format_heuristic = function(input, log_obj, pos, part)
...
+  if not content then
+    return
+  end
+
+  local pe_ptr_bin = content:sub(60, 64)
+  local pe_ptr = rspamd_util.unpack("<H", pe_ptr_bin)
```
vstakhov, line 481: what if `pe_ptr_bin` is not long enough?

lualib/lua_magic/heuristics.lua
fatalbanana: I failed to realise that `input` is what I wanted.

lualib/lua_magic/heuristics.lua
```diff
@@ -471,4 +471,22 @@ exports.pdf_format_heuristic = function(input, log_obj, pos, part)
...
+  if #pe_ptr_bin ~= 4 then
+    return
+  end
+
+  local pe_ptr = rspamd_util.unpack("<H", pe_ptr_bin)
+  if pe_ptr ~= pos then
+    return
+  end
+
+  return 'exe',15
```
vstakhov, line 489: I'd also suggest to increase 15 to 30 for full confidence in this case.

vstakhov:
Also added https://github.com/rspamd/rspamd/commit/d5bf9ae28d5d1a06c04da4af303fa8edb2b1f098

PR #3599, opened 2021-01-07
[Minor] lua_maps - maybe fix using multiple maps in multimap

c-rosenberg:
When using more than one hash map the multimap plugin is not loading the map files:

  map = [
        "file:///etc/rspamd/maps.d/whitelist_email.txt",
        "file:///etc/rspamd/maps.d/whitelist_email2.txt"
  ];


vstakhov:
This form has been designed for simple elements as a map, like `["key1", "key2"]` so this change might break this form.

Issue #3400, opened 2020-06-10
[Feature] map-support for extended_headers_rcpt / milter_headers


Manawyrm:
## Summary
The extended_headers_rcpt option in the milter_headers plugin allows for configuration of addresses that should receive a full detail info about the spam score. It would be helpful to be able to set this value as a map instead of just a list in the configuration files. 

## Motivation

To figure out why a certain (type of) mail is getting filtered, it would be handy to enable detailed spam scores in the header. We don't want to enable the detailed score for all users. 
The best solution to that would be a map (because the configuration is built and operated by a configuration managment software and written by the server administrators instead of the mail operators). 

## Describe alternatives you've considered

Setting a map as the extended_headers_rcpt doesn't seem to be supported in the plugin. I can't think of another good way to get a detailed spam score only for a few addresses. 
Using the webinterface history is not an option because of the large volume of incoming mail. 

## Target
```
# local.d/milter_headers.conf:
extended_headers_rcpt = "${DBDIR}/extended_headers_rcpt.map";
```

vstakhov:
https://github.com/rspamd/rspamd/commit/0c90e73443254ec4684bd5a142eb627d20e7df18

Manawyrm:
Hello Vsevolod,

thanks a lot for your commit. 
I've tested it and could not reproduce the expected behaviour. 

I looked at the code and added some debug output. 
It looks like the logic is inverted. I've changed the order of operations and inverted the if-statements: 
https://github.com/Manawyrm/rspamd/commit/c562c27e0c8a48b6838d9b08a06cdebf7959d547

The feature now works exactly as intended. 
Should I open a PR or could you pull the commit directly? 

Thanks for your work,
Tobias

vstakhov:
> Should I open a PR or could you pull the commit directly?
Yes please.

Manawyrm:
OK! I've created 
https://github.com/rspamd/rspamd/pull/3607 
:smiley:

Manawyrm:
I guess this is fixed now, by #3607 and 0c90e73443254ec4684bd5a142eb627d20e7df18.

Thanks!

PR #3607, opened 2021-01-14
[Fix] #3400 milter_headers: fix inverted logic for extended_headers_rcpt

Manawyrm:
This commit fixes the inverted logic of skip_wanted/match_extended_headers_rcpt to allow a user to enable extended headers selectively for only a specific set of recipients.

### DIFF
--- a/src/plugins/lua/milter_headers.lua
+++ b/src/plugins/lua/milter_headers.lua
@@ -158,10 +158,6 @@ local function milter_headers(task)
     end
 
 
-    if settings.extended_headers_rcpt and match_extended_headers_rcpt() then
-      return false
-    end
-
     if settings.skip_local and not settings.local_headers[hdr] then
       local ip = task:get_ip()
       if (ip and ip:is_local()) then return true end
@@ -169,6 +165,10 @@ local function milter_headers(task)
 
     if settings.skip_authenticated and not settings.authenticated_headers[hdr] then
       if task:get_user() ~= nil then return true end
+    end
+
+    if settings.extended_headers_rcpt and not match_extended_headers_rcpt() then
+      return true
     end
 
     return false


moisseev:
Now recipients listed in the `extended_headers_rcpt` are ignored for mail originated from authenticated users or our_networks. It is not possible to add extended headers at all for that mails.
Was it intentional?

Manawyrm:
@moisseev 
Thanks for your message.
No, that's most certainly not intentional. 

I will send a new PR tomorrow with an updated and tested routine.

Manawyrm:
@moisseev 
I've reworked the code a bit and pushed it here: 
https://github.com/Manawyrm/rspamd/commit/976bfd2182016190bf107d032e883d8f3c901b15

Would you mind taking a look at the new code? I could now add extended headers for authenticated mails in my testing.
Is it behaving the way you'd expect? 

Thanks!

moisseev:
@Manawyrm What are you trying to achieve? Do you want to add extended headers ONLY for specified recipients?

 I don't understand your original commit in this PR. https://github.com/rspamd/rspamd/pull/3607/commits/c562c27e0c8a48b6838d9b08a06cdebf7959d547

If we have `extended_headers_rcpt` in the configuration and all recipients are matched then we want to add headers regardless of other things. Is that right?
If so, why do we ever need to change these lines?

In any case, IMO we should better introduce another configuration option for that purpose then change plugin logic as it breaks existing configurations.

Manawyrm:
@moisseev 
Previous to my commits, there was the issue #3400, which I opened and wanted to have map support for the extended_headers_rcpt. 
Vsevolod kindly added this support in 0c90e73443254ec4684bd5a142eb627d20e7df18. 

I'm not sure how the feature worked originally, as I was never able to get the extended_headers_rcpt function to work (even without the 0c90e73443254ec4684bd5a142eb627d20e7df18 commit). 
Either I'm very much misreading the documentation for the config file or this feature just never worked correctly (or was broken a while ago and nobody noticed?).

What I'm personally trying to achieve is, I need a map / list of recipients, whose mails are always getting the extended spam headers. I don't have local/authenticated users, as all my Rspamd instance is doing is redirections. I don't want to "bloat" every mail with the extended results header, so I'm just tagging mails for specific recipients in order to allow me to debug the symbols and configuration of the filter (and do something about false positives/negatives). 

From: https://rspamd.com/doc/modules/milter_headers.html
> **skip_local** - Set false to always add headers for local IPs (default true)
> **skip_authenticated** - Set false to always add headers for authenticated users (default true)
> **extended_headers_rcpt** - List of recipients (default empty)
> _extended_headers_rcpt has higher precedence than skip_local and skip_authenticated._

This is not the behaviour I've implemented and I'll need to change that.

What I would expect: 
- if extended_headers_rcpt is not set: default is true, every mail gets the headers
- if extended_headers_rcpt is set: every mail that matches the rcpt-map gets the headers
skip_local and skip_authenticated will be used as described in the docs.

moisseev:
It is not flexible.
Let's say someone has local senders (local IPs or authenticated users) and want default plugin behavior( "Add “extended Rspamd headers” to messages NOT originated from authenticated users or our_networks") plus add the headers to mail for some external mailboxes/domains. With your approach it is not possible.

I propose to introduce another option,  let's say:

> `skip_all` (default false): Do not add extended headers for any messages. 
> `extended_headers_rcpt` has higher precedence than `skip_local`, `skip_authenticated` and `skip_all`.

In this way we are not breaking existing configurations and can configure any desired behavior.

Manawyrm:
@moisseev 
I agree. Your skip_all parameter is a very neat solution for this problem. I've implemented and tested it and it seems to be a perfect solution. It's backward-compatible, works for my usecase and was tidy to implement. :tada:  

Let's do further discussion in: 
https://github.com/rspamd/rspamd/pull/3610
https://github.com/rspamd/rspamd.com/pull/487

vstakhov:
Thank you for your help guys!

PR #3610, opened 2021-01-18
[Minor] milter_headers: new option skip_all

Manawyrm:
This PR introduces the new configuration option "skip_all" for milter_headers.
This config option will skip the addition of extended spam headers for every mail, except those with recipients listed in the extended_headers_rcpt map.

Documentation is updated in https://github.com/rspamd/rspamd.com/pull/487

This solution was proposed by @moisseev in https://github.com/rspamd/rspamd/pull/3607#issuecomment-762319736. 
It works well, is backward-compatible with previous configurations and enables the new usage of only adding extended headers to a specific list of recipients.

### DIFF
--- a/src/plugins/lua/milter_headers.lua
+++ b/src/plugins/lua/milter_headers.lua
@@ -35,6 +35,7 @@ local settings = {
   remove_upstream_spam_flag = true;
   skip_local = true,
   skip_authenticated = true,
+  skip_all = false,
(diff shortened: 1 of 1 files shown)


moisseev:
It is fine, but for next time I'd rather revert c562c27e0c8a48b6838d9b08a06cdebf7959d547 in separate commit. It is easier track changes this way.

PR #3612, opened 2021-01-21
New list from https://publicsuffix.org/list/public_suffix_list.dat

gratuxri:
Maybe it should be automated to take new https://publicsuffix.org/list/public_suffix_list.dat on every release. But now manual change. Please contact me if you have own idea, how to automate it.

### DIFF
--- a/contrib/publicsuffix/effective_tld_names.dat
+++ b/contrib/publicsuffix/effective_tld_names.dat
@@ -721,6 +721,7 @@ ac.ci
 net.ci
 go.ci
 asso.ci
+xn--aroport-bya.ci
 aéroport.ci
 int.ci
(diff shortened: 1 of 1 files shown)


vstakhov:
I'm sorry but you have to preprocess public_suffix via `idn.pl` script. Your patch removes all idn domains which is... bad.

gratuxri:
Sorry for my fault. It should be correct now.

vstakhov:
Thank you! I usually sync this file each release. We have discussed automatic updates with @fatalbanana but have concluded that it is not so trivial to implement it as a map as tld extraction is involved in many things on configuration stage, in tests and in the CLI utilities.

PR #3648, opened 2021-02-16
[Fix] RBL: collapse chained selectors into single label

fatalbanana:
This changes behaviour in a backward-incompatible way but it matches my expectation of how it should have worked and possibly the general expectation.

If old behaviour is really desired, one can unchain one's selectors & use table format of `selector` option.

### DIFF
--- a/src/plugins/lua/rbl.lua
+++ b/src/plugins/lua/rbl.lua
@@ -601,10 +601,8 @@ local function gen_rbl_callback(rule)
       local res = selector(task)
 
       if res then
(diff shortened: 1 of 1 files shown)


vstakhov:
I'm not sure it is the correct change. We flatten selectors closure and then treat it as unflattened. So probably a proper fix is to use unflattened closure or make it configurable. For flattened selector we can use your suggested change and for unflattened the original code seems to be correct.

See also https://lists.rspamd.com/pipermail/users/2021-February/001953.html

PR #3681, opened 2021-03-18
[Minor] antivirus/external_services - fix symbol registration

c-rosenberg:


vstakhov:
I would sugest to assign some default score for these symbols (0 is a good option).

PR #3695, opened 2021-03-31
[Minor] Update dependency info

FelixSchwarz:
still probably incomplete but at least some more info


vstakhov:
Yes, thanks for that! I think there should be a column for a source web site (or github repo) for each dependency.

PR #3734, opened 2021-04-28
further test framework updates

fatalbanana:
Done:
 - Stopped using `Replace Variables` on Rspamd configs, now we use Lupa & prefixed variables instead

In the next PR:
  - Merge tests and use settings
  
 Maybe some day:
  - Think about parallelism again


fatalbanana:
Further changes planned:

* Stop passing variables to export to setup routines, give these RSPAMD_ prefix and automagically export those as environment variables
* Merge tests, use settings

fatalbanana:
Also I should investigate the story about controller errors ....

PR #3769, opened 2021-05-24
[Minor] change headers for zstd and include path

a16bitsysop:
follow on patch from #3764
change headers for zstd depending on bundled or system used


a16bitsysop:
I am trying to build the git master to check this and I get the following error with gcc 10 on alpine using `cmake -DSYSTEM_ZSTD=OFF -DENABLE_LUAJIT=OFF ../rspamd/` or `cmake -DSYSTEM_ZSTD=ON -DENABLE_LUAJIT=OFF ../working/`:

```
contrib/fmt/include/fmt/format-inl.h: In function 'snprintf_float':
/usr/include/fortify/stdio.h:95:27: error: inlining failed in call to 'always_inlin ' 'snprintf': function body can be overwritten at link time
   95 | _FORTIFY_FN(snprintf) int snprintf(char *__s, size_t __n, const char *__f, ...)
      |                           ^
contrib/fmt/include/fmt/format-inl.h:2535:36: note: called from here
 2535 |                      ? snprintf_ptr(begin, capacity, format, precision, value)
      |                                    ^
/usr/include/fortify/stdio.h:95:27: error: inlining failed in call to 'always_inlin ' 'snprintf': function body can be overwritten at link time
   95 | _FORTIFY_FN(snprintf) int snprintf(char *__s, size_t __n, const char *__f, ...)
      |                           ^
contrib/fmt/include/fmt/format-inl.h:2536:36: note: called from here
 2536 |                      : snprintf_ptr(begin, capacity, format, value);
      |                                    ^
contrib/fmt/include/fmt/format-inl.h: In function 'snprintf_float':
/usr/include/fortify/stdio.h:95:27: error: inlining failed in call to 'always_inlin ' 'snprintf': function body can be overwritten at link time
   95 | _FORTIFY_FN(snprintf) int snprintf(char *__s, size_t __n, const char *__f, ...)
      |                           ^
contrib/fmt/include/fmt/format-inl.h:2535:36: note: called from here
 2535 |                      ? snprintf_ptr(begin, capacity, format, precision, value)
      |                                    ^
/usr/include/fortify/stdio.h:95:27: error: inlining failed in call to 'always_inlin ' 'snprintf': function body can be overwritten at link time
   95 | _FORTIFY_FN(snprintf) int snprintf(char *__s, size_t __n, const char *__f, ...)
      |                           ^
contrib/fmt/include/fmt/format-inl.h:2536:36: note: called from here
 2536 |                      : snprintf_ptr(begin, capacity, format, value);
      |                                    ^
lto-wrapper: fatal error: /usr/bin/c++ returned 1 exit status
compilation terminated.
/usr/lib/gcc/aarch64-alpine-linux-musl/10.3.1/../../../../aarch64-alpine-linux-musl/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```

a16bitsysop:
This compiles on ubuntu now, will have a look at alpine build failure later

systemcrash:
Sorry I'm a bit late to the game here: but does this regard zstd in the kernel, or an external (and more recent) dep?

a16bitsysop:
It is:
https://github.com/rspamd/rspamd/tree/master/contrib/zstd

Which is a bundled version of:
https://github.com/facebook/zstd

It is available for most distro's I think.

src/client/rspamdclient.c
```diff
@@ -19,7 +19,7 @@
 #include "libserver/http/http_private.h"
 #include "libserver/protocol_internal.h"
 #include "unix-std.h"
-#include "contrib/zstd/zstd.h"
```
vstakhov, line 22: Can we add #ifdef/#ifndef here to keep `"contrib/zstd/zstd.h"` if we don't enable this option. Otherwise there is a risk of messign the build up.

a16bitsysop: yes it could be like that, I thought the `INCLUDE_DIRECTORIES(BEFORE "${CMAKE_SOURCE_DIR}/contrib/zstd")` would make it the same apart from changing `"contrib/zstd/error_public.h"` to `"zstd_errors.h"` which actually look the same but only `zstd_errors.h` is in the distributed library .

a16bitsysop: How about now? I couldn’t get it to work without defining a new definition.

PR #3772, opened 2021-05-26
[Feature] external_relay plugin

fatalbanana:
Thoughts about having such a plugin / how that should work?

It'd be better that people would do filtering on the MX rather than somewhere beyond it - but if they can't/won't do that a plugin like this might be useful (people seem to ask after something like this periodically & I've been referring them to [this old mailing post](https://groups.google.com/g/rspamd/c/_JusHAeeG74/m/xpsSyXdyDwAJ)).

Currently four different strategies are supported:

`local`: use first non-local IP, or last header
`authenticated`: use first unauthenticated hop
`hostname_map`: look for direct/indirect relays by hostname
`count`: use Received header at some set position (without condition but one might use `rspamd_config:add_condition()`)

### DIFF
--- a/src/plugins/lua/external_relay.lua
+++ b/src/plugins/lua/external_relay.lua
@@ -1,0 +1,250 @@
+--[[
+Copyright (c) 2021, Vsevolod Stakhov <vsevolod@highsecure.ru>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


jonmoesli:
It could be useful to set the number of hops to use (for example when you have one or more internal gateways) for determining the sending host/ip.

src/plugins/lua/external_relay.lua
```diff
@@ -0,0 +1,132 @@
...
+      else
+        -- Keep checking ...
+        next_idx = i + 1
+        break
+      end
+    end
+  end
+
+  if not next_idx then
+    rspamd_logger.errx(task, "couldn't find %s in received headers")
```
andryyy, line 82: Is this missing a variable for %s? :)

andryyy:
Perhaps a check for local/non-local would be helpful, too? Or skipping authenticated rcvd headers?

andryyy:
Will it be able to read ips (v4 and v6) from a radix map? :) And match against subnets? That would be SO cool!

PR #3788, opened 2021-06-08
Add health and readiness checks

mrueg:
These endpoints allow an orchestrator like kubernetes to verify the
status of rspamd (https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

Currently, implementation for both is minimal. 

The health endpoint should allow to verify rspamd internally. It could check for internal configuration and ensure that rspamd itself is healthy and available.
The ready endpoint signals that rspamd is ready to receive and process traffic and thus ensures that configured external components are available. The readiness check could for example test if configured redis servers or at least one rspamd upstreams is available. 
If desired, I can try to figure out how to improve the checks on both endpoints, I was hoping for some feedback on what to check or which functions to call.

### DIFF
--- a/src/controller.c
+++ b/src/controller.c
@@ -46,11 +46,13 @@
 #define PATH_GET_MAP "/getmap"
 #define PATH_GRAPH "/graph"
 #define PATH_PIE_CHART "/pie"
+#define PATH_HEALTHY "/healthy"
 #define PATH_HISTORY "/history"
 #define PATH_HISTORY_RESET "/historyreset"
 #define PATH_LEARN_SPAM "/learnspam"
 #define PATH_LEARN_HAM "/learnham"
 #define PATH_METRICS "/metrics"
+#define PATH_READY "/ready"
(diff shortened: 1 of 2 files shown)


vstakhov:
It looks interesting, thanks! I will check it tomorrow, probably, the health check could be improved as there are many mechanisms in Rspamd that do it (e.g. heartbeats and workers <-> main pipes).

nlewo:
@mrueg thx for these routes!
We would be glad to use it instead of this [hack](https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/blob/8b287056215cac91438a671054e7eb2c932ab21a/tests/multiple.nix#L71).

For our use case, it would actually be even better to notify systemd when rspamd is ready with [sd_notify](https://www.freedesktop.org/software/systemd/man/sd_notify.html). I will try to look where this call could be inserted.

moisseev:
@mrueg In `self-scan` mode the `/ready` endpoint returns `{"error":"no healthy scanner workers are running"}`.

PR #3804, opened 2021-06-23
[Fix] Fix an edge case in BITCOIN_ADDR rule

fatalbanana:
- when using PCRE
 - and different address types are present


rules/bitcoin.lua
```diff
@@ -180,7 +180,7 @@ local btc_bleach_re = [[/\b(?:(?:[a-zA-Z]\w+:)|(?:bc1))?[qpzry9x8gf2tvdw0s3jn54k
 
 config.regexp['BITCOIN_ADDR'] = {
   description = 'Message has a valid bitcoin wallet address',
-  re = string.format('(%s) || (%s)', normal_wallet_re, btc_bleach_re),
+  re = string.format('(%s) + (%s) > 0', normal_wallet_re, btc_bleach_re),
```
citrin, line 183: @fatalbanana could you please add a comment why `+` is needed here instead  of `||` - it's not obvious.

PR #3810, opened 2021-06-30
Switch PGP keyserver to hkps://keys.openpgp.org

piquan:
The SKS keyserver infrastructure has been under attack for some time, and at the moment is unreachable.  The keys.openpgp.org keyserver has the rspamd key, and is resistant to the sort of attack that is hurting the SKS keyservers.  This change switches to that keyserver.

gpg won't load new keys from keys.openpgp.org unless the owner consented to publish the uids (see https://keys.openpgp.org/about/faq#older-gnupg).  This means you need the key imported first with the uid attached.  Here, we include the essential part of the key (the master key, uid, and its self-signature) in the Dockerfile.  We still poll keys.openpgp.org for updates, such as expiration extensions and revocations.

See https://gist.github.com/rjhansen/67ab921ffb4084c865b3618d6955275f for background.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,11 +6,42 @@ RUN	apt-get update && apt-get install -y --no-install-recommends \
 		gnupg \
 	&& rm -rf /var/lib/apt/lists/*
 
+SHELL ["/bin/bash", "-c"]
 RUN	set -x \
 # gpg: key FFA232EDBF21E25E: public key "Rspamd Nightly Builds (Rspamd Nightly Builds) <vsevolod@highsecure.ru>" imported
 	&& key='3FA347D5E599BE4595CA2576FFA232EDBF21E25E' \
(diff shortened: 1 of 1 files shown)


vstakhov:
It doesn't look like my key used for packages signing...

piquan:
Really?  I just double-checked and it seems to be the key I thought you used:
```
pub   rsa4096/FFA232EDBF21E25E 2015-07-28 [SC]
      3FA347D5E599BE4595CA2576FFA232EDBF21E25E
uid                           Rspamd Nightly Builds (Rspamd Nightly Builds) <vsevolod@highsecure.ru>
```
If you're just looking at the armored block, then it's going to look shorter.  That's because I only included the essential parts of the key; I didn't see a need for the encryption subkey, for instance, just the signing key.

Here's the command I used:

vstakhov:
That's my mistake, I have compared my original pubkey and not a stripped version. Stripped version is the same indeed.

PR #3816, opened 2021-07-10
[Feature] Re-add pyzor support

defkev:
Based on https://github.com/rspamd/rspamd/commit/d159074e35b6dd8d3fd2a27f106aa3e12fe92445

Module

* Install `pyzor` e.g. on CentOS/Rocky Linux `yum install pyzor`
* Enable the module in `/etc/rspamd/local.d/external_services.conf`
```
...
pyzor {
  # default pyzor settings
  servers = "127.0.0.1:5953"
}
```
* Restart rspamd  `systemctl restart rspamd`

Socket

* Create `pyzor.socket` in `/usr/lib/systemd/system`
```
[Unit]
Description=Pyzor socket

[Socket]
ListenStream=127.0.0.1:5953
Accept=yes

[Install]
WantedBy=sockets.target
```
* Create `pyzor@.service` in `/usr/lib/systemd/system`
```
[Unit]
Description=Pyzor Socket Service
Requires=pyzor.socket

[Service]
Type=simple
ExecStart=-/usr/bin/pyzor check -
StandardInput=socket
StandardError=journal
TimeoutStopSec=10

User=_rspamd
NoNewPrivileges=true
PrivateDevices=true
PrivateTmp=true
PrivateUsers=true
ProtectControlGroups=true
ProtectHome=true
ProtectKernelModules=true
ProtectKernelTunables=true
ProtectSystem=strict

[Install]
WantedBy=multi-user.target
```
* Reload systemd `systemctl daemon-reload`
* Enable and start pyzor socket `systemctl enable pyzor.socket && systemctl start pyzor.socket`

`pyzor.socket` will call `pyzor@.service` so there is no need to enable/start `pyzor@.service`

Since `pyzor` is executed as `_rspamd` you probably have to create the pyzor home directory `mkdir /var/lib/rspamd/.pyzor`, alternatively create a dedicated pyzor user and edit `pyzor@.service`.

The `PYZOR` symbole weight is calculated dynamically based on the number of times a message has been seen and whitelisted by pyzor (Count - WL-Count of default_score in percent)

### DIFF
--- a/lualib/lua_scanners/pyzor.lua
+++ b/lualib/lua_scanners/pyzor.lua
@@ -41,7 +41,7 @@ local function pyzor_config(opts)
     detection_category = "hash",
     cache_expire = 7200, -- expire redis in one hour
     message = '${SCANNER}: Pyzor bulk message found: "${VIRUS}"',
-    default_score = 1,
+    default_score = 1.5,
     action = false,
   }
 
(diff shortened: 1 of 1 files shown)


vstakhov:
@c-rosenberg How does it look like to you? I'm not very familiar with pyzor I'm afraid.

vstakhov:
I think I will merge it for now, but I'm awaiting docs PR as well :) Thank you!

PR #3817, opened 2021-07-11
[Fix]: Lua ASN - CB func should also return in case of an error

jpbede:
In case of an dns timeout or error in common, the DNS callback function does not return. 
It then still tries to process the results. This interrupts the further processing of the mail.

Example: 
CleanShot 2021-07-11 at 20 35 34@2x
CleanShot 2021-07-11 at 20 35 49@2x

### DIFF
--- a/src/plugins/lua/asn.lua
+++ b/src/plugins/lua/asn.lua
@@ -74,7 +74,7 @@ local function asn_check(task)
         task:insert_result(options['symbol'] .. '_FAIL', 1, string.format('%s:%s', req_name, dns_err))
         return
       end
-      if not (results and results[1]) then
+      if not (results or results[1]) then
         rspamd_logger.infox(task, 'cannot query ip %s on %s: no results',
             req_name, serv)
         return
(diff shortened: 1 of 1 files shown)


vstakhov:
I'm not sure this change will fix the issue (in fact, I'm sure it won't). The following `if` block will handle returning and this make this change quite negligible. Also, `ASN_CHECK_FAIL`  is not registered anywhere in the metric and it will cause issues. Your issue is somewhere in the timeouts/retransmits handling in `options.dns`.

jpbede:
Okay, indeed this is right, but interestingly, the following `if` block doesn't handle it, there is no single line with this message in the logs.

vstakhov:
That's because `and` should be changed to `or` by De-Morgan rules :) Please feel free to include this in this PR if you'd like. I think your change is right but this is not the main culprit of the problem you observe (see https://github.com/mailcow/mailcow-dockerized/commit/b5bf97eec9047511211b48344e6590ae2cd26e3b - this should 'fix' the issue).

jpbede:
Sure, this wont fix the issues with timeout :) Just saw that the handling isn't consistent with the in `rbl.lua` one.

PR #3829, opened 2021-07-28
Test selector that returns list of rspamd_text

fatalbanana:


lualib/lua_selectors/init.lua
```diff
@@ -498,7 +498,7 @@ exports.combine_selectors = function(_, selectors, delimiter)
     end
 
     fun.each(function(...)
-      table.insert(res, table.concat({...}, delimiter))
+      table.insert(res, rspamd_text.fromtable({...}, delimiter))
```
vstakhov, line 501: That will convert all to texts even intended strings. I have another idea here...

PR #3840, opened 2021-08-06
Escape redis+selector

foron:
**Describe the bug**

redis+selector maps are not recognized as valid map types.

**Steps to Reproduce**
```

cat /etc/rspamd/local.d/multimap.conf

test_map {
        type = "from"
        filter = "email"
        map = "redis+selector://id('blacklist:');rcpts:addr.take_n(5).lower";
        description = "Test map";
        symbol = "TEST_MAP"
        score = 5.0;
}
```

**Expected behavior**

The map should be read at startup. Redis keys should be processed accordingly.

**Versions**

Both 2.7 and master

**Additional Information**

I believe multimap.lua should have the "redis+selector" pattern escaped as "redis%+selector". This would make it easier for string.find to match the documented map string.

### DIFF
--- a/src/plugins/lua/multimap.lua
+++ b/src/plugins/lua/multimap.lua
@@ -1099,14 +1099,14 @@ local function add_multimap_rule(key, newrule)
       ret = true
     end
   elseif type(newrule['map']) == 'string' and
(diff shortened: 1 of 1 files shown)


vstakhov:
Good catch, thanks!

Issue #3842, opened 2021-08-06
[BUG] The dkim signing implementation does not insert a whitespace after a comma in the header


sebastianas:
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
I noticed a failed DKIM validation on an email. Upon investigation that email had a missing white space after a comma (used for separating multiple addresses) in the Cc: header. The Email was signed without the space and later the Cc: header was reformed somewhere between a mailing list and the archive. This reformatting added the white space between the comma and the following email address.

### Steps to Reproduce

1. wget https://lore.kernel.org/lkml/87wnp9idso.ffs@tglx/raw
2. $ /usr/sbin/opendkim-testmsg < raw 
   opendkim-testmsg: dkim_eom(): Bad signature
3. edit the raw file and remove all white spaces before Peter and put right after the previous comma.
4. $ /usr/sbin/opendkim-testmsg < raw 
    $ echo $?
   All good.

**Expected behavior**
Not entirely sure. I expect the signing part to add a white space after a comma. But then opendkim seems not to do it. On the other hand the reformatting looks valid and follows the whitespace unfold rules. 

### Versions
Rspamd daemon version 2.7

sebastianas:
The MUA seems to be fixed so the problem seems to be gone as in does not reproduce right now.

PR #3843, opened 2021-08-06
dkim: Enforce a space character after comma in the recipient list

sebastianas:
A Cc list like:

  cc:Person A <a@example.org>,Person B <b@example.org>

will be signed as such. The problem is that there should be a white space
after the comma. If the line gets (later) reformed such as

  Cc: Person A <a@example.org>,
  \TABPerson B <b@example.org>

then line is interpreted as
  cc: Person A <a@example.org>, Person B <b@example.org>

which is different and the verification will failed.

In RFC 6376 section 3.4.2. (The "relaxed" Header Canonicalization
Algorithm) it is described how to unfold/replace white space
characters. It refers to RFC 5322 to how unfold continuation lines. I've
been looking for official behaviour but couldn't find. The space after
the comma is what I observed the missing space being the exception.

Enforce a space after a comma. Ignore this if the comma is found within
a quote block ( started by ' or ") or if the comma is escaped (by \).

Issue: #3842

Signed-off-by: Sebastian Andrzej Siewior <sebastian@breakpoint.cc>

### DIFF
--- a/src/libserver/dkim.c
+++ b/src/libserver/dkim.c
@@ -2210,7 +2210,7 @@ rspamd_dkim_canonize_header_relaxed_str (const gchar *hname,
 {
 	gchar *t;
 	const guchar *h;
-	gboolean got_sp;
+	gboolean got_sp, got_quote_1, got_quote_2, got_escape, got_seperator;
 
 	/* Name part */
 	t = out;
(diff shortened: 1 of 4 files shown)


vstakhov:
Hum, from the first glance, it looks like a DKIM standard violation. @citrin, @moisseev could you please also take a look on that?

sebastianas:
So at a first glace it looks that a space should be enforced by rspamd if the MUA fails to insert one?

vstakhov:
I would suggest to consider fixing MUA in this case... Alternatively, you can alter some headers using Rspamd itself before signing. AFAIR, dkim signing honours modified headers, so that would not be a problem. Unfortunately, the sollution you suggest is actually not compatible with DKIM standard and it can break legit signatures easily.

sebastianas:
Ah okay. I've been told the MUA is a texteditor and the output is fed 1:1 via smtp to the server.
Would you be so kind to give me a pointer how I could use rspamd to modify headers before signing? If that is not too complicated I could try that. Otherwise I might ask to use a proper MUA or live with results…

sebastianas:
Thanks.
The MUA seems to be fixed so the problem seems to be gone as in does not reproduce right now.

PR #3855, opened 2021-08-21
Add SYSTEM_FMT cmake option

arkamar:
This gives packagers option to use system version of fmt rather than bundled one. It is disabled by default.

I would like to point out that it is very unpleasant to maintain rspamd package in Gentoo with this huge amount of bundled libraries, especially for libraries usually present in the system. I believe other maintainers would appreciate possibility to choose between system or bundled version as well. https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,7 @@ OPTION(ENABLE_LIBUNWIND    "Use libunwind to print crash traces [default: OFF]" 
 OPTION(ENABLE_LUA_TRACE    "Trace all Lua C API invocations [default: OFF]" OFF)
 OPTION(ENABLE_LUA_REPL     "Enables Lua repl (requires C++11 compiler) [default: ON]" ON)
 OPTION(SYSTEM_ZSTD         "Use system zstd instead of bundled one [default: OFF]" OFF)
+OPTION(SYSTEM_FMT          "Use system fmt instead of bundled one [defalut: OFF]" OFF)
 
 ############################# INCLUDE SECTION #############################################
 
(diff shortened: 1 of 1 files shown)


vstakhov:
> I would like to point out that it is very unpleasant to maintain rspamd package in Gentoo with this huge amount of bundled libraries

I can only say that it is a mess to maintain Rspamd for that amount of platforms. Especially with broken libicu (thai tokenizer), with some unoptimal libraries which are simple vulnerable to simple DoS, with broken hiredis etc. I wish we can just set that Rspamd wants x.y.z version of some library and let some sane package management system to resolve that stuff. Unfortunately, it is simply impossible.

vstakhov:
Adding an option to use a system library is a good sollution on the other hand, so I'm not against it.

arkamar:
Thanks for the merge! I understand that it is not easy to maintain rspamd for that amount of platforms, but I needed to show my (and probably Gentoo) point of view. Anyway, I am glad that you are not against `-DSYSTEM_*` solution, which is probably good enough for all sides.

bsdice:
Thanks for making this only optional. I prefer bundled libs, because disk space is no longer an argument and stability is often wrecked when system libs are too old (bugs) or too new (untested, more bugs, API breakage). More importantly this may not even be the case on initial setup, everything may work fine. Until a month later a library maintainer decides to push an update and the service starts to crash or misbehave and you don't know why. Such bugs can be very subtle, no chance to diagnose with "systemctl --failed" or "findbrokenpkgs".

Issue #3563, opened 2020-11-29
[BUG] rspamd segfaults on startup on aarch64


CrawX:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
Cannot start rspamd 2.6 on archlinuxarm aarch64 on the [odroid-n2](https://archlinuxarm.org/platforms/armv8/amlogic/odroid-n2) running [4.9.219](https://github.com/hardkernel/linux/tree/odroidn2-4.9.y).

### Steps to Reproduce

1. Install official archlinux-arm rspamd package (same problem when compiling from source)
2. Attempt to start rspamd
3. Starting fails with the following output in the journal:
```
Nov 04 21:04:00 odroid-n2 systemd[1]: Started rapid spam filtering system.
Nov 04 21:04:01 odroid-n2 systemd[1]: rspamd.service: Main process exited, code=dumped, status=11/SEGV
Nov 04 21:04:01 odroid-n2 systemd[1]: rspamd.service: Failed with result 'core-dump'.
Nov 04 21:04:01 odroid-n2 systemd[1]: rspamd.service: Scheduled restart job, restart counter is at 1.
Nov 04 21:04:01 odroid-n2 systemd[1]: Stopped rapid spam filtering system.
Nov 04 21:04:01 odroid-n2 systemd[1]: Started rapid spam filtering system.
Nov 04 21:04:02 odroid-n2 systemd[1]: rspamd.service: Main process exited, code=dumped, status=11/SEGV
Nov 04 21:04:02 odroid-n2 systemd[1]: rspamd.service: Failed with result 'core-dump'.
[...]

PR #3863, opened 2021-08-24
[Fix] Remove aarch64 GC64 workaround

CrawX:
Removes the 64bit LUD workaround since 64bit LUD natively supported in LuaJIT for a while. This fixes the issues described in #3563, please see this issue for details.


vstakhov:
The only problem is that it won't work for the currenly released luajit 2.0 :(

vstakhov:
Or probably not, as it has no GC64 nor aarch64 support.

PR #3865, opened 2021-08-25
set pyzor to check all mail

biji:


c-rosenberg:
I've added this setting to work around false positives with pyzor matching on empty mail bodies.

https://www.mail-archive.com/users@spamassassin.apache.org/msg99644.html

Don't know if this problem still exists.

You could set text_part_min_words = 0; in config if you want.

All kinds of those hashing algorithms (Razor, DCC, IXHash) should be susceptible for FP's somehow on super short or empty mail parts.

biji:
Setting `text_part_min_words = 0` or `nil` still skips checking some email

PR #3874, opened 2021-09-01
[Minor] rspamd_stats: support rescoring symbols

fatalbanana:


utils/rspamd_stats.pl
```diff
@@ -932,6 +935,10 @@ =head1 OPTIONS
 with B<--start> select entries between B<--start> and B<--end>. The omitted date defaults to the current date if you
 supply the time.
 
+=item B<--mult=symbol=integer>
```
vstakhov, line 938: I think mult is not an integer as it can have a fractional part.

utils/rspamd_stats.pl
```diff
@@ -47,6 +48,7 @@
     "ignore=s@"             => \@symbols_ignored,
     "group|g=s@"            => \@symbols_groups,
     "log|l=s"               => \$log_file,
+    "mult=s@"               => \%symbols_mult,
```
vstakhov, line 51: Not sure if s@ works for % hash tables, but I might be wrong...

vstakhov: According to https://perldoc.perl.org/Getopt::Long#Options-with-multiple-values neither `@` nor `%` are needed if passing direct references to arrays/hashes.

vstakhov:
Thank you!

PR #3875, opened 2021-09-02
Add SYSTEM_DOCTEST cmake option

arkamar:
This gives packagers option to use system version of `doctest` rather than bundled one. It is disabled by default. Additionally, there is no need to link `rspamd-test-cxx` with `doctest` as it is is single header library.

Main reason for this change is that currently bundled version of `doctest-2.4.5` can't compile with `glibc-2.34` [1]. However, this issue was already fixed in upstream `doctest-2.4.6` [2] in commit [3]. With this, packagers can just bump doctest to fixed version.

[1] https://github.com/onqtam/doctest/issues/473
[2] https://github.com/onqtam/doctest/blob/master/CHANGELOG.md#246-2021-03-22
[3] https://github.com/onqtam/doctest/commit/099d5414e97244ec44cf46b14cd176b3a3dc52e3


vstakhov:
I should probably update the bundled doctest as well.

sonic42:
Thanks for the in info...

This bug can be closed.

PR #3754, opened 2021-05-14
Implement an endpoint for OpenMetrics / Prometheus

mrueg:
This is a first attempt to have OpenMetrics / Prometheus formatted metrics natively included in rspamd.

If the /stat endpoint is requested with a parameter format=openmetrics,
it will respond in OpenMetrics format instead of JSON.

Currently this is a WIP to get early feedback on the approach, as I'm using horrible string concatenation to get a formatted output.

Linking discussion: https://github.com/rspamd/rspamd/discussions/3484

PR #3883, opened 2021-09-08
[Fix] Namespace and add metadata for OpenMetrics, fix interleaving

pdf:
_Let me preface this PR by saying that I don't write C, so reviewers may wish to handle with care._

Per discussion in #3754 this PR makes the following changes:
- All metrics have `HELP` and `TYPE` defined (with varying usefulness of HELP, since some of the stats are non-obvious to me)
- Metrics are now successfully parsed as the specified `openmetrics-text` content-type (spurious spaces between labels were causing Prometheus to barf)
- Metrics are no longer interleaved for `statfiles_*`
- rspamd-specific metrics are prefixed with `rspamd_`

Some possibly contentious decisions:
1. Metrics that appear to me to be monotonically increasing values have been marked as `counter` type, and consequently have `_total` appended to their metric names. If these values are not in fact monotonically increasing, these metrics will need to be changed to `gauge` type, but if that's the case it would be preferable for those metrics to contain a delta of the value between collections (though that is not critical).
2. The `statfiles_total` metric was renamed to `rspamd_statfiles_totals` (note the trailing `s`), since the `_total` suffix is reserved for `counter` values, and I'm not sure whether that value meets the requirements for a `counter` (it's `0` on my instances, and I have no idea what it denotes).

### DIFF
--- a/src/controller.c
+++ b/src/controller.c
@@ -2803,59 +2803,99 @@ rspamd_controller_metrics_fin_task (void *ud) {
 			ucl_object_fromint (cbdata->learned), "total_learns", 0, false);
 
 	output = rspamd_fstring_sized_new (1024);
-	rspamd_printf_fstring (&output, "build_info{version=\"%s\"} 1\n",
+	rspamd_printf_fstring (&output, "# HELP rspamd_build_info A metric with a constant '1' value labeled by version from which rspamd was built.\n");
+	rspamd_printf_fstring (&output, "# TYPE rspamd_build_info gauge\n");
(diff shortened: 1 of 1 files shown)


vstakhov:
Thank you for the PR! The only comment is that there is a clear memory leak of the strings you have introduced. But I will fix that, no worries.

pdf:
> Thank you for the PR! The only comment is that there is a clear memory leak of the strings you have introduced. But I will fix that, no worries.

Ah right, I should have called `rspamd_fstring_free` on the temporary strings after appending them to the output? Apologies.

PR #3889, opened 2021-09-09
Reflect recent doctest version bump in dependency info table

arkamar:
This just bumps the version number of currently bundled doctest library in dependency info table.


vstakhov:
Oh, indeed, thank you!

PR #3892, opened 2021-09-15
[Test] Avoid importing user configuration

fatalbanana:


vstakhov:
It should do the trick indeed. Thank you!

PR #3934, opened 2021-10-13
Support i386 with optimizations

xtkoba:
Downstream issue: https://bugs.gentoo.org/817866


vstakhov:
Well, it breaks i386 build on FreeBSD: http://beefy17.nyi.freebsd.org/build.html?mastername=main-i386-default&build=pd16266887b66_se900338c09. Since I have no i386 machine to test/reproduce, I will revert this commit.

PR #3959, opened 2021-11-04
[Fix] lua_cfg_transform - actions without score (discard)

c-rosenberg:
Defined actions without score are deactivated in sanity check (#3693)


c-rosenberg:
Should also apply to #3926

PR #3967, opened 2021-11-07
[Fix] Further fix for i386 compilation

a16bitsysop:
Compiles and passes tests on alpine Linux

### DIFF
--- a/src/libcryptobox/cryptobox.c
+++ b/src/libcryptobox/cryptobox.c
@@ -118,7 +118,7 @@ rspamd_cryptobox_test_instr (gint instr)
 	}
 
 	switch (instr) {
-#ifdef HAVE_SSE2
+#if defined HAVE_SSE2 && defined (__x86_64__)
 	case CPUID_SSE2:
 		__asm__ volatile ("psubb %xmm0, %xmm0");
 		break;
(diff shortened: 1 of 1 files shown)


a16bitsysop:
Is `SET(HAVE_AVX2 0)` in the right place?

vstakhov:
I have another idea, as we have AVX2 in C intriniscs which will work and AVX2 in assembly code which will be broken. So we need to distinguish these two things. I'd better just check for architecture in the `chacha` dispatcher.

vstakhov:
Can it now be compiled on x86?

a16bitsysop:
Will check later, thanks.

a16bitsysop:
The new commit still has the rax error:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/533563#L957

https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/27238/diffs

a16bitsysop:
I added this patch:
```
--- a/src/libcryptobox/cryptobox.c
+++ b/src/libcryptobox/cryptobox.c
@@ -146,7 +146,7 @@
                __asm__ volatile ("pcmpeqq %xmm0, %xmm0");
                break;
 #endif
-#ifdef HAVE_SSE42
+#if defined HAVE_SSE42 && defined(__x86_64__)
        case CPUID_SSE42:
                __asm__ volatile ("pushq %rax\n"
                                "xorq %rax, %rax\n"
```

and then it compiles, but segfaults on the run-test, which was the pre patch behaviour.

https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/533814#L1659

a16bitsysop:
With this version it builds and passes tests on alpine:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/533830

PR #3968, opened 2021-11-08
[Minor] Prevent duplicate rows from being inserted into neural traini…

cpragadeesh:
I looked into ordered set vs regular set in redis and I don't think its worth maintaining a rank for each of the training vectors. Please let me know if ordered sets have a benefit that I am not aware of and I'll get the code changed.

### DIFF
--- a/src/plugins/lua/neural.lua
+++ b/src/plugins/lua/neural.lua
@@ -225,7 +225,7 @@ local function ann_push_task_result(rule, task, verdict, score, set)
           local vec = neural_common.result_to_vector(task, set)
 
           local str = rspamd_util.zstd_compress(table.concat(vec, ';'))
-          local target_key = set.ann.redis_key .. '_' .. learn_type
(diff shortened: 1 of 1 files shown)


vstakhov:
We were thinking about storing emphasis in priority (e.g. classification errors). However, upon more thinking I believe the same thing could be achieved by using just a structure inside Redis vectors (e.g. serialising them using message pack and add extra fields if needed) and not using Redis `Z` structures as they require much more maintenance indeed.

vstakhov:
Oh, and it has broken the tests.

cpragadeesh:
> Oh, and it has broken the tests.

Yeah, i'll get it fixed

vstakhov:
Fixing this test will be tricky... It uses the same learning vectors all the time.

vstakhov:
I've almost fixed manual training, but PCA version still fails... Probably I'll just disable it for now. Manual training is not familiar to me, so please fix it if possible.

PR #3969, opened 2021-11-08
[Fix] lua_cfg_transform - silly break breaks actions

c-rosenberg:


c-rosenberg:
Would it makes sense to also check for flags = ["no_threshold"] ?

PR #4003, opened 2021-12-10
Fix Typo at Settings Key

msuslu:
Fixes collect_garbadge table key usage as collect_garbage.


vstakhov:
Good catch, thank you!

PR #4010, opened 2021-12-16
Assign weights to Mailspike Whitelists

dupondje:
Assign a weight to the MAILSPIKE whitelist RBL symbols.
Use the same but negative weight as its blacklist counterparts.

Also fix some indents in the file


moisseev:
IMO 1.0, 1.5, 2.0 scores are too much for IP reputation whitelists.
It would indeed be nice if "good" symbols have negative weights. So I would suggest to set weights about `-0.01` to just indicate their goodness.

vstakhov:
I do agree: some small negative weight would be fine but not a significant one.

dupondje:
Adjusted to some smaller values.

PR #4031, opened 2022-01-03
[Minor] Add no soft reject selector for metadata exporter

dupondje:
I think its quite common to not export soft rejects via the metadata
exporter. Because when its not spam, the sender will most likely try to
redeliver it again anyway.


vstakhov:
Thank you! It would be cool if you can update the documentation for this module as well.

PR #4052, opened 2022-01-30
[Enhancement] Query HELOs, PTRs, and Reply-To's against Spamhaus DBL as well

twesterhever:
On https://www.spamhaus.org/dbl/, Spamhaus explicitly recommends doing so:

> It is effective for filtering email during the SMTP session for all header domain checks - rDNS, HELO, MAIL FROM, From, Reply-To, and Message-ID domains - as well as URLs in messages.

In several productive environments, this was found to pose a small, but noticeable improvement to spam detection rates. Therefore, it makes sense to bring this change upstream IMHO.

### DIFF
--- a/conf/modules.d/rbl.conf
+++ b/conf/modules.d/rbl.conf
@@ -238,7 +238,7 @@ rbl {
       ignore_defaults = true;
       rbl = "dbl.spamhaus.org";
       no_ip = true;
-      checks = ['emails', 'dkim', 'urls'];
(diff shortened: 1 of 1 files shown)


twesterhever:
Updated commit to align with rspamd's coding guidelines.

vstakhov:
It looks like a good idea, and another good thing is that it doesn't increase the number of DNS requests in uncontrolled way (e.g. there can be dozens of DKIM singatures and thousands of URLs in a message).

PR #4056, opened 2022-02-01
[Fix] Lua: task:remove_result didn't work in some cases

fatalbanana:


vstakhov (changes requested):
I would suggest another approach: use a separate `khiter_t` within `is (!isnan (res->score))` branch instead of reusing `k`. It is safe to call for `kh_del` earlier indeed but it is not sound tbh. We operate with `res` and if `results->symbol` would not be a map of pointers, we would have a use-after-free. It is not the case here but I would prefer not to reuse `k` in the inner `if` and leave `kh_del` where it is. Aside that, a good catch, thank you!

src/libmime/scan_result.c
```diff
@@ -1083,4 +1084,4 @@ rspamd_find_metric_result (struct rspamd_task *task,
 	}
 
 	return NULL;
-}
\ No newline at end of file
+}
```
vstakhov, line 1087: 👍

src/libmime/scan_result.c
```diff
@@ -1016,15 +1016,16 @@ struct rspamd_symbol_result* rspamd_task_remove_symbol_result (
 			if (result->sym_groups && res->sym) {
 				struct rspamd_symbol_group *gr;
 				gint i;
+				khiter_t k2;
```
vstakhov, line 1019: `k_groups` ?

vstakhov (approved):

PR #4064, opened 2022-02-05
Rename DBL_ABUSE to DBL_ABUSE_SPAM

slavkoja:
All DBL_ABUSE* symbols have its description in name (DBL_ABUSE_REDIR, DBL_ABUSE_PHISH, DBL_ABUSE_MALWARE, DBL_ABUSE_BOTNET) except this one, which is little confusing. Thus add _SPAM to symbol name, to fit it with other related symbols


vstakhov:
Whilst it looks fine I think it breaks POLA principle as users can use the existing symbol name in their composites. There are many poorly named symbols in Rspamd and it is difficult to rename them in a more consistent way without breaking POLA...

slavkoja:
Yes, this change can break existing composites and/or force actions, but how many users will depend on exactly this symbol, especially when it is abused legit domain symbol, thus prone to false positives. As more probable i consider that users have lowered its score, as it is unusual high for possible legit domain (PR will follow after decide on this).

Yobania (approved):
449A11D5-4CC9-461A-AD8E-CA0CC44F9BA3 de tamaño pequeño

Yobania:
> 449A11D5-4CC9-461A-AD8E-CA0CC44F9BA3 de tamaño pequeñoYY

PR #4066, opened 2022-02-08
[Minor] Tune Lua GC for Lua unit tests

citrin:
Fixes sporadic unit test failures (on Mac OS X).


vstakhov:
I don't see those failures you've mentioned but it seems that this change is totally innocent when the tests are ok.

PR #4070, opened 2022-02-10
[Minor] Disable Received tokens

moisseev:


vstakhov:
TBH, it is not a `Minor` it is a `Fix` for the case when people cannot re-learn messages to a different class as it has a different set of tokens due to this issue with the Received part!

PR #4074, opened 2022-02-10
[Fix] ClickHouse cleanup of old partitions

citrin:
If `/var/lib/rspamd/clickhouse_retention_run` file is missing or has no
number cleanup was not performed because of a following Lua error:
```
lua_periodic_callback_error: call to periodic script (registered at clickhouse.lua:1531) failed: /usr/share/rspamd/plugins/clickhouse.lua:991: attempt to compare number with nil; trace: [1]:{/usr/share/rspamd/plugins/clickhouse.lua:1069 - <unknown> [Lua]};
```

### DIFF
--- a/src/plugins/lua/clickhouse.lua
+++ b/src/plugins/lua/clickhouse.lua
@@ -988,16 +988,16 @@ local function get_last_removal_ago()
     f:close()
   end
 
+  if last_ts == nil or (last_ts + settings.retention.period) <= current_ts then
+    return write_ts_to_file() and 0
+  end
+
   if last_ts > current_ts then
     -- Clock skew detected, overwrite last_ts with current_ts and wait for the next
     -- retention period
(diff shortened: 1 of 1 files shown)


src/plugins/lua/clickhouse.lua
```diff
@@ -988,6 +988,10 @@ local function get_last_removal_ago()
     f:close()
   end
 
+  if last_ts == nil or (last_ts + settings.retention.period) <= current_ts then
```
vstakhov, line 991: I would suggest to use `if type(last_ts) ~= 'number' ... ` as it covers more potential misuse problems.

citrin: `last_ts = tonumber(f:read('*number'))` - `tonumber` should return `number` or `nil`, there should be no other types.

PR #4079, opened 2022-02-15
Lua scanners/icap

c-rosenberg:
- some adjustments for McAfee Web Gateway 9
- check Checkpoint Sandblast compatibility

### DIFF
--- a/lualib/lua_scanners/icap.lua
+++ b/lualib/lua_scanners/icap.lua
@@ -172,6 +172,7 @@ local function icap_check(task, content, digest, rule, maybe_part)
     local http_headers = {}
     local req_headers = {}
     local tcp_options = {}
+    local threat_table = {}
 
     -- Build extended User Agent
     if rule.user_agent == "extended" then
@@ -345,9 +346,21 @@ local function icap_check(task, content, digest, rule, maybe_part)
         return icap_headers
       end
 
-      local function icap_parse_result(headers)
+      local function threat_table_add(icap_threat, maybe_split)
 
-        local threat_string = {}
+        if maybe_split and string.find(icap_threat, ',') then
+          local threats = lua_util.str_split(string.gsub(icap_threat, "%s", ""), ',') or {}
+
+          for _,v in ipairs(threats) do
+            table.insert(threat_table, v)
+          end
+        else
+          table.insert(threat_table, icap_threat)
+        end
+        return true
+      end
+
+      local function icap_parse_result(headers)
 
         --[[
         @ToDo: handle type in response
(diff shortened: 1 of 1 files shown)


lualib/lua_scanners/icap.lua
```diff
@@ -399,37 +403,47 @@ local function icap_check(task, content, digest, rule, maybe_part)
...
+          end
+        end
+
         -- Generic ICAP Headers
         if headers['x-infection-found'] then
           local _,_,icap_type,_,icap_threat =
             headers['x-infection-found']:find("Type=(.-); Resolution=(.-); Threat=(.-);$")
 
-          if not icap_type or icap_type == 2 then
+          if icap_type == 2 then
```
vstakhov, line 423: I don't see the way how it should work. Lua is not a JavaScript with shitty implicit conversions between strings and numbers. In this case, you compare a string `icap_type` with a number `2`. It is always false.

lualib/lua_scanners/icap.lua
```diff
@@ -399,37 +403,47 @@ local function icap_check(task, content, digest, rule, maybe_part)
...
                 '%s: icap error X-Infection-Found: %s', rule.log_prefix, icap_threat)
             common.yield_result(task, rule, icap_threat, 0,
                 'fail', maybe_part)
             return true
-          else
+          elseif icap_threat ~= nil then
             lua_util.debugm(rule.name, task,
                 '%s: icap X-Infection-Found: %s', rule.log_prefix, icap_threat)
             table.insert(threat_string, icap_threat)
+          elseif not icap_threat and headers['x-virus-name']then
```
vstakhov, line 434: Nit: add a space after `]`

lualib/lua_scanners/icap.lua
```diff
@@ -399,37 +403,47 @@ local function icap_check(task, content, digest, rule, maybe_part)
           http: HTTP/1.0 307 Temporary Redirect
         ]] --
 
+        local function string_to_threat_table(header_string)
```
vstakhov, line 406: I would move this function outside of the enclosing function as it does not need to capture any external variables from it's scope.

vstakhov: E.g. it can accept a `header_string` as a source and `threat_string` as a destination (wtf, it should be named as `threat_table` or a `threat_strings_list` for clarity!). In the closure it is totally unclear which side effects this function has. I would refrain from that.

vstakhov (changes requested):
LGTM but I would suggest some small changes.

vstakhov:
@c-rosenberg could you please check my first comment, as it seems to be useful to resolve.

c-rosenberg:
Sorry, for my late response. I ran a bit out of time. Let me update the PR tomorrow evening.

Thank you for your suggestions. That's always welcome.

vstakhov (approved):

vstakhov:
@c-rosenberg thank you! Could you please also merge your branch with the master - it has some conflicts now.

c-rosenberg:
Ups, tried to rebase. Didn't work well :/

c-rosenberg:
Thanks for merging. I try to update the documentation in the next days.

PR #4138, opened 2022-04-06
New SYSTEM_XXHASH option and zstd MODULES  fix

arkamar:
This PR presents two changes:
1. Introduction of `SYSTEM_XXHASH` option for packagers
2. Appended `libzstd` to `MODULES` list because this is a default name, see https://github.com/facebook/zstd/blob/dev/lib/libzstd.pc.in


arkamar:
Thanks!

PR #4140, opened 2022-04-08
Migrate ZSTD_reset{C,D}Stream to a new API

arkamar:
`ZSTD_reset{C,D}Stream` functions are deprecated since 1.4.0 version. This change migrates them to new equivalents as described in documentation.


vstakhov:
LGTM, thank you!

PR #4144, opened 2022-04-09
[Fix] Do not exclude authenticated users from URIBL lookups

twesterhever:
Due to [this line](https://github.com/rspamd/rspamd/blob/master/conf/modules.d/rbl.conf#L16) in the RBL module default configuration file, mails from authenticated users are not subject to any DNSBL lookups. This certainly makes sense for RBLs (i. e. IP-based DNSBLs), but in order to improve detection on outbound spam, URIBLs should be enabled for authenticated users too, so `rspamd` will notice if a compromised account starts emitting mails containing known bad FQDNs.

Therefore, this patch enables URIBL lookups against SURBL, URIBL, Spamhaus DBL and `rspamd.com`'s DBL for authenticated users. Hopefully it helps reducing the amount of spam sent out though cracked accounts a bit. :-)

### DIFF
--- a/conf/modules.d/rbl.conf
+++ b/conf/modules.d/rbl.conf
@@ -194,6 +194,7 @@ rbl {
       rbl = "multi.surbl.org";
       checks = ['emails', 'dkim', 'urls'];
       emails_domainonly = true;
+      exclude_users = false;
 
       returnbits = {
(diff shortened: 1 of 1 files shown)


vstakhov:
Yes, that's a good point!

PR #4146, opened 2022-04-11
[Fix] - Fix URL extraction method that ignores spaces

cpragadeesh:
Rspamd ignores spaces in URL while extracting them -- Consider this example:

``

Get extracted as:

 ```   
{
        "url": "https://rcbcbankard.com/img_email/socialicons.png",
        "tld": "rcbcbankard.com",
        "host": "rcbcbankard.com",
        "flags": [
            "image"
        ]
    }
```
(the difference is that Rspamd ignore the space in `social icons.png`)

Although space is disallowed in URL, Apple mail and web browsers correctly substitute them with `%20` and this causes problems.

I realized by looking at code comments that there are complexities involved with spaces in URL and this bug might introduce regressions in other cases, I have tested this change with multiple scenarios and it seems to work fine with this change. Please let me know if there are cases that I might have missed or if this design was intended :)

### DIFF
--- a/src/libserver/html/html_url.cxx
+++ b/src/libserver/html/html_url.cxx
@@ -380,10 +380,7 @@ html_process_url(rspamd_mempool_t *pool, std::string_view &input)
 	 * including essential ones
 	 */
 	for (auto i = 0; i < sz; i++) {
-		if (G_UNLIKELY (g_ascii_isspace(s[i]))) {
-			continue;
-		}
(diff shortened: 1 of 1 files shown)


vstakhov:
I'm under impression that it is not a correct fix: https://github.com/rspamd/rspamd/commit/9c23b08876a546e6de060a2b7820189f8acf813f

It will also break newlines removal, which is unexpected.

Issue #4164, opened 2022-05-01
[Feature] add scan time and country/asn info to metadata exporter


lucasRolff:
## Summary

I would like to have more metadata available in the metadata exporter, since it's somewhat limited to e.g. what the ElasticSearch or Clickhouse exporter exposes.

Data such as:
- scan_time
- ASN
- Country
- rspamd_server

## Motivation

For people whom uses other tools than ES or ClickHouse, it's useful to have this information available as well - while ASN and country can often be gathered by an external system, it's already available from rspamd itself.

However, for the scan time and rspamd_server, it's useful to know which rspamd server sent a given piece of metadata, as well as the scan time of a given message.

## Describe alternatives you've considered

I looked into whether one could add custom metadata but it doesn't seem to be the case in the current exporter (at least per the docs). What I personally miss are the **scan time**.

PR #4166, opened 2022-05-02
[Minor] Add scan_time and hostname to metadata_exporter

lucasRolff:
Adds scan time and hostname as wished for in #4164 

The code is similar to the `elastic.lua` plugin

### DIFF
--- a/src/plugins/lua/metadata_exporter.lua
+++ b/src/plugins/lua/metadata_exporter.lua
@@ -29,6 +29,7 @@ local rspamd_logger = require "rspamd_logger"
 local ucl = require "ucl"
 local E = {}
 local N = 'metadata_exporter'
+local HOSTNAME = rspamd_util.get_hostname()
 
 local settings = {
   pusher_enabled = {},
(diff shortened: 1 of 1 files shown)


src/plugins/lua/metadata_exporter.lua
```diff
@@ -151,6 +153,18 @@ local function get_general_metadata(task, flatten, no_content)
       return 'unknown'
     end
   end
+
+  local scan_real = task:get_scan_time()
+  scan_real = math.floor(scan_real * 1000)
+  if scan_real < 0 then
+    rspamd_logger.messagex(task,
```
vstakhov, line 160: Does it really happen? Just curious :)

lucasRolff: Ideally it shouldn't, though it's in fact a copy of https://github.com/rspamd/rspamd/commit/6e105da2247e85be60ff31f6046dc8e24c87d807 - when implemented I'm guessing there's a reason why it was done.

I can remove the condition, that's no problem - I'd simply keep it consistent with the elastic plugin

vstakhov: Keeping it consistent is a good thing, no worries. Thank you!

PR #4173, opened 2022-05-20
[Minor] fix lower transformation for utf symbols

tierpod:
Now, lower transformation uses default lower function and doesn't work with utf symbols as expected. Test case:

multimap.conf:

```
bl_test_subj {
    map = "${CONFDIR}/maps/bl-test.map";
    type = "selector";
    selector = "header('Subject').lower";
    score = 0.0;
    symbol = "BL_TEST";
}
```

/etc/rspamd/maps/bl-test.map

```
тест
```

```bash
$ echo test | mail -s "Тест" XXX@XXX:
multimap.lua:437: check value Тест for multimap BL_TEST
multimap.lua:476: found return "false" for multimap BL_TEST

$ echo test | mail -s "тест" XXX@XXX:
multimap.lua:437: check value тест for multimap BL_TEST
multimap.lua:476: found return "" for multimap BL_TEST
```

But there is rspamd_text:lower method that works with utf symbols. With this patch, it works as expected:

```bash
$ echo test | mail -s "Тест" XXX@XXX:
multimap.lua:437: check value тест for multimap BL_TEST
multimap.lua:476: found return "" for multimap BL_TEST

$ echo test | mail -s "тест" XXX@XXX:
multimap.lua:437: check value тест for multimap BL_TEST
multimap.lua:476: found return "" for multimap BL_TEST
```

### DIFF
--- a/lualib/lua_selectors/transforms.lua
+++ b/lualib/lua_selectors/transforms.lua
@@ -40,6 +40,17 @@ local transform_function = {
     end,
     ['description'] = 'Returns the lowercased string',
   },
+  -- Returns the lowercased utf8 string
+  ['lower_utf8'] = {
+    ['types'] = {
+      ['string'] = true,
+    },
+    ['map_type'] = 'string',
+    ['process'] = function(inp, t)
+      return rspamd_util.lower_utf8(inp), t
+    end,
+    ['description'] = 'Returns the lowercased utf8 string',
+  },
   -- Returns the first element
   ['first'] = {
     ['types'] = {
--- a/lualib/lua_selectors/transforms.lua
+++ b/lualib/lua_selectors/transforms.lua
@@ -17,6 +17,7 @@ limitations under the License.
 local fun = require 'fun'
 local lua_util = require "lua_util"
 local rspamd_util = require "rspamd_util"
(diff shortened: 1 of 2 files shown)


lualib/lua_selectors/transforms.lua
```diff
@@ -36,7 +37,7 @@ local transform_function = {
     },
     ['map_type'] = 'string',
     ['process'] = function(inp, _)
-      return inp:lower(),'string'
```
vstakhov, line 39: I would suggest to add an additional selector instead of introducing a slashing change. For the most of the cases `lower` should be treated as `ascii_lower` for performance and compatibility reasons. If we want a `utf8` version of lower, it is good to have a `lower_utf8` selector. Instead of transforming all to text, I would suggest to use the following code:

```lua
function(inp, ty)
  return rspamd_util.lower_utf8(inp),ty
end
```

It will deal with both native strings and texts without implicit copy and type transformation.

vstakhov (changes requested):

lualib/lua_selectors/transforms.lua
tierpod: I see, that makes sense. I pushed another version with lower_utf8 selector

PR #4203, opened 2022-06-27
Add dumpex command to mime plugin

nickliebmann:
Add dumpex command, which extends the dump command by outputting attachments into separate '.raw' files. 
Outputs all generated filenames to stdout


vstakhov:
I would suggest to add a `split` flag to the `rspamadm mime dump` command. Adding a separate command for that purpose is not very fortunate as it implies copying of the `dump` subcommand logic in two places. With a flag approach it could all be done in `dump` subcommand I suppose.

nickliebmann:
Superseded by https://github.com/rspamd/rspamd/pull/4222

PR #4204, opened 2022-06-27
msgpack 'content' field is written as binary always

nickliebmann:
When writing out msgpack file, the content field is better written consistently as binary for consistent handling at decode time

### DIFF
--- a/lualib/lua_mime.lua
+++ b/lualib/lua_mime.lua
@@ -615,7 +615,7 @@ exports.message_to_ucl = function(task, stringify_content)
         type = string.format('%s/%s', part:get_type()),
         detected_type = string.format('%s/%s', part:get_detected_type()),
         filename = part:get_filename(),
-        content = maybe_stringify_f(part:get_content()),
+        content = part:get_content(),
         headers =  part:get_headers(true) or E,
         boundary = part:get_enclosing_boundary(),
       }
--- a/lualib/lua_mime.lua
+++ b/lualib/lua_mime.lua
@@ -579,8 +579,6 @@ end
 exports.message_to_ucl = function(task, stringify_content)
   local E = {}
 
-  local maybe_stringify_f = stringify_content and
-    tostring or function(t) return t  end
   local result = {
     size = task:get_size(),
     digest = task:get_digest(),


vstakhov:
Hum, I've not got it. The function `message_to_ucl` accepts an argument that allows to stringify content, e.g. to convert it from `rspamd_text` to a plain Lua string. I have checked the invocations of this function in the Rspamd code, and I have found merely one call where this parameter is `true`: https://github.com/rspamd/rspamd/blob/master/test/lua/unit/lua_mime.message_to_ucl.lua#L214

In all other places this parameter is `false` meaning that the content will not be stringified. Am I missing anything here?

nickliebmann:
As part of testing this, I was able to see the msgpack generated with a string field (when using the old code - quoted-printable attachment) Vs binary with the new code

vstakhov:
Could you please elaborate more on that: how have you called this method to have a quoted-printable attachment? From what I see, `message_to_ucl(task, false)` should be equal to your desired change.

nickliebmann:
[Zipfile_attachment_test.zip](https://github.com/rspamd/rspamd/files/9001112/Zipfile_attachment_test.zip)

Attached is a sample email which encodes to msgpack differently depending on whether the change is present.
 
From (poor) memory what I saw was the same data in the msgpack, apart from the byte identifier which indicates the type of the data in the specific field.
 
Contrary to what I said earlier, the attachment in this case is not encoded quoted-printable; My comment is related to the fact that we first saw this issue when we attached zip files to an email. Previously we have thought attachments were working fine, but they had all been quoted-printable encoded, which may have masked the issue.

vstakhov:
No, in both cases it is treated as normal string instead of the binary string. I know how to fix it properly I think - there is a special attribute called `UCL_OBJECT_BINARY`, so the Lua code should use it when there are raw 8 bit characters in the strings/texts. The tricky stuff is to detect that for any text and distingusish the case of the valid utf8.

nickliebmann:
This has been addressed in commit https://github.com/rspamd/rspamd/commit/d87f4e9f5759b4cb85072b422d353cea1437df54

PR #4206, opened 2022-06-28
Update rspamd.spec, delete old el6 files

ghtm2:
After not receiving any feedback on [my question](https://github.com/rspamd/rspamd.com/issues/565) I've decided to go ahead and update the spec file myself.
I've removed a bit of cruft in the process, which is hopefully OK.

These changes are build and run tested on el7, el8 and el9, everything works fine as far as I can tell.

A bit of info on building in case this gets accepted and I or someone else wants to update the [downloads section](https://rspamd.com/downloads.html):
All require the EPEL repo as the docs already state, additionally el7 based distros build require devtoolset-8 from the [SCL repo](https://wiki.centos.org/AdditionalResources/Repositories/SCL) and hyperscale-devel from the [Open Fusion repo](http://www.openfusion.net/linux/openfusion_rpm_repository), el8 based distros the PowerTools repo and el9 based distros the CRB repo.


vstakhov:
That looks reasonable, thank you! For now, I will merge this PR and then test it with my build scripts.

ghtm2:
Not a problem, thank you for rspamd!

PR #4208, opened 2022-06-29
[Fix] Upstreams: Don't ignore revive_time config option

citrin:


vstakhov:
Oh crap, thank you very much for noticing that! We definitely need some unit testing for the upstreams module...

PR #4211, opened 2022-07-03
Update rspamd.spec to fix Fedora 36 build

ghtm2:
The build on Fedora 36 is failing due to [a bug in ld.gold](https://bugzilla.redhat.com/show_bug.cgi?id=2043178) that prevents successful linking.
The recommended work around is to use ld.bfd instead.

I've also tried [ld.mold](https://github.com/rui314/mold) which works as well, but I was unsure of the stance on adding another build time dependency.

With this I was able to build and run test rspamd on el7, el8, el9 as well as Fedora 34, 35 and 36.


vstakhov:
I don't have this particular OS to test, but I'm pretty much sure that your investigations are right. Thank you!

PR #4214, opened 2022-07-13
[Minor] rspamc: fix crash on non-string element in `messages`

fatalbanana:
Lists and dictionaries can be added to the protocol response inside `messages`, e.g.

```
          task:append_message({foo = 'bar'}, 'yolo')
          task:append_message({'bar','foo'}, 'swag')
```

Currently this is causing a crash in `rspamc`.

```
(gdb) p ucl_object_tostring(cmesg)
$3 = 0x0
```

This doesn't look quite as I might have imagined it but it's better than crashing:
```
Message - yolo: object
Message - swag: array
```

### DIFF
--- a/src/client/rspamc.cxx
+++ b/src/client/rspamc.cxx
@@ -1030,8 +1030,16 @@ rspamc_symbols_output(FILE *out, ucl_object_t *obj)
 		const ucl_object_t *cmesg;
 
 		while ((cmesg = ucl_object_iterate (elt, &mit, true)) != nullptr) {
-			fmt::print(out, "Message - {}: {}\n",
-					ucl_object_key(cmesg), ucl_object_tostring(cmesg));
+			if (ucl_object_type(cmesg) == UCL_STRING) {
+				fmt::print(out, "Message - {}: {}\n",
+						ucl_object_key(cmesg), ucl_object_tostring(cmesg));
+			} else {
+				unsigned char *rendered_message;
+				rendered_message = ucl_object_emit(cmesg, UCL_EMIT_JSON_COMPACT);
+				fmt::print(out, "Message - {}: {:.60}\n",
+						ucl_object_key(cmesg), rendered_message);
+				free(rendered_message);
+			}
 		}
 	}
 


fatalbanana:
the patch as currently proposed mishandles strings. :\ I'll revisit.

vstakhov:
I don't like the usage of the `ucl_object_tostring_forced` here, as it is way too bogus. For a human readable implementation I would suggest just to emit ucl object into, well, ucl. `ucl_object_tostring_forced` was designed to do something similar but in a more intrusive and state modifying way: `deconst->trash_stack[UCL_TRASH_VALUE] = ucl_object_emit_single_json (obj);`. Please note that it also emits json instead of a more logical choice of the `UCL_EMIT_CONFIG` (at least for the human readable output).

fatalbanana:
```
Message - yolo: {"foo":"bar"}
Message - swag: ["bar","foo"]
Message - barf: lol
```

looks nicer

src/client/rspamc.cxx
```diff
@@ -1030,8 +1030,14 @@ rspamc_symbols_output(FILE *out, ucl_object_t *obj)
...
 		while ((cmesg = ucl_object_iterate (elt, &mit, true)) != nullptr) {
-			fmt::print(out, "Message - {}: {}\n",
-					ucl_object_key(cmesg), ucl_object_tostring(cmesg));
+			if (cmesg->type == UCL_STRING) {
+				fmt::print(out, "Message - {}: {}\n",
+						ucl_object_key(cmesg), ucl_object_tostring(cmesg));
+			} else {
+				fmt::print(out, "Message - {}: {}\n",
+						ucl_object_key(cmesg),
+						ucl_object_emit(cmesg, UCL_EMIT_JSON_COMPACT));
```
vstakhov, line 1039: We have discussed that with @citrin recently, and I would also suggest to make it in a different way.

1) When you do `ucl_object_emit` you allocate memory that is never freed and that's a memory leak
2) `cmesg` can be lengthy and not fitting into the terminal, so I suggest to call emit first, then check the length of the string and truncate it if the length is more than ~60 characters. We are talking about human readable representation after all.

src/client/rspamc.cxx
```diff
@@ -1030,8 +1030,14 @@ rspamc_symbols_output(FILE *out, ucl_object_t *obj)
 		const ucl_object_t *cmesg;
 
 		while ((cmesg = ucl_object_iterate (elt, &mit, true)) != nullptr) {
-			fmt::print(out, "Message - {}: {}\n",
-					ucl_object_key(cmesg), ucl_object_tostring(cmesg));
+			if (cmesg->type == UCL_STRING) {
```
vstakhov, line 1033: 👍

vstakhov (changes requested):

src/client/rspamc.cxx
vstakhov, line 1033:
```suggestion
			if (ucl_object_type(cmesg) == UCL_STRING) {
```

fatalbanana:
Using `UCL_EMIT_CONFIG` as suggested:

```
Message - yolo: foo = "bar";

Message - swag: [
    "bar",
    "foo",
]
```

looks worse to me.

vstakhov:
Ok, let's go with the JSON approach, but please read my other comments.

vstakhov (approved):

PR #4222, opened 2022-07-22
[Minor] Mimedump improvements

JasonStephenson:
## Summary
Add the ability to write to a specified directory instead of stdout
Add the ability to split contents out into files, when output directory is specified (-o) and formatting is performed (--json/--messagepack etc)

This is a rework of https://github.com/rspamd/rspamd/pull/4203

### [Existing] Single file, no formatting, to stdout
```
rspamadm mime dump ./samples/very_simple.eml
```

### [Existing] Multiple files, no formatting, to stdout
```
rspamadm mime dump ./samples/very_simple.eml ./samples/Zipfile_test.eml
```

### [Existing] Multiple files, formatted, to stdout
```
rspamadm mime dump --json ./samples/very_simple.eml ./samples/Zipfile_test.eml
rspamadm mime dump --messagepack ./samples/very_simple.eml ./samples/Zipfile_test.eml
```

### [New] Multiple files, formatted, to output directory (filenames written to stdout instead)
```
rspamadm mime dump -o /tmp --json ./samples/very_simple.eml ./samples/Zipfile_test.eml
rspamadm mime dump -o /tmp/ --messagepack ./samples/very_simple.eml ./samples/Zipfile_test.eml
```

### [New] Multiple files, formatted, to output directory, content split (filenames written to stdout instead)
```
rspamadm mime dump -o /tmp --json --split ./samples/very_simple.eml ./samples/Zipfile_test.eml
rspamadm mime dump -o /tmp --messagepack --split ./samples/very_simple.eml ./samples/Zipfile_test.eml
```

## Example --split usage
```
> rspamadm mime dump -o /tmp --json --split ./samples/Zipfile_test.eml
got empty text part
/tmp/Zipfile_test-part3.raw
/tmp/Zipfile_test-part4.raw
/tmp/Zipfile_test-part5.raw
/tmp/Zipfile_test.json

> cat /tmp/Zipfile_test.json
{
    "size": 927862,
    "digest": "0c0853fbae3edeccf358d29d1529b90e",
    "parts": [
        {
            "multipart_boundary": "_004_VI1P194MB0495A8C9FFAD6200D0582E4FDAA49VI1P194MB0495EURP_",
            "size": 0,
            "type": "multipart/mixed",
            "headers": []
        },
        {
            "size": 0,
            "multipart_boundary": "_000_VI1P194MB0495A8C9FFAD6200D0582E4FDAA49VI1P194MB0495EURP_",
            "type": "multipart/alternative",
            "boundary": "_004_VI1P194MB0495A8C9FFAD6200D0582E4FDAA49VI1P194MB0495EURP_",
            "headers": [
                ...
            ]
        },
        {
            "content": null,
            "size": 4,
            "boundary": "_000_VI1P194MB0495A8C9FFAD6200D0582E4FDAA49VI1P194MB0495EURP_",
            "type": "text/plain",
            "content_path": "/tmp/Zipfile_test-part3.raw",
            "detected_type": "text/plain",
            "headers": [
                ...
            ]
        },
        {
            "content": null,
            "size": 2,
            "boundary": "_000_VI1P194MB0495A8C9FFAD6200D0582E4FDAA49VI1P194MB0495EURP_",
            "type": "text/html",
            "content_path": "/tmp/Zipfile_test-part4.raw",
            "detected_type": "text/html",
            "headers": [
                ...
            ]
        },
        {
            "content": null,
            "size": 676901,
            "boundary": "_004_VI1P194MB0495A8C9FFAD6200D0582E4FDAA49VI1P194MB0495EURP_",
            "type": "application/zip",
            "content_path": "/tmp/Zipfile_test-part5.raw",
            "detected_type": "application/zip",
            "filename": "Embedded_plain.zip",
            "headers": [
                ...
            ]
        }
        ...
        
> ls -ltr /tmp | grep Zip
-rw-rw-r--. 1 jst  jst   10534 Jul 22 10:50 Zipfile_test.json
-rw-rw-r--. 1 jst  jst       4 Jul 22 10:50 Zipfile_test-part3.raw
-rw-rw-r--. 1 jst  jst       2 Jul 22 10:50 Zipfile_test-part4.raw
-rw-rw-r--. 1 jst  jst  676901 Jul 22 10:50 Zipfile_test-part5.raw
```

### DIFF
--- a/lualib/rspamadm/mime.lua
+++ b/lualib/rspamadm/mime.lua
@@ -919,7 +919,6 @@ local function write_dump_content(dump_content, fname, extension, outdir)
   else
       io.write(dump_content)
   end
-luacheck
   return wrote_filepath
 end
 
--- a/lualib/rspamadm/mime.lua
+++ b/lualib/rspamadm/mime.lua
@@ -949,14 +949,12 @@ local function get_dump_content(task, opts, fname)
     local extension = output_fmt(opts)
     return ucl.to_format(ucl_object, extension), extension
   end
-  
   local content = task:get_content()
   if type(content) == "string" then
     return content, "mime"
   else
     return tostring(content), "mime"
   end
-  
 end
 
 local function dump_handler(opts)
--- a/lualib/rspamadm/mime.lua
+++ b/lualib/rspamadm/mime.lua
@@ -209,10 +209,11 @@ dump:mutex(
     parser:flag "-U --ucl"
           :description "UCL output",
     parser:flag "-M --messagepack"
-          :description "MessagePack output",
-    parser:flag "-E --extract"
-          :description "Always extract content to disk"
+          :description "MessagePack output"
 )
+dump:flag "-s --split"
+      :description "Split the output file contents such that no content is embedded"
+
 dump:option "-o --outdir"
       :description "Output directory"
       :argname("<directory>")
@@ -878,26 +879,28 @@ local function sign_handler(opts)
   end
 end
 
---Strips directories and .extensions (if present) from a filepath
+-- Strips directories and .extensions (if present) from a filepath
+  -- very_simple
+  -- /home/very_simple.eml
+  -- very_simple.eml
+  -- very_simple.example.eml
+  -- /home/very_simple
+  -- home/very_simple
+  -- ./home/very_simple
+  -- ../home/very_simple.eml
+--All the above end up as very_simple
 local function filename_only(filepath)
-  filename = filepath:match(".*%/([^%.]*)")
+  filename = filepath:match(".*%/([^%.]+)")
   if not filename then
-    filename = filepath:match("(.*)%.")
+    filename = filepath:match("([^%.]+)")
   end
   return filename
 end
 
-local function get_dump_content(task, opts)
-  if opts.ucl or opts.json or opts.messagepack then
-    local ucl_object = lua_mime.message_to_ucl(task)
-    local extension = output_fmt(opts)
-    return ucl.to_format(ucl_object, out_extension), extension
-  end
-  return tostring(task:get_content()), "mime"
-end
-
 --Write the dump content to file or standard out
-local function write_dump_content(data, fname, extension, outdir)
+local function write_dump_content(dump_content, fname, extension, outdir)
+  
+  wrote_filepath = nil
   if outdir then
     if outdir:sub(-1) ~= "/" then
       outdir = outdir .. "/"
@@ -907,15 +910,43 @@ local function write_dump_content(data, fname, extension, outdir)
     local outfile = io.open(outpath, "w")
 
     if outfile then
-      outfile:write(data)
+      outfile:write(dump_content)
       outfile:close()
       io.write(outpath.."\n")
+      wrote_filepath = outpath
     else
       io.stderr:write(string.format("Unable to open: %s\n", outpath))
     end
   else
-      io.write(data)
+      io.write(dump_content)
   end
+
+  return wrote_filepath
+end
+
+-- Get the formatted ucl (split or unsplit) or the raw task content
+local function get_dump_content(task, opts, fname)
+  if opts.ucl or opts.json or opts.messagepack then
+    local ucl_object = lua_mime.message_to_ucl(task)
+    
+    --Split out the content field into separate raws and update the ucl
+    if opts.split then
+      for i, part in ipairs(ucl_object.parts) do
+        if part.content then
+          local part_filename = string.format("%s-part%d", filename_only(fname), i)
+          local part_path = write_dump_content(tostring(part.content), part_filename, "raw", opts.outdir)
+          if part_path then
+            part.content = ucl.null
+            part.content_path = part_path
+          end
+        end
+      end
+    end
+
+    local extension = output_fmt(opts)
+    return ucl.to_format(ucl_object, extension), extension
+  end
+  return tostring(task:get_content()), "mime"
 end
 
 local function dump_handler(opts)
@@ -925,7 +956,7 @@ local function dump_handler(opts)
   for _,fname in ipairs(opts.file) do
     local task = load_task(opts, fname)
     
-    local data, extension = get_dump_content(task, opts)
+    local data, extension = get_dump_content(task, opts, fname)
     write_dump_content(data, fname, extension, opts.outdir)
 
     task:destroy() -- No automatic dtor
--- a/lualib/rspamadm/mime.lua
+++ b/lualib/rspamadm/mime.lua
@@ -890,7 +890,7 @@ end
   -- ../home/very_simple.eml
 --All the above end up as very_simple
 local function filename_only(filepath)
-  filename = filepath:match(".*%/([^%.]+)")
+  local filename = filepath:match(".*%/([^%.]+)")
   if not filename then
     filename = filepath:match("([^%.]+)")
   end
@@ -899,8 +899,7 @@ end
 
 --Write the dump content to file or standard out
 local function write_dump_content(dump_content, fname, extension, outdir)
-  
-  wrote_filepath = nil
+  local wrote_filepath = nil
   if outdir then
     if outdir:sub(-1) ~= "/" then
       outdir = outdir .. "/"
@@ -920,7 +919,7 @@ local function write_dump_content(dump_content, fname, extension, outdir)
   else
       io.write(dump_content)
   end
-
+luacheck
   return wrote_filepath
 end
 
@@ -928,8 +927,7 @@ end
 local function get_dump_content(task, opts, fname)
   if opts.ucl or opts.json or opts.messagepack then
     local ucl_object = lua_mime.message_to_ucl(task)
-    
-    --Split out the content field into separate raws and update the ucl
+    -- Split out the content field into separate raws and update the ucl
     if opts.split then
       for i, part in ipairs(ucl_object.parts) do
         if part.content then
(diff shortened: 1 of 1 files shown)


lualib/rspamadm/mime.lua
```diff
@@ -873,19 +879,85 @@ local function sign_handler(opts)
...
+  -- very_simple
+  -- /home/very_simple.eml
+  -- very_simple.eml
+  -- very_simple.example.eml
+  -- /home/very_simple
+  -- home/very_simple
+  -- ./home/very_simple
+  -- ../home/very_simple.eml
+--All the above end up as very_simple
+local function filename_only(filepath)
```
vstakhov, line 892: What if there are more than one extension, e.g. `tar.gz`? Currently, there is a code that does it in the following way: https://github.com/rspamd/rspamd/blob/master/src/plugins/lua/mime_types.lua#L210
But it will likely be broken with a full path: `dir.with.dot/ololo` will not be processed correctly. However, I'm not sure that your function will process it correctly either.

lualib/rspamadm/mime.lua
```diff
@@ -873,19 +879,85 @@ local function sign_handler(opts)
...
+local function get_dump_content(task, opts, fname)
+  if opts.ucl or opts.json or opts.messagepack then
+    local ucl_object = lua_mime.message_to_ucl(task)
+    
+    --Split out the content field into separate raws and update the ucl
+    if opts.split then
+      for i, part in ipairs(ucl_object.parts) do
+        if part.content then
+          local part_filename = string.format("%s-part%d", filename_only(fname), i)
+          local part_path = write_dump_content(tostring(part.content), part_filename, "raw", opts.outdir)
```
vstakhov, line 937: I don't understand why do we call `tostring` here. It will allocate and intern a potentially large string in Lua. It would be better to adopt `write_dump_content` function for the `rspamd_text` case.

lualib/rspamadm/mime.lua
```diff
@@ -873,19 +879,85 @@ local function sign_handler(opts)
...
+            part.content_path = part_path
+          end
+        end
+      end
+    end
+
+    local extension = output_fmt(opts)
+    return ucl.to_format(ucl_object, extension), extension
+  end
+  return tostring(task:get_content()), "mime"
```
vstakhov, line 949: Same here - it is not clear to me why `tostring` is used.

vstakhov (changes requested):
It looks very good, thank you! I would suggest some small changes to avoid unnecessary copy where it is relatively easy not to do that.

vstakhov:
It also seems that the CI pipeline fails due to `luacheck` failures: https://ci.rspamd.com/rspamd/rspamd/3115/1/7

lualib/rspamadm/mime.lua
JasonStephenson: Multiple file extensions are handled as per [line 886](https://github.com/JasonStephenson/rspamd/blob/6460f4088b8f8fcaa167ee0cef3505d7ded9d6ec/lualib/rspamadm/mime.lua#L886)

My function correctly handles directory names with dots
`print(filename_only("/home/a.directory/very_simple.eml")) --> very_simple`

lualib/rspamadm/mime.lua
JasonStephenson: Further down the stack I was getting errors from file write because the type wasn't a string it was userdata. A quick google explained that tostring could be used if the type implemented it (via something like __string - i forget). I tried it, it worked so I stopped there.

If there's a better way please point me at some examples. I am not very good with lua.

lualib/rspamadm/mime.lua
vstakhov: Could you please add some asserts like that directly in the code, so we can ensure that it is not broken by further changes. It will slightly slow down the loading of the file but it is negligible comparing to the benefits from the correctness. You can check the following example to get an idea: https://github.com/rspamd/rspamd/blob/master/lualib/lua_aws.lua#L281

lualib/rspamadm/mime.lua
vstakhov: Yes, so I would suggest to add a simple `if` to check that `type(data) == 'string'` or `type(data) == 'userdata'`. For `rspamd_text` you could use [save_in_file](https://rspamd.com/doc/lua/rspamd_text.html#m0d82f) method. I don't understand also, why do you add `\n` at the end.

lualib/rspamadm/mime.lua
JasonStephenson: done

lualib/rspamadm/mime.lua
JasonStephenson: added if type check before tostring calls

lualib/rspamadm/mime.lua
JasonStephenson: as above

lualib/rspamadm/mime.lua
```diff
@@ -873,19 +879,92 @@ local function sign_handler(opts)
...
+--Write the dump content to file or standard out
+local function write_dump_content(dump_content, fname, extension, outdir)
+  local wrote_filepath = nil
+  if outdir then
+    if outdir:sub(-1) ~= "/" then
+      outdir = outdir .. "/"
+    end
+
+    local outpath = string.format("%s%s.%s", outdir, filename_only(fname), extension)
+    local outfile = io.open(outpath, "w")
```
vstakhov, line 910: I'd suggest to distingiush `userdata` and `string` here.

lualib/rspamadm/mime.lua
```diff
@@ -873,19 +879,92 @@ local function sign_handler(opts)
...
+
+    if outfile then
+      outfile:write(dump_content)
+      outfile:close()
+      io.write(outpath.."\n")
```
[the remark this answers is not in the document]
JasonStephenson: There is no other instance of print in this file, but there are instances of io.write("\n"). io.write seems the more consistant approach?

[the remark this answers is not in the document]
vstakhov: Well, Lua has no real guide about when to use each of the functions. It looks like `print` is more like debug print, while `io.write` is more like `real` writing. In this particular case, the usage of `print` seems to be more reasonable than `io.write`. But both options are fine after all.

lualib/rspamadm/mime.lua
```diff
@@ -873,19 +879,92 @@ local function sign_handler(opts)
...
+      io.stderr:write(string.format("Unable to open: %s\n", outpath))
+    end
+  else
+      io.write(dump_content)
+  end
+  return wrote_filepath
+end
+
+-- Get the formatted ucl (split or unsplit) or the raw task content
+local function get_dump_content(task, opts, fname)
```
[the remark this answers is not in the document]
JasonStephenson: Understood. Presumably if `dump_content` within the `write_dump_content` is a string I would NOT load it into rspamd_text, and just write it to file to avoid the copy. If it is userdata I would load it into rspamd_text and use the save_in_file method? 

How would I instantiate rspamd_text from userdata. I thought `fromtable` but that appears to be incorrect. The docs appear unclear on this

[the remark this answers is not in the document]
vstakhov: Methods to get content return userdata that is `rspamd_text`. In theory, you can ensure that `userdata` is a text by comparing it's `.cookie` field with `rspamd_text.cookie`. But if you operate with stuff returned from `task` methods, they can realistically return merely `rspamd_text` or a `string`. `rspamd_text` was designed for a sole purpose to provide opaque structure to prevent large strings interning in Lua which could be quite expensive and requires many copying when passing that between Lua and C.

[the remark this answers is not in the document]
JasonStephenson: This approach of having `get_dump_content` not always returning a string (using tostring) complicates the writing a fair bit. The conditional type checks need to be duplicated in the "write to file" code and the "write to stdout" code.

I tried to normalise it by writing to a fd. Defaulting to 1, but set to the return value of `rspamd_util.create_file` if -o is specified. However, it appears that the rspamd code `save_in_file` and `create_file` do not support overwritting existing files in the same manner `io.open(.., "w")` does?

```
if type(dump_content) == "userdata" then
      if dump_content:save_in_file(outpath, "w") then
        wrote_filepath = outpath
      else
        io.stderr:write(string.format("Unable to save_in_file: %s\n", outpath))
      end
```

Getting `Unable to save_in_file: /tmp/Zipfile_test-part3.raw`  etc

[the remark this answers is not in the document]
vstakhov: AFAIR, you can use `-` as a file name to write to stdout. But I agree that the best approach would be to accept io handle from Lua. Last time I have checked that it was quite difficult to implement though...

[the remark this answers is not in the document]
JasonStephenson: I believe my latest iteration is a happy balance between performance and readability. In summary:

- Instead of `tostring` on userdata in `get_dump_content` the content is returned directly
- Within  `write_dump_content` string `dump_content` will be loaded into rspamd_text type

While there is an additional copy taken here during the `rspamd_text.fromstring` call I believe this is acceptable because:
- it is still going to be more performant than the old "userdata to string" approach
- It is inline with code elsewhere. For example: https://github.com/rspamd/rspamd/blob/a15e3d27f41d18d917f93c25a1f475f09ad14e41/lualib/lua_scanners/avast.lua#L125 
- It allows to rest of `write_dump_content` to not care about type, and use `save_in_file`

lualib/rspamadm/mime.lua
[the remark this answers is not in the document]
JasonStephenson: I would prefer to leave it as `io.write`. I think it is worth keeping it consistant

vstakhov (approved):

PR #4223, opened 2022-07-22
Fix DMARC reporting to subdomains

bilym:
[Fix] DMARC reports for subdomains (of Organizational Domain) with their own DMARC policy.
Reports were wrongly addressed to rua address(es) taken from Organizational Domain policy, not to rua from subdomain policy.

### DIFF
--- a/lualib/rspamadm/dmarc_report.lua
+++ b/lualib/rspamadm/dmarc_report.lua
@@ -345,7 +345,7 @@ end
 local function validate_reporting_domain(reporting_domain)
   -- Now check the domain policy
   -- DMARC domain is a esld for the reporting domain
-  local dmarc_domain = rspamd_util.get_tld(reporting_domain)
+  local dmarc_domain = reporting_domain
   local is_ok, results = rspamd_dns.request({
     config = rspamd_config,
     session = rspamadm_session,
--- a/src/plugins/lua/dmarc.lua
+++ b/src/plugins/lua/dmarc.lua
@@ -309,7 +309,7 @@ local function dmarc_validate_policy(task, policy, hdrfromdom, dmarc_esld)
 
     -- Dmarc domain key must include dmarc domain, rua and period
     local dmarc_domain_key = table.concat(
-        {settings.reporting.redis_keys.report_prefix, dmarc_esld, policy.rua, period},
+        {settings.reporting.redis_keys.report_prefix, policy.domain, policy.rua, period},
         settings.reporting.redis_keys.join_char)
     local report_data = dmarc_common.dmarc_report(task, settings, {
         spf_ok = spf_ok and 'pass' or 'fail',


lualib/rspamadm/dmarc_report.lua
```diff
@@ -345,7 +345,7 @@ end
 local function validate_reporting_domain(reporting_domain)
   -- Now check the domain policy
   -- DMARC domain is a esld for the reporting domain
-  local dmarc_domain = rspamd_util.get_tld(reporting_domain)
+  local dmarc_domain = reporting_domain
```
vstakhov, line 348: The comment says that `DMARC domain is a esld for the reporting domain` but the code does the opposite thing now. And I clearly can say that such a change is incorrect by standard. We need to check the policy for eSLD and then check the subdomains policy.

src/plugins/lua/dmarc.lua
```diff
@@ -309,7 +309,7 @@ local function dmarc_validate_policy(task, policy, hdrfromdom, dmarc_esld)
 
     -- Dmarc domain key must include dmarc domain, rua and period
     local dmarc_domain_key = table.concat(
-        {settings.reporting.redis_keys.report_prefix, dmarc_esld, policy.rua, period},
+        {settings.reporting.redis_keys.report_prefix, policy.domain, policy.rua, period},
```
vstakhov, line 312: Same here.

vstakhov (changes requested):
I think the whole idea of this PR actually contradicts with the [DMARC standard](https://datatracker.ietf.org/doc/html/rfc7489#section-3.2)

bilym:
Let me explain my request in more detail using the following example.

Let's have two domains:

1. example.com with dns dmarc record containing rua=rua1-address (abbreviated notation)
2. sub.example.com with dns dmarc record containing rua=rua2-address

- Mail is coming from user@sub.example.com.
- Rspamd on the receiver side scans this mail, incl. dmarc tests based on dmarc record belonging to the sub.example.com domain. It's ok for now.
- Next, rspamd prepares data for future dmarc reporting as record(object) saved into redis database.
- Redis record contains rua1-address, organizational domain name (OD) example.com and other items.
- Now the problem: Original domain name sub.example.com is lost in this moment for future processing.
- I propose to save the domain name (sub.example.com) where policy is defined, instead of OD name.

- In reporting phase (rspamadm dmarc_report) the program reads records from redis - domain example.com and rua1-address.
- Dmarc record of OD example.com is read from dns and verified (validity, authorization etc).
- In the second half of prepare_report() the mail and report for recipient rua1-address is prepared.
- As the result, dmarc report related to sub.example.com is sent to wrong address rua1-address.

I'm not sw developer. I have prepared the PR as minimalistic code change. Yes, i probably mis-use names of some variables. Sorry, I don't understand what the esld abbreviation is.

slavkoja:
While i cannot comment the code in this PR, i can confirm, that subdomain reporting behavior is broken. The RFC 7489, section 6.6.3.  Policy Discovery clearly states, that if subdomain has its own DMARC record, that record have to be used, and organizational domain have to be checked only if subdomain has not own record. rspamd does this properly in DMARC checks (at least i see appropriate symbols), but wrongly record these messages for reporting under organizational domain.

I am receiving messages from one subdomain, where the organizational domain hasn't own DMARC record at all, only the subdomain has one, but every time then report fails, that (organizational) domain has not valid rua (or so)... I add that domain to reports exceptions map, thus i have no exact errors now.

bilym:
Thanks @fatalbanana for the solution in https://github.com/fatalbanana/rspamd/commit/400826a4583b101215c940dcfd3cbb3dc30f4064

PR #4233, opened 2022-08-02
[Minor] Set `diacritics` flag for more languages

fatalbanana:


vstakhov:
We should maybe add some sanity check that if we have diacritics in some high probability n-gramms, then this language should be considered as `diacritic` language. Such a check should not be hard to add. On the other hand, we have quite a small number of languages to do this check manually :)

Issue #4241, opened 2022-08-16
rspamadm dmarc_report --no-opt still sending reports


rekup:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
We are currently testing the `rspamadm dmarc_report` functionality and observed that even tough the `-n` / `--no-opt` option is set, rspamadm still sends out emails. The help text for this option is `Do not reset reporting data/send reports`

### Steps to Reproduce

1. enable dmarc reporting as described [here](https://rspamd.com/doc/modules/dmarc.html#reporting)
2. wait for some mails which violate the DMARC policy
3. run `rspamadm dmarc_report -v -n `

**Expected behavior**
I would expect to see the generated mails on stdout, but no email should be sent using the SMTP server configured in the `reporting` section of the rspamd config.

### Versions

```bash
[root@mx ~]# rspamd --version
Rspamd daemon version 3.2
[root@mx ~]# rpm -q rspamd
rspamd-3.2-1.x86_64
[root@mx ~]# cat /etc/centos-release
CentOS Linux release 7.9.2009 (Core)
```

PR #4242, opened 2022-08-16
[Fix] Skip sending dmarc reports in no-opt mode

rekup:
Skip sending DMARC reports if the `-n` / `--no-opt` option is set

fixes https://github.com/rspamd/rspamd/issues/4241


vstakhov:
Oh, that got missing during rework, thank you!

PR #4244, opened 2022-08-19
Fix: BAD_REP_POLICIES does not trigger for Bayes

Player701:
When a message is classified as spam by the bayes classifier, the ```BAD_REP_POLICIES``` symbol does not trigger, even though it says in its description that it should. This is because the expression for the symbol contains a reference to a non-existing group "bayes", while the actual group is called "statistics". Proposing a change to fix this.


vstakhov:
That's a good catch, thank you!

Issue #3644, opened 2021-02-11
[Feature] Insert RSpamd headers at the position in chain where the scan happens


strangePo:
## Summary

Currently RSpamd adds its headers at the end of the headers section.
This makes it really hard to spot at which point RSpamd took its action.

## Motivation

Get a clear picture of the chain. 

## Describe alternatives you've considered

There is no alternative.

## Additional context

Example Email Headers:

```
Return-Path: <support@example.com>
Received: from mail01.example.com (127.0.0.1:40786)
	by mail01 (kopano-dagent) with LMTP;
	Thu, 11 Feb 2021 11:32:27 +0100 (CET)
Received: from mail01.example.com (localhost [127.0.0.1])
	by mail01.example.com (Postfix) with ESMTP id EE19B14094A1
	for <support@example.com>; Thu, 11 Feb 2021 11:32:26 +0100 (CET)
Received: from localhost (localhost [127.0.0.1])
	by mail01.example.com (Postfix) with ESMTP id 834A314096D1
	for <support@example.com>; Thu, 11 Feb 2021 11:32:26 +0100 (CET)
X-Virus-Scanned: by amavisd-new-2.10.1 (20141025) (Debian) at example.com
Received: from mail01.example.com ([127.0.0.1])
	by localhost (mail01.example.com [127.0.0.1]) (amavisd-new, port 10024)
	with ESMTP id 73p3X1wZbvXq for <support@example.com>;
	Thu, 11 Feb 2021 11:32:21 +0100 (CET)
Received: from mail01.example.com (localhost [127.0.0.1])
	by mail01.example.com (Postfix) with ESMTP id 3127C14094A1
	for <support@example.com>; Thu, 11 Feb 2021 11:32:21 +0100 (CET)
Received: from mx00.example.com (cipher01.example.com [10.0.5.1])
	by mail01.example.com (Postfix) with ESMTP id 0B24E14094A1
	for <support@example.com>; Thu, 11 Feb 2021 11:32:21 +0100 (CET)
Date: Thu, 11 Feb 2021 11:32:20 +0100
From: support@example.com
To: support@example.com
Subject: test example
Message-ID: <20210211103220.Lbpdd%support@example.com>
User-Agent: s-nail v14.9.11
MIME-Version: 1.0
X-Spamd-Result: default: False [1.00 / 99.00];
	 ARC_NA(0.00)[];
	 TO_MATCH_ENVRCPT_ALL(0.00)[];
[...]

Issue #3398, opened 2020-06-09
[BUG] ARC headers are added after current headers


levitte:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

ARC headers seems to be added at the end of the headers.  Here's an example, where the email goes submisson-\>MTA-\>mailman(local)-\>MTA-\>recipients:

```
Return-Path: <l-sandbox-bounces@openssl.org>
X-Original-To: levitte@mta.openssl.org
Delivered-To: levitte@mta.openssl.org
Received: from mta.openssl.org (localhost [IPv6:::1])
	by mta.openssl.org (Postfix) with ESMTP id 93C79E6695;
	Tue,  9 Jun 2020 07:09:53 +0000 (UTC)
X-Original-To: l-sandbox@openssl.org
Delivered-To: l-sandbox@mta.openssl.org
Received: from lapdog.home.levitte.org (unknown [IPv6:2001:470:28:a97::1:2])
 (using TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits))
 (No client certificate requested)
 by mta.openssl.org (Postfix) with ESMTPSA id 6F360E668A
 for <l-sandbox@openssl.org>; Tue,  9 Jun 2020 07:09:48 +0000 (UTC)
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=openssl.org;
 s=dkim-2020-2; t=1591686591;
 h=from:from:reply-to:subject:subject:date:date:message-id:message-id:
 to:to:cc:mime-version:mime-version:content-type:content-type;
[...]

Issue #1844, opened 2017-09-23
rspamd: headers being added at the very end of header block (affects MIME/PGP)


Yuubari:
### Classification (Please choose *one* option):

* [ ] Crash/Hang/Data loss
* [ ] WebUI/Usability
* [ ] Serious bug
* [X] Other bug
* [ ] Feature
* [ ] Enhancement

### Reproducibility (Please choose *one* option):

* [X] Always
* [ ] Sometimes
* [ ] Rarely
* [ ] Unable
* [ ] I didn’t try
* [ ] Not applicable

### Rspamd version:
1.6.4

### Operation system, CPU, memory and environment:
FreeBSD 11-STABLE, amd64, 24 GB RAM, MTA is Postfix, rspamd milter configuration is as provided in the documentation.

### Description (Please provide a descriptive summary of the issue):
rspamd adds its headers below every other header in the message.

### Steps to reproduce:
Receive an email from an SPF/DKIM enabled relay; or
Send an email from a domain with DKIM signing enabled to any recipient.

### Expected results:
`Authentication-Results` and `DKIM-Signature` (along with other related headers save for `X-Spam` stuff) are inserted *before* every other header (usually after the `Received` header) to keep the header chain in its MTA processing order.

### Actual results:
The headers are inserted *after* the presumed-last header line in the main header block, without regard for MIME and PGP.
Examples:

Sent mail:
```
To: some address at gmail
From: "someone" <someone at DKIM signed domain>
Subject: Test
Message-ID: <b87f7e5d-3678-6d6d-2e41-a1217b380ffb@DKIM signed domain>
Date: Sat, 23 Sep 2017 19:20:55 +09:00
User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64; rv:52.0) Gecko/20100101
 Thunderbird/52.3.0
MIME-Version: 1.0
Content-Type: multipart/signed; micalg=pgp-sha256;
 protocol="application/pgp-signature";
 boundary="nRkmTjRiMlHfbbwfiXJQHoM0heCpkfxwV"
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; ...
```
Received mail:
```
User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64; rv:52.0) Gecko/20100101
 Thunderbird/52.3.0
MIME-Version: 1.0
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 7bit
[...]

PR #4246, opened 2022-08-22
milter_headers: The order set by add_header() was omitted

ikedas:
The header fields added by milter_headers plugin are placed at the bottom, even if order was specified, e.g. authentication-results:
``` code
...

MIME-Version: 1.0
Content-Language: en-US
Content-Type: text/plain; charset=UTF-8; format=flowed
Content-Transfer-Encoding: 7bit
Authentication-Results: example.co.jp;
        none

body
```

This problem has been addressed by issues (afaik) #1844, #3398 and #3644.

### DIFF
--- a/src/libserver/milter.c
+++ b/src/libserver/milter.c
@@ -1781,13 +1781,13 @@ rspamd_milter_process_milter_block (struct rspamd_milter_session *session,
 			it = NULL;
 
 			while ((cur = ucl_object_iterate (elt, &it, true)) != NULL) {
-				ucl_object_iter_t *elt_it;
+				const char *key;
 
-				elt_it = ucl_object_iterate_new (cur);
+				key = ucl_object_key (cur);
 
-				while ((cur_elt = ucl_object_iterate_safe (elt_it, true)) != NULL) {
+				if (key != NULL && (cur_elt = ucl_object_lookup(elt, key)) != NULL) {
 					if (ucl_object_type (cur_elt) == UCL_STRING) {
-						hname = g_string_new (ucl_object_key (cur));
+						hname = g_string_new (key);
 						hvalue = g_string_new (ucl_object_tostring (cur_elt));
 
 						rspamd_milter_send_action (session,
@@ -1798,7 +1798,7 @@ rspamd_milter_process_milter_block (struct rspamd_milter_session *session,
 					}
 					else if (ucl_object_type (cur_elt) == UCL_OBJECT) {
 						rspamd_milter_extract_single_header (session,
-								ucl_object_key (cur), cur_elt);
+								key, cur_elt);
 					}
 					else if (ucl_object_type (cur_elt) == UCL_ARRAY) {
 						/* Multiple values for the same key */
@@ -1810,14 +1810,12 @@ rspamd_milter_process_milter_block (struct rspamd_milter_session *session,
 						while ((array_elt = ucl_object_iterate_safe (array_it,
 								true)) != NULL) {
 							rspamd_milter_extract_single_header (session,
-									ucl_object_key (cur), array_elt);
+									key, array_elt);
 						}
 
 						ucl_object_iterate_free (array_it);
 					}
 				}
-
-				ucl_object_iterate_free (elt_it);
 			}
 		}
 


src/libserver/milter.c
```diff
@@ -1781,13 +1781,13 @@ rspamd_milter_process_milter_block (struct rspamd_milter_session *session,
...
 
 			while ((cur = ucl_object_iterate (elt, &it, true)) != NULL) {
-				ucl_object_iter_t *elt_it;
+				const char *key;
 
-				elt_it = ucl_object_iterate_new (cur);
+				key = ucl_object_key (cur);
 
-				while ((cur_elt = ucl_object_iterate_safe (elt_it, true)) != NULL) {
+				if (key != NULL && (cur_elt = ucl_object_lookup(elt, key)) != NULL) {
```
vstakhov, line 1788: I'm sorry, but your change is totally not clear to me. Initially, the idea was to perform iteration on each header, as it can have multiple value. What you do by `cur_elt = ucl_object_lookup(elt, key)` is efficiently the same as assigning `cur = cur_elt` and nothing else.

vstakhov (changes requested):
I really don't understand your proposed change and it looks like that it breaks the existing logic. Could you please elaborate more on what problem do you have and how do you propose to solve it. Ideally, there should be some test cases.

src/libserver/milter.c
ikedas: I'm not familiar with libucl, but as far as I understood that, if `elt` is a hash, `ucl_object_iterate()` returns the result of [`ucl_hash_iterate2()`](https://github.com/rspamd/rspamd/blob/3.2/contrib/libucl/ucl_hash.c#L310):

https://github.com/rspamd/rspamd/blob/c1b767c23463735b645725419adbf7ef60c7e948/contrib/libucl/ucl_util.c#L2715-L2729

It is _not_ the same as the object we want _but_ each hash element (`ucl_hash_elt *`) in `elt`.  That's why I executed `ucl_object_lookup()` to get `cur_elt`.

Second, I don't understand why the iteration is doubled in the current code.  you say "_it can have multiple value_", however, the multi-value case looks made by handling the case that the type of `cur_elt` is `UCL_ARRAY` (Conversely, since the current code handles the `UCL_ARRAY` case in the second iteration, IMHO the third iteration to process `UCL_ARRAY` is never executed).

Iterating on the `UCL_OBJECT` case makes no sense.  In fact, current code treats a hash like `{value = 'header value', order = 1}` as "multiple value" `'header value'` and `1` (latter is ignored because it is not UC_STRING).

ikedas: > Iterating on the `UCL_OBJECT` case makes no sense. In fact, current code treats a hash like `{value = 'header value', order = 1}` as "multiple value" `'header value'` and `1` (latter is ignored because it is not `UCL_STRING`).

To confirm above, modify `milter_headers.lua` as:
``` diff
--- a/src/plugins/lua/milter_headers.lua
+++ b/src/plugins/lua/milter_headers.lua
@@ -528,7 +528,7 @@ local function milter_headers(task)
             settings.routines['authentication-results']))

     if res then
-      add_header('authentication-results', res, ';', 1)
+      add_header('authentication-results', res, ';', '1')
     end
   end
```
then send a message.  As the result with current code, header modified by rspamd is like:
``` code
...

MIME-Version: 1.0
Content-Language: en-US
Content-Type: text/plain; charset=UTF-8; format=flowed
Content-Transfer-Encoding: 7bit
Authentication-Results: 1
Authentication-Results: example.co.jp;
        none

body
```
Because `add_header()` generates a hash `{value = 'header value', order = '1'}` for `Authentication-Results`. As the type of `'1'` is `UCL_STRING`, it is not ignored.

(Perhaps it is also possible to set `default_headers_order = "1"` (a string), but I haven't tried it.)

vstakhov: Ok, I probably see what do you mean, and there is another solution: `ucl_object_iterate_safe` accepts an argument `expand_values` that either iterate over multiple values (when set to `false`) or over each individual element (when set to `true`). So, in this particular case, the sollution might be just replacing `true` to `false` as it was intended by the logic of the code.

vstakhov: Well, it won't really help, as there is a logical mistake in the code (or in the documentation). Let me think about other possibilities...

vstakhov:
Ok, I think the recent change should really fix the problem by splitting up branches to handle legacy and non-legacy formats completely. I would appreciate if you can test these patches.

ikedas:
OK I'll test it. Please wait for a few days.

Besides I guess [this patch to lua script](https://github.com/rspamd/rspamd/compare/master...ikedas:rspamd:wa/headers_added_at_bottom) may be a workaround for the people who cannot (re-)compile the new release immediately.

ikedas:
@vstakhov , I tested your commits:

  * [Result by 4dd9c48](https://gist.github.com/ikedas/45400238b193f85a7c324c6cb5300a57)
  * [Result by fe9d9d3](https://gist.github.com/ikedas/1814354079f6a465125658dfb5993c68)

vstakhov:
I'm not quite sure that your tests are correct. The problem is that the headers are inserted in the order of the original headers, not modified headers like you might probably expect. The case where `order = nil` is somehow strange, but it is yet a mistake in the user's code...

vstakhov:
Anyway, the bug is really somewhere in the lua code: either in the `milter_headers.lua` plugin (and some of the code comments tell me that it needs refactoring badly) or in the `lua_mime` module that is responsible for interacting with the C API. The current logic in the C code looks correct.

ikedas:
> Anyway, the bug is really somewhere in the lua code: either in the `milter_headers.lua` plugin (and some of the code comments tell me that it needs refactoring badly) or in the `lua_mime` module that is responsible for interacting with the C API. The current logic in the C code looks correct.

How the lua code should be corrected?  In my opinion it should be as follows:

  * The values of `add` table are not legacy strings but objects (`{value=...}`) and/or arrays including objects.
  * The position of inserted header with negative `order` is calculated not based on the original header but on the modified header.
  * The position of inserted header with `order = nil`  is the same as the case when `order` is omitted.

vstakhov:
Uh-oh, one more thing: do you actually use milter or is it mime message modification?

ikedas:
I don’t know the difference, but afaik, my test cases insert/append the new header fields using corresponding milter API and never modifys message body.

Anyways I’ll submit another PR in near future according to your suggestion.

ikedas:
FYI: I realized “`{order = nil}`“ should be “`{order = null}`“ in UCL.

Kennethjohn1234:
hello guys

ikedas:
@vstakhov , currently I am seeing two error messages.

``` code
2022-08-26 03:18:48 #245874(normal) <003e22>; task; rspamd_message_set_modified_header: internal error: calling for set_modified_header with invalid header
```
I found this to be a problem on the lua code, and will include the fix in a PR to be submitted at a later date.

``` code
2022-08-29 00:25:28 #273922(normal) <fa2d7d>; task; rspamd_message_set_modified_header: internal error: calling for set_modified_header with negative add order header
```
This seems to be a problem on libmime: It looks not accepting cases where `order` is _less than_ `-1`.  Any ideas?

vstakhov:
> This seems to be a problem on libmime: It looks not accepting cases where `order` is _less than_ `-1` (however this behavior does not affect the results of milter_headers plugin). Any ideas?

This is `NYI` error: it is not implemented as I can hardly imagine any useful use case for that.

vstakhov:
> I've found this to be a problem on the lua code, and will include the fix in a PR to be submitted at a later date.

That would be great, as I was not able to make it happen :)

ikedas:
Please see #4259.

PR #4259, opened 2022-09-02
milter_headers: Header fields may be inserted at wrong position.

ikedas:
* `lualib/lua_mime.lua`: `modify_headers()` did not handle header altration properly.
  * `libserver/milter.c`: `rspamd_milter_extract_single_header()` inserted the headers with the `order` _less than_ `-1` at unintuitive position.
  * `plugins/lua/milter_headers.lua`: `default_headers_order` was not always used as default order; local function `add_header()` could be simplified.

Test cases are attached (note that the results of some test cases are fluctuating):
[rspamd.tests.tar.gz](https://github.com/rspamd/rspamd/files/9475300/rspamd.tests.tar.gz)

### DIFF
--- a/lualib/lua_mime.lua
+++ b/lualib/lua_mime.lua
@@ -19,6 +19,7 @@ limitations under the License.
 -- This module contains helper functions to modify mime parts
 --]]
 
+local logger = require "rspamd_logger"
 local rspamd_util = require "rspamd_util"
 local rspamd_text = require "rspamd_text"
 local ucl = require "ucl"
@@ -519,21 +520,35 @@ exports.modify_headers = function(task, hdr_alterations)
   local add = hdr_alterations.add or {}
   local remove = hdr_alterations.remove or {}
 
+  local add_headers = {} -- For Milter reply
   local hdr_flattened = {} -- For C API
 
   local function flatten_add_header(hname, hdr)
+    if not add_headers[hname] then
+      add_headers[hname] = {}
+    end
     if not hdr_flattened[hname] then
       hdr_flattened[hname] = {add = {}}
     end
     local add_tbl = hdr_flattened[hname].add
     if hdr.value then
+      table.insert(add_headers[hname], {
+        order = (tonumber(hdr.order) or -1),
+        value = hdr.value,
+      })
       table.insert(add_tbl, {tonumber(hdr.order) or -1, hdr.value})
     elseif type(hdr) == 'table' then
       for _,v in ipairs(hdr) do
-        table.insert(add_tbl, {-1, v})
+        flatten_add_header(hname, v)
       end
     elseif type(hdr) == 'string' then
+      table.insert(add_headers[hname], {
+        order = -1,
+        value = hdr,
+      })
       table.insert(add_tbl, {-1, hdr})
+    else
+      logger.errx(task, 'invalid modification of header: %s', hdr)
     end
   end
   if hdr_alterations.order then
@@ -566,7 +581,7 @@ exports.modify_headers = function(task, hdr_alterations)
   end
 
   task:set_milter_reply({
-    add_headers = hdr_alterations.add,
+    add_headers = add_headers,
     remove_headers = hdr_alterations.remove
   })
 
--- a/src/libserver/milter.c
+++ b/src/libserver/milter.c
@@ -1652,17 +1652,21 @@ rspamd_milter_remove_header_safe (struct rspamd_milter_session *session,
 		hname = g_string_new (key);
 		hvalue = g_string_new ("");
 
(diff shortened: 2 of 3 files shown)


lualib/lua_mime.lua
```diff
@@ -519,21 +520,35 @@ exports.modify_headers = function(task, hdr_alterations)
...
     if hdr.value then
+      table.insert(add_headers[hname], {
+        order = (tonumber(hdr.order) or -1),
+        value = hdr.value,
+      })
       table.insert(add_tbl, {tonumber(hdr.order) or -1, hdr.value})
     elseif type(hdr) == 'table' then
       for _,v in ipairs(hdr) do
-        table.insert(add_tbl, {-1, v})
+        flatten_add_header(hname, v)
```
vstakhov, line 542: 👍

lualib/lua_mime.lua
```diff
@@ -519,21 +520,35 @@ exports.modify_headers = function(task, hdr_alterations)
...
       end
     elseif type(hdr) == 'string' then
+      table.insert(add_headers[hname], {
+        order = -1,
+        value = hdr,
+      })
       table.insert(add_tbl, {-1, hdr})
+    else
+      logger.errx("invalid modification of header: %s", hdr)
```
vstakhov, line 551:
```suggestion
      logger.errx(task, "invalid modification of header: %s", hdr)
```

src/libserver/milter.c
```diff
@@ -1652,17 +1652,21 @@ rspamd_milter_remove_header_safe (struct rspamd_milter_session *session,
...
+			}
 		}
-		else if (nhdr == 0 && ar->len > 0) {
+		else if (nhdr == 0) {
 			/* We need to clear all headers */
 			for (i = ar->len; i > 0; i --) {
 				rspamd_milter_send_action (session,
 						RSPAMD_MILTER_CHGHEADER,
 						i, hname, hvalue);
+				priv->cur_hdr --;
```
vstakhov, line 1669: 👍

src/libserver/milter.c
```diff
@@ -1671,11 +1675,14 @@ rspamd_milter_remove_header_safe (struct rspamd_milter_session *session,
...
 						RSPAMD_MILTER_CHGHEADER,
 						ar->len + nhdr + 1, hname, hvalue);
+				priv->cur_hdr --;
 			}
 		}
 
 		g_string_free (hname, TRUE);
 		g_string_free (hvalue, TRUE);
+
+		g_assert (priv->cur_hdr >= 0);
```
vstakhov, line 1685: While it seems correct, I would suggest to replace it with an `msg_err`. The reason why is that we can get any potentially malformed milter block from the Lua code, but that should not trigger assertions!

src/libserver/milter.c
```diff
@@ -1715,10 +1722,19 @@ rspamd_milter_extract_single_header (struct rspamd_milter_session *session,
 			else {
 				/* Calculate negative offset */
 
-				if (-idx <= priv->cur_hdr) {
+				if (idx == -1) {
+					rspamd_milter_send_action (session,
+							RSPAMD_MILTER_ADDHEADER,
+							hname, hvalue);
```
vstakhov, line 1728: 👍

src/libserver/milter.c
```diff
@@ -1715,10 +1722,19 @@ rspamd_milter_extract_single_header (struct rspamd_milter_session *session,
...
+				}
+				else if (-idx <= priv->cur_hdr) {
+					/*
+					 * Note: We should account MTA's own "Received:" field
+					 * which wasn't passed by Milter's header command.
+					 */
```
vstakhov, line 1734: 👍
However, some MTA can include that header in Milter dialog sadly. Postfix and Sendmail are fine here to my very best knowledge.

vstakhov (approved):
I have some small neats, but this PR looks very good and useful, thank you very much!

src/libserver/milter.c
ikedas: I also guessed it possible (I tested only with Postfix), so I isolated the `-1` case.  Thanks for confirming my concern.

lualib/lua_mime.lua
ikedas: Oops, I'll re-push correction. Thanks.

src/libserver/milter.c
ikedas: OK, I'll re-push the change.

ikedas:
I've corrected the things pointed out and re-pushed.  Thank you for suggestions!

vstakhov:
Thank you for working on this issue, much appreciated!

PR #4268, opened 2022-09-14
[Fix] Avoid overriding IP with Sender IP

dragoangel:
This PR provides a fix for:
1. Not override IP with a user-supplied header
2. DoS in writing logs to Elastic which external sender can exploit. IP validation in Elastic is stricter than in Rspamd, so the user can supply data that looks for Rspamd as valid IP, f.e.: `.23.24.1`, but fail to pass validation on Elastic side. Now basic string is used which should not result in such errors

### DIFF
--- a/src/plugins/lua/elastic.lua
+++ b/src/plugins/lua/elastic.lua
@@ -121,27 +121,24 @@ local function get_general_metadata(task)
   local r = {}
   local ip_addr = task:get_ip()
 
-  r.webmail = false
-
   if ip_addr  and ip_addr:is_valid() then
     r.is_local = ip_addr:is_local()
-    local origin = task:get_header('X-Originating-IP')
-    if origin then
-      origin = string.sub(origin, 2, -2)
-      local rspamd_ip = require "rspamd_ip"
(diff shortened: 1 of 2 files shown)


src/plugins/lua/elastic.lua
```diff
@@ -121,29 +121,26 @@ local function get_general_metadata(task)
...
-    else
-      r.ip = tostring(ip_addr)
-    end
+    r.ip = tostring(ip_addr)
   else
     r.ip = '127.0.0.1'
   end
 
+  r.webmail = false
+  r.sender_ip = 'unknown'
```
vstakhov, line 132: Why do we need `webmail` and `sender_ip` both?

dragoangel: I thought about existing people setups (possibly filters, etc) so decided to leave webmail bool as-is, to not break it for existing setups. Generally, if `sender_ip` isn't unknown then it means that it was sent via webmail, yes.

vstakhov: No, I think your arguments are perfectly valid, thank you!

Issue #4281, opened 2022-09-23
[BUG] Content corruption on empty body


JasonStephenson:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Unexpected data in content field**
When using rspamd to decompose mime into msgpack/ucl/json we are seeing incorrect data for the content. I highly suspect this is related to a buffer overrun as the value here often appears elsewhere in the email.

### Steps to Reproduce
![image](https://user-images.githubusercontent.com/5558488/192043513-4cf054b9-a8ad-4e8b-9655-b1e74cb15cba.png)
[1.txt](https://github.com/rspamd/rspamd/files/9636673/1.txt)

1. Create a .eml from the content above
2. `rspamadm mime dump --ucl 1.eml`
3. Check output below
4. While in this run it is garbage I have seen it have `content ="ipients:;"`

[1.ucl.txt](https://github.com/rspamd/rspamd/files/9636695/1.ucl.txt)

**Expected behavior**
Content should be blank.

### Versions
3.3 latest
3.2

JasonStephenson:
Introduced by https://github.com/rspamd/rspamd/commit/510b61f82bae20b9d8c2b61b51a35d9579c2d347

Previously this part wouldn't have been written out because of a 0 length. Is the fix as simple as checking length != 0 again?

JasonStephenson:
Confirmed fixed

PR #4282, opened 2022-09-23
Don't print empty part or it'll dump random chunks of memory

JasonStephenson:
I believe this fixes https://github.com/rspamd/rspamd/issues/4281


vstakhov:
Is it needed with https://github.com/rspamd/rspamd/commit/bc6ec09a56659b35317ac4413a43e3bf47bc0e28 ?

JasonStephenson:
You're right, it's not needed 👍 . Any chance your change could be patched into 3.2?

vstakhov:
I plan to make 3.3 release quite soon. I'm now polishing some issues about leaks and incorrect behaviour. After those fixes being tested in production, the new release could be cut off.

Issue #4278, opened 2022-09-20
[BUG] fuzzy-worker listening on unix socket leads to assertion crash on learn


BtbN:
**Describe the bug**
When the fuzzy-worker is configured to listen on a Unix-Domain-Socket, rspamd logs an assertion error on rspamc fuzzy_add:
```
Sep 20 16:39:53 btbn.de rspamd[3822225]: <>; glib; rspamd_glib_printerr_function: **\x0AERROR:/var/tmp/portage/mail-filter/rspamd-3.2/work/rspamd-3.2/src/libutil/addr.c:1667:rspamd_inet_address_from_sa: assertion failed: (slen >= sizeof (struct sockaddr))\x0A
Sep 20 16:39:53 btbn.de rspamd[3822225]: <>; ; rspamd_crash_sig_handler: caught fatal signal 6(Aborted), pid: 3822225, trace:
Sep 20 16:39:53 btbn.de rspamd[3822225]: Bail out! ERROR:/var/tmp/portage/mail-filter/rspamd-3.2/work/rspamd-3.2/src/libutil/addr.c:1667:rspamd_inet_address_from_sa: assertion failed: (slen >= sizeof (struct sockaddr))
Sep 20 16:39:53 btbn.de rspamd[3822225]: <>; ; rspamd_crash_sig_handler: please see Rspamd FAQ to learn how to dump core files and how to fill a bug report
Sep 20 16:39:54 btbn.de rspamd[3669153]: <f11b89>; main; rspamd_check_termination_clause: fuzzy process 3822225 terminated abnormally by signal: Aborted and created core file; please see Rspamd FAQ to learn how to extract data from core file and fill a bug report
Sep 20 16:39:54 btbn.de rspamd[3669153]: <f11b89>; main; rspamd_cld_handler: respawn process fuzzy in lieu of terminated process with pid 3822225
Sep 20 16:39:55 btbn.de rspamd[3669244]: <218ed4>; task; fuzzy_controller_io_callback: got error in IO with server /run/rspamd/fuzzy.sock(unix:/run/rspamd/fuzzy.sock), 111, Connection refused
```

Unsurprisingly, adding the mail also fails with "HTTP error: 111, write socket error"

### Steps to Reproduce

1. Have fuzzy worker listening on unix domain socket. For example:
```
worker "fuzzy" {bind_socket = "/run/rspamd/fuzzy.sock mode=0660"; }
```
2. try to add a mail via rspamc fuzzy_add

[...]

PR #4283, opened 2022-09-25
Allow fuzzy worker to listen on AF_UNIX sockets successfully

BtbN:
This aims to fix https://github.com/rspamd/rspamd/issues/4278

The first action is to bind a DGRAM unix client-socket to an anonymous abstract socket.
This is a special case, as documented on man 7 unix: If the address length is sizeof(sa_family_t), it will automatically assign an abstract free for use address for the socket.
Thus, the server has some address to reply to.

This uncovered a few further issues:
rspamd_inet_address_from_sa was not prepared for such addresses.
For one, they can be very short, so the default length check failed.
Then they do contain a null byte in the string, so any strlen/cpy functions will fail. Instead just memcpy the whole struct with the given slen.

Additionally, the allow_update check in the fuzzy worker makes no sense for unix sockets, since there isn't really any IPs to check, so the check always fails, preventing any kind of write-access.
It was thus extended to always succeed the address check on an AF_UNIX socket, no matter what the configured address is. This matches how a lot of other places seem to treat unix sockets.

I have deployed this set to my rspamd instance, backported on top of 3.2, and it seems to work fine so far.

The main issue I see with this is that the autobind behavior seems to be entirely Linux-Specific.
An idea to deal with this would be to just drop the "return -1" from that bind, so it just prints an informational message on other systems, but otherwise continues on as it would right now, at least not making the situation worse.

The entire block could also be put into a preprocessor-check for Linux, which throws an error on other OS to indicate that the given configuration does not work.

### DIFF
--- a/src/libutil/addr.c
+++ b/src/libutil/addr.c
@@ -1060,6 +1060,19 @@ rspamd_inet_address_connect (const rspamd_inet_addr_t *addr, gint type,
 
 	if (addr->af == AF_UNIX) {
 		sa = (const struct sockaddr *)&addr->u.un->addr;
+
+		if (type == (int)SOCK_DGRAM) {
+			struct sockaddr ca;
+
+			memset (&ca, 0, sizeof(ca));
+			ca.sa_family = AF_UNIX;
+
+			r = bind (fd, &ca, sizeof (sa_family_t));
+			if (r == -1) {
+				msg_info ("unix socket client autobind failed: %s, '%s'",
+						addr->u.un->addr.sun_path, strerror (errno));
+			}
+		}
 	}
 	else {
 		sa = &addr->u.in.addr.sa;
--- a/src/libutil/addr.c
+++ b/src/libutil/addr.c
@@ -1680,7 +1680,8 @@ rspamd_inet_address_from_sa (const struct sockaddr *sa, socklen_t slen)
 	rspamd_inet_addr_t *addr;
 
 	g_assert (sa != NULL);
-	g_assert (slen >= sizeof (struct sockaddr));
+	/* Address of an AF_UNIX socket can be tiny */
+	g_assert (slen >= sizeof (sa_family_t) + 1);
 
 	addr = rspamd_inet_addr_create (sa->sa_family, NULL);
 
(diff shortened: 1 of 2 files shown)


src/libutil/addr.c
```diff
@@ -1060,6 +1060,21 @@ rspamd_inet_address_connect (const rspamd_inet_addr_t *addr, gint type,
...
 	if (addr->af == AF_UNIX) {
 		sa = (const struct sockaddr *)&addr->u.un->addr;
+
+		if (type == (int)SOCK_DGRAM) {
+			struct sockaddr ca;
+
+			memset (&ca, 0, sizeof(ca));
+			ca.sa_family = AF_UNIX;
+
+			r = bind (fd, &ca, sizeof (sa_family_t));
```
vstakhov, line 1070: It will likely break all unix sockets on non-Linux OSes, which is not very desired.

src/libutil/addr.c
```diff
@@ -1060,6 +1060,21 @@ rspamd_inet_address_connect (const rspamd_inet_addr_t *addr, gint type,
...
+			if (r == -1) {
+				close (fd);
```
vstakhov, line 1072: I'm more inclined to treat bind errors as non-fatal tbh.

src/libutil/addr.c
```diff
@@ -1060,6 +1060,21 @@ rspamd_inet_address_connect (const rspamd_inet_addr_t *addr, gint type,
...
+				msg_info ("client autobind failed: %d, '%s'",
+						errno, strerror (errno));
```
vstakhov, line 1074:
```suggestion
				msg_info ("unix socket client autobind failed: %s, '%s'",
						addr->u.un->addr.sun_path, strerror (errno));
```

It is way better to see a socket name than errno code in my opinion.

src/libutil/addr.c
```diff
@@ -1676,12 +1692,13 @@ rspamd_inet_address_from_sa (const struct sockaddr *sa, socklen_t slen)
 		const struct sockaddr_un *un = (const struct sockaddr_un *)sa;
 
 		g_assert (slen >= SUN_LEN (un));
+		g_assert (slen <= sizeof (addr->u.un->addr));
```
vstakhov, line 1695: 👍

vstakhov (changes requested):
The biggest issue is that autobind should be done merely on Linux to avoid breaking for other platforms.

src/libutil/addr.c
```diff
@@ -1676,12 +1692,13 @@ rspamd_inet_address_from_sa (const struct sockaddr *sa, socklen_t slen)
...
 
-		rspamd_strlcpy (addr->u.un->addr.sun_path, un->sun_path,
-				sizeof (addr->u.un->addr.sun_path));
-#if defined(FREEBSD) || defined(__APPLE__)
-		addr->u.un->addr.sun_len = un->sun_len;
-#endif
+		/* sun_path can legally contain intermittent NULL bytes */
```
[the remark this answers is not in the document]
BtbN: An Auto-Bound abstract socket path starts with a null byte. Hence this change is neccesary for the fuzzy worker to be able to call this function on the messages sender address.

vstakhov:
Great, thank you!

PR #4294, opened 2022-10-04
Fix metadata_exporter with many recipients

yo000:
Hello and many thanks for this great software :)

This commit fix sending mail alert with backend "send_mail" and many recipients.
Config I use :
```
rules {
SEND_ALERT_ON_SPAM {
  selector = "is_spam";  
  backend = "send_mail";  
  smtp = "smtp.example.org";  
  smtp_port = 25;  
  helo = "localhost.example.intra";  
  mail_to = ["recipent1@example.org","recipient2@example.org"];  
  mail_from = "rspamd@example.org";  
  formatter = "email_alert";  
}  
```

Without this fix, mail is never sent, sendmail_cb is never called.


vstakhov:
Good catch, thank you!

Issue #4301, opened 2022-10-08
[BUG] Wrong Content-Type for favicon.ico


GermanCoding:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

The rspamd web UI controller sends favicons with incorrect Content-Type HTTP headers. This causes issues on some browsers and can cause effects such as favicon not loading correctly.

### Steps to Reproduce

1. Start rspamd with controller listening on localhost:11334
2. Run `curl -v http://localhost:11334/favicon.ico`

**Expected behavior**

Favicon is returned with `Content-Type: image/x-icon` or `Content-Type: image/vnd.microsoft.icon`

**Actual behavior**

```
curl -v http://localhost:11334/favicon.ico
*   Trying 127.0.0.1:11334...
* Connected to localhost (127.0.0.1) port 11334 (#0)
> GET /favicon.ico HTTP/1.1
> Host: localhost:11334
> User-Agent: curl/7.74.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Connection: close
< Server: rspamd/3.3
< Date: Sat, 08 Oct 2022 22:08:28 GMT
< Content-Length: 10734
< Content-Type: text/plain
< Access-Control-Allow-Origin: *
< Cache-Control: no-store
<
```

Content-Type returned is `text/plain`, which leads to browsers loading this ressource incorrectly, or not at all. Especially applies if No-Sniff header is set by reverse proxy.

### Versions

Rspamd daemon version 3.3
OS: Debian 11 x64

### Additional Information

N/A

GermanCoding:
[...]

PR #4302, opened 2022-10-09
[Fix] Fix favicon.ico Content-Type header

moisseev:
Issue: #4301

### DIFF
--- a/src/libserver/http/http_router.c
+++ b/src/libserver/http/http_router.c
@@ -28,6 +28,7 @@ enum http_magic_type {
 	HTTP_MAGIC_HTML,
 	HTTP_MAGIC_CSS,
 	HTTP_MAGIC_JS,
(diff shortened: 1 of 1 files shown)


src/libserver/http/http_router.c
```diff
@@ -40,6 +41,7 @@ static const struct _rspamd_http_magic {
 		[HTTP_MAGIC_HTML] = { "html", "text/html" },
 		[HTTP_MAGIC_CSS] = { "css", "text/css" },
 		[HTTP_MAGIC_JS] = { "js", "application/javascript" },
+		[HTTP_MAGIC_PNG] = { "png", "image/x-icon" },
```
GermanCoding, line 44: Is there a copy-paste error here? Shouldn't this line read:

`[HTTP_MAGIC_ICO] = { "ico", "image/x-icon" },`

instead?

PR #4303, opened 2022-10-09
[Minor] Increase score for BAD_REP_POLICIES

twesterhever:
From my experience with various rspamd setups, the current score of `BAD_REP_POLICIES` seems to be too low, as it does not fully counterbalance the positive scores of a message having valid SPF/DKIM/DMARC policies. [Some projects](https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/rspamd/local.d/groups.conf#L9-L11) even go for a higher score, however, increasing the rules' score from 0.1 to 0.5 seems like a sensible conservative approach to me for the time being.

### DIFF
--- a/conf/composites.conf
+++ b/conf/composites.conf
@@ -125,7 +125,7 @@ composites {
   BAD_REP_POLICIES {
     description = "Contains valid policies but are also marked by fuzzy/bayes/surbl/rbl";
     expression = "(~g-:policies) & (-g+:fuzzy | -g+:statistics | -g+:surbl | -g+:rbl)";
-    score = 0.1;
+    score = 0.5;
   }
 
   VIOLATED_DIRECT_SPF {


moisseev:
Would you explain what  counterbalance you expect and why do we need it?

slavkoja:
Notice, that main purpose of ``BAD_REP_POLICIES`` composite symbol is to neutralize (remove) negative (good) score of multiple symbols, thus its result on final score is/can be much more than +0.1.

twesterhever:
Apologies for my unclear initial description, and thanks for the replies. I am aware of `BAD_REP_POLICIES`' nature - my suggestion just stems from spam observance, where an increased score of this symbol would have helped to push more of them over the reject threshold.

Large `rspamd` installations with reputation tracking enabled will probably need this less, since they already have learned the senders' IP/DKIM/SPF selector as spammy, adding extra points to the message. For smaller installations, they probably won't have such a comprehensive reputation dataset already, due to their limited traffic visibility - here, a higher `BAD_REP_POLICIES` score could help to counterbalance the missing "badness" reputation tracking would have otherwise gained.

Hope to have answered more questions than rising new ones... :-)

moisseev:
Increasing score of the `BAD_REP_POLICIES` symbol effectively means just amplification of all fuzzy/bayes/surbl/rbl scores, but only when policies are valid.
I would selectively increase fuzzy/bayes/surbl/rbl symbols scores as needed instead.

PR #4304, opened 2022-10-09
[Enhancement] Add composite rule against AFF involving freemailers

twesterhever:
This composite rule has been in use for quite a while in various rspamd setups. Since messages originating from freemail providers tend to have more signs of positive reputation, thanks to IP-/DKIM-/FQDN-reputation tracking, I found the existing symbols, such as `R_UNDISC_RCPT` insufficient to push advance free fraud (AFF) messages "over the edge".

Therefore, this composite rule puts additional weight on messages exhibiting strong AFF patterns, and involve freemail addresses, such as the plethora of such mails emitted by GMail, in the hope to gain better catch rates on these.

### DIFF
--- a/conf/composites.conf
+++ b/conf/composites.conf
@@ -154,6 +154,13 @@ composites {
     score = 7.0;
     group = "scams";
   }
+  
+  FREEMAIL_AFF {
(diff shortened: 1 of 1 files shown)


vstakhov:
Thank you for your contribution, this rule seems to be quite sane and it should not have clear FP cases.

PR #4305, opened 2022-10-09
[Minor] Fix copy-paste error

moisseev:
PR: #4302


vstakhov:
I was also confused by `png` but thought that these favicons are nowadays in png format. Thanks for a quick fix anyway!

PR #4308, opened 2022-10-12
Penalize bounce spam

frederikbosch:
This PR includes two changes.

1. It suggests a default rule to include a penalty for bounce messages with AUTH_NA to protect against fake bounce messages (bounce spam).
2. It adds AUTH_NA_OR_FAIL, because when e.g. R_DKIM_PERMFAIL is included R_DKIM_NA does apply anymore and AUTH_NA is not added. This creates an opportunity to willingly create a R_DKIM_PERMFAIL by a spammer.

### DIFF
--- a/conf/composites.conf
+++ b/conf/composites.conf
@@ -53,7 +53,17 @@ composites {
     expression = "R_DKIM_NA & R_SPF_NA & DMARC_NA & ARC_NA";
     score = 1.0;
     policy = "remove_weight";
-    description = "Authenticating message via SPF/DKIM/DMARC/ARC not possible";
+    description = "Authenticating message via SPF/DKIM/DMARC/ARC not available";
+  }
+  AUTH_NA_OR_FAIL {
+    expression = "!AUTH_NA & (R_DKIM_NA | R_DKIM_TEMPFAIL | R_DKIM_PERMFAIL) & (R_SPF_NA | R_SPF_DNSFAIL) & DMARC_NA & (ARC_NA | ARC_DNSFAIL)";
+    score = 1.0;
+    policy = "remove_weight";
+    description = "Authenticating message via SPF/DKIM/DMARC/ARC not available or failed";
+  }
+  BOUNCE_NO_AUTH {
+    expression = "(AUTH_NA | AUTH_NA_OR_FAIL) & (BOUNCE | SUBJ_BOUNCE_WORDS)";
+    score = 1.0;
   }
   DKIM_MIXED {
     expression = "-R_DKIM_ALLOW & (R_DKIM_TEMPFAIL | R_DKIM_PERMFAIL | R_DKIM_REJECT)"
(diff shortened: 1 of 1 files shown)


conf/composites.conf
```diff
@@ -53,7 +53,17 @@ composites {
     expression = "R_DKIM_NA & R_SPF_NA & DMARC_NA & ARC_NA";
     score = 1.0;
     policy = "remove_weight";
-    description = "Authenticating message via SPF/DKIM/DMARC/ARC not possible";
+    description = "Authenticating message via SPF/DKIM/DMARC/ARC not available";
+  }
+  AUTH_NA_OR_FAIL {
+    expression = "!(R_DKIM_NA & R_SPF_NA & DMARC_NA & ARC_NA) & (R_DKIM_NA | R_DKIM_TEMPFAIL | R_DKIM_PERMFAIL) & (R_SPF_NA | R_SPF_DNSFAIL) & DMARC_NA & (ARC_NA | ARC_DNSFAIL)";
```
vstakhov, line 59: I see `R_DKIM_TEMPFAIL` twice in the second braces, it is not likely intended.

conf/composites.conf
```diff
@@ -53,7 +53,17 @@ composites {
...
+    score = 1.0;
+    policy = "remove_weight";
+    description = "Authenticating message via SPF/DKIM/DMARC/ARC not available or failed";
```
vstakhov, line 62: This message is not very descriptive to me. What is `via` here, for example? SPF/DMARC are operating on different definitions of the sender's address...

vstakhov (approved):
I see the idea and it look sane to me but I would like to see more clarification and arguments if possible.

conf/composites.conf
frederikbosch: Now I think of it, maybe `AUTH_FAIL` is a better description and it would be easier to write it as: `R_DKIM_TEMPFAIL | R_DKIM_PERMFAIL | R_SPF_DNSFAIL | ARC_DNSFAIL`.

conf/composites.conf
frederikbosch: I copied this description from `AUTH_NA` and added `or failed`.

conf/composites.conf
frederikbosch: It is because recursive composites are not allowed. If it would, it says.

`!AUTH_NA & (R_DKIM_NA | R_DKIM_TEMPFAIL | R_DKIM_PERMFAIL) & (R_SPF_NA | R_SPF_DNSFAIL) & DMARC_NA & (ARC_NA | ARC_DNSFAIL)`.

So, this applies when AUTH_NA is not applied, but still no authentication method is succeeding. Therefore I called it AUTH_NA_OR_FAIL.

frederikbosch: I could simplify the name to `AUTH_FAIL`.

PR #4310, opened 2022-10-15
[Enhancement] Add IPFS URL heuristic

twesterhever:
Given IPFS' popularity among miscreants for phishing hosting and malware dissemination, the presence of an URL containing both "ipfs" as well as a random string reminiscent of an IPFS content identifier is a strong sign of maliciousness (I have never seen a legitimate IPFS URL so far, certainly not in mail traffic).

Please note that while IPFS CIDv0 are easy to parse due to their fixed syntax, CIDv1 neither have a fixed length nor any other static character sets. To avoid miscreants bypassing this heuristic by increasing the size of hashing algorithms used, the CIDv1 rexep is rather fuzzy, catching anything alphanumeric between 45 and 256 characters. Most CIDv1s seen so far, however, stayed between 60 and 120 characters.

See https://docs.ipfs.tech/concepts/content-addressing/ for details on CIDs, and how to parse them.


rules/regexp/misc.lua
```diff
@@ -99,4 +99,19 @@ reconf['LEAKED_PASSWORD_SCAM_RE'] = {
...
-rspamd_config:register_dependency('LEAKED_PASSWORD_SCAM', 'BITCOIN_ADDR')
\ No newline at end of file
+rspamd_config:register_dependency('LEAKED_PASSWORD_SCAM', 'BITCOIN_ADDR')
+
+-- Heurististic for detecting InterPlanetary File System (IPFS) gateway URLs:
+-- These contain "ipfs" somewhere (either in the FQDN or the URL path) and a
+-- content identifier (CID), comprising of either "qm", followed by 46 alphanumerical
+-- characters (CIDv0), or a CIDv1 of an alphanumerical string of unspecified length,
+-- depending on the hash algorithm used.
+local ipfs_cid = '/(qm[a-z0-9]{44}|[a-z0-9]{45,256})/{url}i'
```
vstakhov, line 109: This regexp will be very bad from the [performance](https://intel.github.io/hyperscan/dev-reference/performance.html#bounded-repeats-in-streaming-mode) considerations for Hyperscan (and probably PCRE as well). I'd appreciate if we can use something else but `{45,256}` here.

vstakhov (changes requested):
The rule looks good itself, but I'd suggest to think about performance considerations

rules/regexp/misc.lua
twesterhever: Thank you. I will look into this, and try to come up with a less costly regexp for parsing IPFS CIDv1s.

twesterhever:
> The rule looks good itself, but I'd suggest to think about performance considerations

Having worked through the CIDv1 specification, the only things we can do against the performance costs of this regexp getting out of hand are:

- Check whether a possible CIDv1 string starts with a [multibase prefix](https://github.com/multiformats/multibase)
- Limit the anticipated size of the total CIDv1 to something like 128 bytes (I have seen 110-bytes CIDv1s in the wild, so anything shorter does not seem to make sense).

I added commits implementing these changes. @vstakhov: What do you think of it?

vstakhov:
@citrin has some thoughts about this rule, so he will probably comment a little later. Thank you for working on that!

twesterhever:
- ping -

twesterhever:
Are there further changes needed from my end? I continue to observe IPFS phishing frequently, and would love to see this rule making it into rspamd, to provide better detection to its users.

PR #4320, opened 2022-10-29
headers: Also treat pound (£) as currency

dasJ:


slavkoja:
I am not sure about hyperscan support, but what about Unicode Currency regex (``\p{Sc}``)? With ``grep -P`` it works...

``file.txt``:
```
1 $
2 €
3 $
4 ¢
5 ¥
6 ₽
7 £
```
And then grep finds all:
```
grep -P '\p{Sc}' file.txt | wc -l
7
```

PR #4324, opened 2022-11-05
Support external maps in Rspamd

vstakhov:
The idea is to extend the existing maps functionality by allowing them to query external resources via HTTP(s).
Here is a sample configuration:

```
map = {
  external = true;
  backend = "https://example.com/${domain}";
  method = "form"; # or header, or body, or query
  encode = "json"; # or message-pack, or form-encode
}
```

So when a module requires some key from a map, it can pass some additional callback to be called upon query. This allows to do an async call and return back at some point.

TODO list:

- [x] Allow callbacks in maps methods
- [x] Implement `external` maps schema
- [x] Implement calls to external HTTP
- [x] Allow different encoding
- [ ] Add external maps support to the Rspamd plugins (e.g. by adopting selectors)
- [ ] Add documentation for the external maps
- [ ] Add release notes article
- [ ] Add tests for external maps

### DIFF
--- a/lualib/lua_maps.lua
+++ b/lualib/lua_maps.lua
@@ -99,9 +99,14 @@ local rspamd_http = require "rspamd_http"
 local ucl = require "ucl"
 
 local function url_encode_string(str)
-  -- TODO: implement encoding
+  str = string.gsub(str, "([^%w _%%%-%.~])",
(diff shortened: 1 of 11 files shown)


vstakhov:
I actually plan to move plugins adoptation to the separate PR/commits, as this type of support can be added without breaking the existing code.

PR #4327, opened 2022-11-07
[Test] Use TMPDIR if available

arkamar:
Couple of tests were designed to use hardcoded `/tmp` directory, which could be problematic in some cases. For example, in case of sqlite3 related tests, multiple users cannot run it in parallel. Additionally some leftovers remain in `/tmp` directory after this test, `*.sqlite-shm` and `*.sqlite-wal` files, which was actually main motivation for me to create this PR.

From perspective of Gentoo packager `$TMPDIR` is set for the package to a safe place controlled by a package manager, which is automatically cleaned when something remains there. Moreover, I can run tests for multiple versions in parallel without risk of file collision.

### DIFF
--- a/src/libutil/cxx/file_util.cxx
+++ b/src/libutil/cxx/file_util.cxx
@@ -368,13 +368,22 @@ TEST_CASE("check lock") {
 	CHECK(serrno == ENOENT);
 }
 
+auto get_tmpdir() -> const char * {
+	const auto *tmpdir = getenv("TMPDIR");
+	if (tmpdir == nullptr) {
+		tmpdir = G_DIR_SEPARATOR_S "tmp";
+	}
+	return tmpdir;
+}
+
 TEST_CASE("tempfile") {
 	std::string tmpname;
+	const std::string tmpdir{get_tmpdir()};
 	{
-		auto raii_locked_file = raii_locked_file::mkstemp("/tmp//doctest-XXXXXXXX",
+		auto raii_locked_file = raii_locked_file::mkstemp(std::string(tmpdir + G_DIR_SEPARATOR_S + "doctest-XXXXXXXX").c_str(),
 				O_RDONLY, 00600);
 		CHECK(raii_locked_file.has_value());
-		CHECK(raii_locked_file.value().get_dir() == "/tmp");
+		CHECK(raii_locked_file.value().get_dir() == tmpdir);
 		CHECK(access(raii_locked_file.value().get_name().data(), R_OK) == 0);
 		auto raii_locked_file2 = raii_locked_file::open(raii_locked_file.value().get_name().data(), O_RDONLY);
 		CHECK(!raii_locked_file2.has_value());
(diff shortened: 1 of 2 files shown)


src/libutil/cxx/file_util.cxx
```diff
@@ -368,13 +368,22 @@ TEST_CASE("check lock") {
...
 }
 
+const char * get_tmpdir() {
+	const char *tmpdir = getenv("TMPDIR");
+	if (tmpdir == nullptr) {
+		tmpdir = G_DIR_SEPARATOR_S "tmp";
+	}
+	return tmpdir;
+}
+const std::string tmpdir{get_tmpdir()};
```
vstakhov, line 378:
```suggestion
static const std::string tmpdir{get_tmpdir()};
```

vstakhov: Or probably it should be a `TEST_CASE` local variable - even better.

src/libutil/cxx/file_util.cxx
```diff
@@ -368,13 +368,22 @@ TEST_CASE("check lock") {
...
+
 TEST_CASE("tempfile") {
 	std::string tmpname;
 	{
-		auto raii_locked_file = raii_locked_file::mkstemp("/tmp//doctest-XXXXXXXX",
+		auto raii_locked_file = raii_locked_file::mkstemp(std::string(tmpdir + "//doctest-XXXXXXXX").c_str(),
```
vstakhov, line 383:
```suggestion
		auto raii_locked_file = raii_locked_file::mkstemp(std::string(tmpdir + G_DIR_SEPARATOR_S + "doctest-XXXXXXXX").c_str(),
```

src/libutil/cxx/file_util.cxx
```diff
@@ -391,10 +400,10 @@ TEST_CASE("tempfile") {
 TEST_CASE("mmap") {
 	std::string tmpname;
 	{
-		auto raii_file = raii_file::mkstemp("/tmp//doctest-XXXXXXXX",
+		auto raii_file = raii_file::mkstemp(std::string(tmpdir + "//doctest-XXXXXXXX").c_str(),
```
vstakhov, line 403:
```suggestion
		auto raii_file = raii_file::mkstemp(std::string(tmpdir + G_DIR_SEPARATOR_S + "doctest-XXXXXXXX").c_str(),
```

vstakhov (approved):
The suggested changes look good in general, however, I'd suggest some minor neats.

arkamar:
I changed it according to your suggestions.

arkamar:
Those failing tests in CI do not seem to be related to my changes:
```
==============================================================================
Lua Check                                                             | FAIL |
Checking src/plugins/lua/antivirus.lua            1 warning

    src/plugins/lua/antivirus.lua:151:1: (W611) line contains only whitespace

Total: 1 warning / 0 errors in 183 files
------------------------------------------------------------------------------
Cases.200 Luacheck                                                    | FAIL |
1 test, 0 passed, 1 failed
==============================================================================
```

PR #4328, opened 2022-11-07
[Feature] Enable Mime part filters on antivirus module

jonmoesli:
Hi, I hope it could be useful to have mime_parts_filter_regex mime_parts_filter_ext filters as in the _External services_ module.


vstakhov:
Could you please check https://ci.rspamd.com/rspamd/rspamd/3319/1/7 as there is a line with just whitespaces that confuses `luacheck`.

jonmoesli:
I removed the whitespaces but the check still fails :-(

vstakhov:
It's probably introduced by something else then. Indeed, l611 is not in the scope of this PR.

PR #4365, opened 2022-12-22
[Fix] Return true from has_urls(true) if only emails are present

cpragadeesh:
`has_urls(true)` returns false if only email addresses are present in an email. It should instead return true and this is a fix for it.


vstakhov:
Sure thing! I completely agree with the changes you have suggested and am happy to accept them.

PR #4367, opened 2023-01-12
[Fix] received: filtering of artificial header

kakochang:


kakochang:
> Oh, that's a good catch, thank you!

You are welcome!
This is a wonderful software, thank you very much!

PR #4376, opened 2023-01-19
[Fix] Replace broken strict_domains with phishing_exceptions

dupondje:
The syntax of the strict_domains is broken since parsing for it was removed in commit ce21f8fe16a45a5a409d3c9f8217eba2de1fbe14 As this can be done with the exception mapping, we can safely remove it.

### DIFF
--- a/src/plugins/lua/phishing.lua
+++ b/src/plugins/lua/phishing.lua
@@ -33,8 +33,8 @@ local openphish_symbol = 'PHISHED_OPENPHISH'
 local phishtank_symbol = 'PHISHED_PHISHTANK'
 local generic_service_name = 'generic service'
 local domains = nil
-local strict_domains = {}
-local exceptions_maps = {}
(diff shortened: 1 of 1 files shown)


vstakhov:
Hum, we cannot **safely** remove something, as it clearly breaks backward compatibility :( I will have to re-add that.

dupondje:
The syntax was broken since almost 2 years (since the above commit) and didn't work. So that should be safe no?

vstakhov:
I don't know how was it broken as it has successfully passed this test all the time: https://github.com/rspamd/rspamd/blob/master/test/functional/cases/001_merged/114_phishing.robot#L18

dupondje:
The documented syntax at least didn't work anymore since that commit in 2021:

`strict_domains = [
		"${CONFDIR}/paypal.map:PAYPAL_PHISHING"
	];`

Issue #4373, opened 2023-01-19
[Feature] rspamc - Option to have human readable report like spamc of spamassassin


amishmm:
## Summary
I have a program which calls "spamc -R" of spamassassin and gets the score/threshold (as first line) followed by **the human readable spam report** back from spamd. When a message is identified as spam (i.e. score >= threshold), my program attaches that report at the top of the email. I am looking for similar functionality from rspamc. i.e. "rspamc -R" option.

## Motivation
I am looking to migrate from spamassassin to rspamd but I have come across a blocker due to missing feature.

As per **Summary** given above, when a message is identified as SPAM, my company prefers that SPAM report should also be attached at the top of the e-mail so that recipient knows the reason why the message was identified as SPAM.

Spamassassin already has this feature and report can also be customized to some extent. For example:

```
clear_report_template
report Content preview:  _PREVIEW_
report
report Content analysis details:   (_HITS_ points, _REQD_ required)
report
report " pts rule name              description"
report  ---- ---------------------- --------------------------------------------------
report _SUMMARY_
```

This gives a report like this: (with spamc -R)

```
6.5/5.0
Content preview:  This email must be viewed in HTML mode. Forget about the COVID
   vaccine, this is the most important discovery of this century... While greedy
   Big Pharma is trying to rapidly push their side effect filled vaccines through
   the door, with the help of world???s elites and top execs...

Content analysis details:   (6.5 points, 5.0 required)

 pts rule name              description
---- ---------------------- --------------------------------------------------
 1.2 URIBL_ABUSE_SURBL      Contains an URL listed in the ABUSE SURBL blocklist
                            [URI: wiseedit.today]
 0.0 SPF_HELO_NONE          SPF: HELO does not publish an SPF Record
-0.0 SPF_PASS               SPF: sender matches SPF record
-0.1 DKIM_VALID             Message has at least one valid DKIM or DK signature
-0.1 DKIM_VALID_AU          Message has a valid DKIM or DK signature from author's
                            domain
 0.1 DKIM_SIGNED            Message has a DKIM or DK signature, not necessarily valid
 0.8 BAYES_50               BODY: Bayes spam probability is 40 to 60%
                            [score: 0.5000]
 0.0 FREEMAIL_FROM          Sender email is commonly abused enduser mail provider
                            [nakamoa436(at)gmail.com]
(report trimmed)
```

[...]

PR #4377, opened 2023-01-20
rspamc: add -R option for human readable report

amishmm:
Spamassassin has a nice feature of printing human readable report. (see spamc -R option)

I had created a feature request for the same which describes the requirement in detail. The feature request can be read here #4373 

This pull request (PR) tries to emulate "spamc -R" option of spamassassin.

First line will print score/threshold,Action=action,Spam=bool and then followed by human readable report.

Sample report is as follows:

```
10.80/15.00,Action=add header,Spam=true
Content analysis details:   (10.80 points, 15.00 required)

 pts rule name              description
---- ---------------------- --------------------------------------------------
 0.0 ARC_NA                 ARC signature absent
 0.0 ASN                    [asn:34245, ipnet:212.17.32.0/19, country:IE]
 0.1 DMARC_POLICY_SOFTFAIL  DMARC failed [buap.mx : No valid SPF, No valid DKI
                            M,none]
 0.0 FROM_NEQ_ENVFROM       From address is different to the envelope [xl6Ety0
                            0V@fismat1.fcfm.buap.mx,dev_null_sample_spam@examp
                            le.com]
 0.0 FROM_NO_DN             From header does not have a display name
 1.5 INVALID_DATE           Malformed date header
 1.7 INVALID_MSGID          Message id is incorrect
-0.1 MIME_GOOD              Known content-type [text/plain]
 0.0 MIME_TRACE             [0:+]
 0.0 RCPT_COUNT_ONE         One recipient [1]
 0.0 RCVD_COUNT_FIVE        Message has 5-7 Received headers [6]
 0.1 RCVD_NO_TLS_LAST       Last hop did not use encrypted transports
 0.0 R_DKIM_NA              Missing DKIM signature
 1.0 R_SPF_FAIL             SPF verification failed [-all:c]
 3.0 R_UNDISC_RCPT          Recipients are absent or undisclosed
 0.0 TO_DN_ALL              All the recipients have display names
 3.5 VIOLATED_DIRECT_SPF    Has no Received (or no trusted received relays) an
                            d SPF policy fails or soft fails

Emails: ["maillistdrop@post.com"]

```

This is more or less same output that spamc of spamassassin gives with its -R option.

I have done some more additions to it, like printing of URLS and Emails (if available) at the end. Which can act as a preview for links embedded in the email.

I have tried to do minimal changes to the existing "rspamc_symbols output()" function. Such that it does not change the original flow of the code.

Please review the PR.

Thank you

### DIFF
--- a/src/client/rspamc.cxx
+++ b/src/client/rspamc.cxx
@@ -71,6 +71,7 @@ static gboolean pass_all;
 static gboolean tty = FALSE;
 static gboolean verbose = FALSE;
 static gboolean print_commands = FALSE;
+static gboolean humanreport = FALSE;
 static gboolean json = FALSE;
 static gboolean compact = FALSE;
 static gboolean headers = FALSE;
@@ -134,6 +135,7 @@ static GOptionEntry entries[] =
 																																	  "Bind to specified ip address",                                             nullptr},
 				{"commands",         0,    0,                          G_OPTION_ARG_NONE,         &print_commands,
 																																	  "List available commands",                                                  nullptr},
+				{"human",            'R',  0,                          G_OPTION_ARG_NONE,         &humanreport,                       "Output human readable report",                                             nullptr},
 				{"json",             'j',  0,                          G_OPTION_ARG_NONE,         &json,                              "Output json reply",                                                        nullptr},
 				{"compact",          '\0', 0,                          G_OPTION_ARG_NONE,         &compact,                           "Output compact json reply",                                                nullptr},
 				{"headers",          0,    0,                          G_OPTION_ARG_NONE,         &headers,                           "Output HTTP headers",
@@ -824,6 +826,59 @@ add_options(GQueue *opts)
 }
 
 static void
+print_indented_line(FILE *out, std::string line, size_t maxlen, size_t indent)
+{
+	if (maxlen < 1) return;
+
+	std::string s;
+	for (size_t pos = 0; pos < line.length(); pos += s.length()) {
+		s = line.substr(pos, pos ? (maxlen-indent) : maxlen);
+		if (indent && pos) fmt::print(out, "{:>{}}", " ", indent);
+		fmt::print(out, "{}\n", s);
+	}
+}
+
+static void
+rspamc_symbol_human_output(FILE *out, const ucl_object_t *obj)
+{
+	auto first = true;
+	double score = 0;
+	const char *key = nullptr, *desc = nullptr;
+
+	const auto *val = ucl_object_lookup(obj, "score");
+	if (val != nullptr) score = ucl_object_todouble(val);
+
+	key = ucl_object_key(obj);
+	val = ucl_object_lookup(obj, "description");
+	if (val != nullptr) desc = ucl_object_tostring(val);
+
+	std::string line = fmt::format("{:>4.1f} {:<22} ", score, key);
+	if (desc != nullptr) line += desc;
+
+	val = ucl_object_lookup(obj, "options");
+	if (val != nullptr && val->type == UCL_ARRAY) {
+		ucl_object_iter_t it = nullptr;
+		const ucl_object_t *cur;
+
+		line += fmt::format("{}[", desc == nullptr ? "" : " ");
+
+		while ((cur = ucl_object_iterate (val, &it, true)) != nullptr) {
+			if (first) {
+				line += fmt::format("{}", ucl_object_tostring(cur));
+				first = false;
+			}
+			else {
+				line += fmt::format(",{}", ucl_object_tostring(cur));
+			}
+		}
+		line += ']';
+	}
+	else if (desc == nullptr) line += '\n';
+
+	print_indented_line(out, line, 78, 28);
+}
+
+static void
 rspamc_symbol_output(FILE *out, const ucl_object_t *obj)
 {
 	auto first = true;
@@ -864,12 +919,12 @@ rspamc_metric_output(FILE *out, const ucl_object_t *obj)
 	auto print_protocol_string = [&](const char *ucl_name, const char *output_message) {
 		auto *elt = ucl_object_lookup(obj, ucl_name);
 		if (elt) {
-			fmt::print(out, "{}: {}\n", output_message,
+			fmt::print(out, fmt::runtime(humanreport ? ",{}={}" : "{}: {}\n"), output_message,
 					emphasis_argument(ucl_object_tostring(elt)));
 		}
 	};
 
-	fmt::print(out, "[Metric: default]\n");
+	if (!humanreport) fmt::print(out, "[Metric: default]\n");
 
 	const auto *elt = ucl_object_lookup(obj, "required_score");
 
@@ -883,6 +938,13 @@ rspamc_metric_output(FILE *out, const ucl_object_t *obj)
 	if (elt) {
 		score = ucl_object_todouble(elt);
 		got_scores++;
+	}
+
+	if (humanreport) {
+		fmt::print(out,
+				"{}/{}",
+				emphasis_argument(score, 2),
+				emphasis_argument(required_score, 2));
 	}
 
 	elt = ucl_object_lookup(obj, "action");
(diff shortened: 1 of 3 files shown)


vstakhov:
It looks veru good in general, thank you! I will review details a bit later.

src/client/rspamc.cxx
```diff
@@ -823,6 +825,59 @@ add_options(GQueue *opts)
...
 
+static void
+print_indented_line(FILE *out, std::string line, size_t maxlen, size_t indent)
+{
+	if (maxlen < 1) return;
+
+	std::string s;
+	for (size_t pos = 0; pos < line.length(); pos += s.length()) {
+		s = line.substr(pos, pos ? (maxlen-indent) : maxlen);
+		if (indent && pos) fmt::print(out, "{:>{}}", " ", indent);
```
vstakhov, line 836:
```suggestion
	for (size_t pos = 0; pos < line.length(); pos += s.length()) {
		auto s = std::string_view(line).substr(pos, pos ? (maxlen-indent) : maxlen);
		if (indent && pos) {
			fmt::print(out, "{:>{}}", " ", indent);
		}
```

src/client/rspamc.cxx
```diff
@@ -823,6 +825,59 @@ add_options(GQueue *opts)
 	}
 }
 
+static void
+print_indented_line(FILE *out, std::string line, size_t maxlen, size_t indent)
+{
+	if (maxlen < 1) return;
```
vstakhov, line 831:
```suggestion
	if (maxlen < 1) {
		return;
	}
```

src/client/rspamc.cxx
```diff
@@ -823,6 +825,59 @@ add_options(GQueue *opts)
...
+		if (indent && pos) fmt::print(out, "{:>{}}", " ", indent);
+		fmt::print(out, "{}\n", s);
+	}
+}
+
+static void
+rspamc_symbol_human_output(FILE *out, const ucl_object_t *obj)
+{
+	auto first = true;
+	double score = 0;
```
vstakhov, line 845:
```suggestion
	auto score = 0.0;
```

src/client/rspamc.cxx
```diff
@@ -823,6 +825,59 @@ add_options(GQueue *opts)
...
+	const char *key = nullptr, *desc = nullptr;
+
+	const auto *val = ucl_object_lookup(obj, "score");
+	if (val != nullptr) score = ucl_object_todouble(val);
+
+	key = ucl_object_key(obj);
```
vstakhov, line 851:
```suggestion
	const auto *key = ucl_object_key(obj);
```

src/client/rspamc.cxx
```diff
@@ -823,6 +825,59 @@ add_options(GQueue *opts)
...
+		fmt::print(out, "{}\n", s);
+	}
+}
+
+static void
+rspamc_symbol_human_output(FILE *out, const ucl_object_t *obj)
+{
+	auto first = true;
+	double score = 0;
+	const char *key = nullptr, *desc = nullptr;
```
vstakhov, line 846:
```suggestion
	const char *desc = nullptr;
```

src/client/rspamc.cxx
```diff
@@ -823,6 +825,59 @@ add_options(GQueue *opts)
...
+
+	const auto *val = ucl_object_lookup(obj, "score");
+	if (val != nullptr) score = ucl_object_todouble(val);
+
+	key = ucl_object_key(obj);
+	val = ucl_object_lookup(obj, "description");
+	if (val != nullptr) desc = ucl_object_tostring(val);
+
+	std::string line = fmt::format("{:>4.1f} {:<22} ", score, key);
```
vstakhov, line 855:
```suggestion
	auto line = fmt::format("{:>4.1f} {:<22} ", score, key);
```

src/client/rspamc.cxx
```diff
@@ -823,6 +825,59 @@ add_options(GQueue *opts)
...
+	if (desc != nullptr) line += desc;
+
+	val = ucl_object_lookup(obj, "options");
+	if (val != nullptr && val->type == UCL_ARRAY) {
```
vstakhov, line 859:
```suggestion
	if (val != nullptr && ucl_object_type(val) == UCL_ARRAY) {
```

src/client/rspamc.cxx
```diff
@@ -864,12 +919,12 @@ rspamc_metric_output(FILE *out, const ucl_object_t *obj)
 	auto print_protocol_string = [&](const char *ucl_name, const char *output_message) {
 		auto *elt = ucl_object_lookup(obj, ucl_name);
 		if (elt) {
-			fmt::print(out, "{}: {}\n", output_message,
+			fmt::print(out, fmt::runtime(humanreport ? ",{}={}" : "{}: {}\n"), output_message,
 					emphasis_argument(ucl_object_tostring(elt)));
```
vstakhov, line 923:
```suggestion
			if (humanreport) {
				fmt::print(out, {}={}", output_message, emphasis_argument(ucl_object_tostring(elt)));
			}
			else {
				fmt::print(out, "{}: {}\n", output_message,
					emphasis_argument(ucl_object_tostring(elt)));
			}
```

src/client/rspamc.cxx
```diff
@@ -1003,7 +1079,7 @@ rspamc_symbols_output(FILE *out, ucl_object_t *obj)
 			emitted = (char *)ucl_object_emit(elt, UCL_EMIT_JSON);
 		}
 
-		fmt::print(out, "Urls: {}\n", emitted);
+		if (emitted && strcmp(emitted, "[]")) fmt::print(out, "Urls: {}\n", emitted);
```
vstakhov, line 1082: This will change the existing behaviour which is not a good thing. I'd suggest to use this logic merely for reports.

src/client/rspamc.cxx
```diff
@@ -915,26 +977,36 @@ rspamc_metric_output(FILE *out, const ucl_object_t *obj)
 					colorized_action = fmt::format(fmt::emphasis::bold, ucl_object_tostring(elt));
 					break;
 				}
-				fmt::print(out, "Action: {}\n", colorized_action);
+				fmt::print(out, fmt::runtime(humanreport ? ",Action={}" : "Action: {}\n"), colorized_action);
 			}
 
-			fmt::print(out, "Spam: {}\n", emphasis_argument(act.value() < METRIC_ACTION_GREYLIST ?
-											  "true" : "false"));
+			fmt::print(out, fmt::runtime(humanreport ? ",Spam={}" : "Spam: {}\n"),
```
vstakhov, line 983: I'd suggest to convert `fmt::runtime` to a simple `if...else` to keep fmt constexpr providing better compile time checks.

src/client/rspamc.cxx
```diff
@@ -915,26 +977,36 @@ rspamc_metric_output(FILE *out, const ucl_object_t *obj)
 					colorized_action = fmt::format(fmt::emphasis::bold, ucl_object_tostring(elt));
 					break;
 				}
-				fmt::print(out, "Action: {}\n", colorized_action);
+				fmt::print(out, fmt::runtime(humanreport ? ",Action={}" : "Action: {}\n"), colorized_action);
```
vstakhov, line 980: Same, no `fmt::runtime` please.

src/client/rspamc.cxx
```diff
@@ -1018,11 +1094,12 @@ rspamc_symbols_output(FILE *out, ucl_object_t *obj)
 			emitted = (char *)ucl_object_emit(elt, UCL_EMIT_JSON);
 		}
 
-		fmt::print(out, "Emails: {}\n", emitted);
+		if (emitted && strcmp(emitted, "[]")) fmt::print(out, "Emails: {}\n", emitted);
```
vstakhov, line 1097: Same here, please let it be `humanreport` specific behaviour.

vstakhov (approved):
Few nits added.

src/client/rspamc.cxx
amishmm: I did this intentionally. But noted and will change it back. Thanks.

src/client/rspamc.cxx
amishmm: Can't define "s" as "auto" inside the "for" loop as otherwise "s.length()" becomes outside the scope and compiler throws the error.

src/client/rspamc.cxx
amishmm: This line was copied from rspamc_symbol_output() - so I have made above changes to this line as well as the line which was copied.

amishmm:
I have updated the PR, implementing the suggestions given by reviewer @vstakhov .

There are new additions as follow:

1. The first line of the output contains the message score and three threshold scores, in this format:<br>
score/greylist/addheader/reject,action=N:ACTION,spam=0|1,skipped=0|1<br>
This is done because rspamd is bit different from spamassassin and uses 3 types of score thresholds.<br>
This allows to generate statistics of messages greylisted or detected as spam or rejected based on the score alone without using action or spam fields of the first line.<br>
action=N:ACTION, here N is the number indicating the action type. ACTION is the textual desciption of the same.<br>
NOTE: Currently rspamd does not provide "greylist_score" and "addheader_score" in JSON / UCL reply. But I have added the same in rspamc code.<br>
When I have time, I will try to look at the rspamd code and implement those scores.

2. Emails and Urls line now respect 78 character limit (for human report) 

Please review the code again.

Thank you

amishmm:
Thank you very much for the merge.

Just a suggestion, its better to squash and merge as a single commit so that the "trial / incomplete / test" commits don't go in master branch but only the final approved code goes to master branch.

PR #4378, opened 2023-01-22
rspamc: rewrite wrapping logic

amishmm:
Hello

Current wrapping logic in print_indented_line() is complicated and has bugs. (as reported [here](https://github.com/rspamd/rspamd/commit/6fea5899c693bb934ba7a9263f602d32baa76fc7#commitcomment-97642367))

This pull request rewrites wrapping logic completely.

1. It is much faster and does not do long substr() based calculations.
2. Does not run for loop inside for loop
3. Treats opening brackets differently from closing bracket (which makes report more human readable)
4. Skips leading white space for indented line.

Please check

Thank you.

### DIFF
--- a/src/client/rspamc.cxx
+++ b/src/client/rspamc.cxx
@@ -871,38 +871,33 @@ rspamc_symbol_human_output(FILE *out, const ucl_object_t *obj)
 	}
 
 	auto print_indented_line = [&](size_t maxlen, size_t indent) {
-		if (maxlen < 1 || maxlen < indent) {
+		if (maxlen < 1 || maxlen <= indent) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


vstakhov:
Uh-oh, I have also decided to do some adjustments to the algorithm.

vstakhov:
I'll probably try your algorithm as it looks cleaner than mine

vstakhov:
I have used your ideas, thank you!

vstakhov:
Oh, and I have also excluded `.` from delimiters as it breaks domains.

vstakhov:
There is also one minor issue with this function when an empty line is passed. I will check and fix.

PR #4380, opened 2023-01-23
add greylist_score and addheader_score to scan result

amishmm:
"rspamc" (human report option) needs greylist and add header thresholds.

This helps a calling program (program that calls rspamc) to decide "own" custom actions OR create on custom report.

This PR adds greylist_score and addheader_score to scan result.

### DIFF
--- a/src/client/rspamc.cxx
+++ b/src/client/rspamc.cxx
@@ -969,13 +969,11 @@ rspamc_metric_output(FILE *out, const ucl_object_t *obj)
 		got_scores++;
 	}
 
-	/* XXX: greylist_score is not yet in checkv2 */
 	elt = ucl_object_lookup(obj, "greylist_score");
 	if (elt) {
 		greylist_score = ucl_object_todouble(elt);
 	}
 
-	/* XXX: addheader_score is not yet in checkv2 */
 	elt = ucl_object_lookup(obj, "addheader_score");
 	if (elt) {
 		addheader_score = ucl_object_todouble(elt);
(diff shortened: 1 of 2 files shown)


vstakhov:
Ah, one last thing: action thresholds might be modified for a specific task by multiple plugins, so it is better to get the real threshold instead one from `cfg`

vstakhov:
I think that this approach will be more flexible in general.

amishmm:
Thank you very much for implementing this.

Just wanted to be to clear if "required_score" is guaranteed to be same as "thresholds -> reject" in all cases?

Then in protocol.c "required_score" may be deprecated (and removed in future version) in favor of "thresholds -> reject". rspamc should also use thresholds -> reject for setting required_score.

And "required_score" may be renamed to "reject_score", for more appropriate name which matches with name in config file.

vstakhov:
> Just wanted to be to clear if "required_score" in JSON is guaranteed to be same as "thresholds -> reject" in all cases?

It should be to my best knowledge.

> Then in protocol.c "required_score" may be deprecated (and removed in future version) in favor of "thresholds -> reject".

Deprecation is always an issue. We really don't know who and how are using these fields. So it is better to keep them to avoid any breaking. Another story is when you just cannot expand protocol or do something cool with the current schema. Then, it is still better to add another endpoint (that's how `checkv2` has been added).

amishmm:
Thank you for your reply.

So just in case there is a chance that both can be different, which one best represents the score (for that email)?

"required_score" OR "thresholds -> reject"?

It "thresholds -> reject" then "rspamc" should use that.

PR #4383, opened 2023-01-26
Set required_score = reject score and minor change to the report

amishmm:
In human report (rspamc -R option) when message is detected as SPAM, it means that message score is greater than or equal to the **add header** threshold but it may not  be greater than or equal to **reject score**.

For example we get output like this in human report:
Content analysis details:   (6.62 points, 15.00 required)

This creates a confusion that message score was low (6.62 is less than 15) and yet it was detected as SPAM. But actually message was detected as SPAM because its score was more than 6 (the add header score).

So this PR fixes this in human report. i.e. it prints add header score as required score instead of reject score as required score.

New output will be like this:
Content analysis details:   (6.62 points, 6.00 required)

Additionally, after a new "thresholds" object was introduced recently in JSON output. If thresholds -> reject score is available then this PR (re-) sets "required_score" to "reject score". Because that seems more appropriate.

### DIFF
--- a/src/client/rspamc.cxx
+++ b/src/client/rspamc.cxx
@@ -971,14 +971,19 @@ rspamc_metric_output(FILE *out, const ucl_object_t *obj)
 	const auto *thresholds_obj = ucl_object_lookup(obj, "thresholds");
 
 	if (thresholds_obj && ucl_object_type(thresholds_obj) == UCL_OBJECT) {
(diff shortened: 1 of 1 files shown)


vstakhov:
We should also think about the possibilities when some other action, such as `rewrite subject`, takes place, but it is a good initial approach, I agree, merging, thanks!

PR #4387, opened 2023-01-31
conf/modules.d/arc.conf: fix parameter name

jendis:
rename symbol_sign -> sign_symbol


vstakhov:
Good catch, thank you!

PR #4399, opened 2023-02-17
[Enhancement] Make Google URL redirection rules productive

twesterhever:
Both rules frequently trigger in spam messages, and seem to do their job quite well, without causing any harm to legitimate mails. Therefore, this patch suggests to raise their score to 1.0 each, and retag them to the `url` group rather than `experimental`.

While the `googleusercontent.com` rule is complete, the `HAS_GOOGLE_REDIR` one could be bypassed by using a different Google domain, e.g. google.at - is there an easy way to write a rule triggering on any `google.[TLD]` FQDN? If so, that might be a potential enhancement as well.


vstakhov:
I do agree, thanks for the PR!

PR #4416, opened 2023-03-01
Fix for x-rspamd-action in Milter headers module

nishils:
When metric action is not defined, it reads a score causing a type error where the value should be a string and the actual action.

As a side note, the documentation for this module seems fairly out of date.


vstakhov:
That's a good catch, thank you!

PR #4430, opened 2023-03-11
[Minor] remove check for sys/cdefs.h in CMakelists.txt

a16bitsysop:
rspamd uses G_(BEGIN|END)_DECLS not __(BEGIN|END)_DECLS


vstakhov:
`cdefs` include is quite useful for *BSD systems where it defines some essential macro for other includes. On the other hand, it is normally included by other system files. Have you tested BSD build without this check?

a16bitsysop:
No I haven’t will give it a go

a16bitsysop:
I tried on freeBSD with luajit-openresty and it compiles and passes all the tests with this change

PR #4448, opened 2023-03-29
fix: avoid double-dip of nrcpt when calculating ratelimit condition

simbuerg:
Before this commit, rspamd would consider the number of recipients of the current message twice when calculating the ratelimit condition.


vstakhov:
That's a good catch - we aready have this value in the `pending` field. Thank you!

PR #4449, opened 2023-03-30
Fix arc

jendis:
This PR fixes ARC Sealing functionality, more specifically the computation of hash of an ARC-Seal header, with respect to the https://www.rfc-editor.org/rfc/rfc8617.html#section-5.1.1

### DIFF
--- a/src/plugins/lua/arc.lua
+++ b/src/plugins/lua/arc.lua
@@ -530,6 +530,8 @@ local function arc_sign_seal(task, params, header)
   local cur_idx = 1
   if arc_seals then
     cur_idx = #arc_seals + 1
+    -- We use the cached version per each ARC-* header field individually, already sorted by instance
+    -- value in ascending order
     for i = 1, #arc_seals, 1 do
       if arc_auth_results[i] then
         local s = dkim_canonicalize('ARC-Authentication-Results',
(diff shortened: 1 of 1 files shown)


vstakhov (approved):
It seems that our arc signing is definitely broken for i > 1, which is bad news... Thank you for a fix, we really need to have some interoperation checks with other implementation.

vstakhov:
The question is to find any suitable ARC implementation to check the cross compatibility. For now Rspamd can check signatures from itself but this might be broken for signatures with `i>1`...

jendis:
So I've done several tests where the ARC validator was Microsoft. I set three ARC sealers in my environment where the last node (i=3) was rspamd where I used fixed and old broken `arc.lua`.

The result was as expected, in case of sealing with old broken `arc.lua` the next Sealer on MS added header
`ARC-Seal: i=4; a=rsa-sha256; s=arcselector9901; d=microsoft.com; cv=pass;`
In case of fixed `arc.lua`:
`ARC-Seal: i=4; a=rsa-sha256; s=arcselector9901; d=microsoft.com; cv=fail;`

So I consider this fix to be sufficient.

vstakhov:
Thank you very much, it is very valuable fix indeed!

PR #4497, opened 2023-05-26
[Enhancement] Improve detection of Google redirection URLs

twesterhever:
The list is derived from Firefox' static HPKP entires, retrieved from [this location](https://searchfox.org/mozilla-central/source/security/manager/ssl/StaticHPKPins.h). While the regular expression is rather complex to read, I lack ideas on how to improve its readability - suggestions welcome!

### DIFF
--- a/rules/regexp/headers.lua
+++ b/rules/regexp/headers.lua
@@ -915,8 +915,8 @@ reconf['HAS_GUC_PROXY_URI'] = {
 }
 
 reconf['HAS_GOOGLE_REDIR'] = {
-  re = '/\\.google\\.com\\/url\\?/{url}i',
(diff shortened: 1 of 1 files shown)


vstakhov:
We should probably simplify this regexp to something like `[a-z]{2,3}(.[a-z]{2,3})?`

twesterhever:
> We should probably simplify this regexp to something like `[a-z]{2,3}(.[a-z]{2,3})?`

Yes, I guess that makes sense at this stage. Will provide a commit accordingly.

twesterhever:
Here you are, apologies for the delay. Had to adjust it a little bit to ensure `google.info` and `google.jobs` are also covered.

PR #4499, opened 2023-05-26
[WebUI] Add ability to compute fuzzy hashes

moisseev:
from sample message source


vstakhov:
That's great, thank you!

PR #4518, opened 2023-06-20
[Minor] Increase score for .chm attachments

twesterhever:
Please see https://asec.ahnlab.com/ko/53426/ for the rationale of this measure.

On a general note, would it be desirable to have all file extensions already blocked by GMail and Outlook set to "4" instead of "2"? If so, happy to provide a patch for that.


vstakhov:
> On a general note, would it be desirable to have all file extensions already blocked by GMail and Outlook set to "4" instead of "2"?

Sounds like a good idea!

PR #4519, opened 2023-06-22
[Enhancement] Treat HTML attachments as (slightly) bad

twesterhever:
As an attempt to improve `rspamd`'s detection rates on malicious e-mails leveraging "HTML smuggling", this merge request proposes to treat HTML attachments as (slightly) bad, adding +2 points to a message if an attachment having such a filename suffix is found.

In contrast to HTML MIME parts, there is little legitimate reason for such attachments. To avoid collateral damage, HTML files in archives are not covered in this merge request.

See also:
- https://www.malwarebytes.com/blog/news/2021/11/evasive-maneuvers-html-smuggling-explained
- https://www.mcafee.com/blogs/other-blogs/mcafee-labs/shtml-phishing-attack-with-blurred-image/


src/plugins/lua/mime_types.lua
```diff
@@ -45,8 +45,10 @@ local settings = {
   extension_map = { -- extension -> mime_type
     html = 'text/html',
     htm = 'text/html',
-    txt = 'text/plain',
```
moisseev, line 48:
```diff
   extension_map = { -- extension -> mime_type
     html = 'text/html',
     htm = 'text/html',
-    pdf = 'application/pdf'
+    pdf = 'application/pdf',
     shtm = 'text/html',
     shtml = 'text/html',
-    txt = 'text/plain',
+    txt = 'text/plain'
   },

   bad_extensions = {
```

twesterhever: Gak, thanks for spotting that! Fixed.

PR #4537, opened 2023-07-11
[Feature] external_relay: add ip_map strategy

moesoha:
Add a strategy named `ip_map` for finding relays by IP/network map.

Example config:
```
rules {
  EXTERNAL_RELAY_HOST {
    strategy = "ip_map";
    ip_map = [
      "2001:db8::/32",
      "192.0.2.0/24"
    ];
  }
}
```

This PR should not be squashed as it contains commits for two goals.


fatalbanana:
For testing it may be useful to set a lower priority (like 9?) on all rules in external_relay configuration; probably these tests are just barely hanging together as it is- sorry about it.

In general this LGTM.

src/plugins/lua/external_relay.lua
```diff
@@ -182,6 +188,42 @@ strategies.hostname_map = function(rule)
...
+    for i, rcvd in ipairs(rcvd_hdrs) do
+      if rcvd.real_ip then
+        local rcvd_ip = rcvd.real_ip
+        if rcvd_ip:is_valid() and (not ip_map:get_key(rcvd_ip) or i == num_rcvd) then
+          return set_from_rcvd(task, rcvd)
+        end
+      end
+    end
+
+    rspamd_logger.errx(task, 'found nothing useful in Received headers')
```
vstakhov, line 223: Should it be a error notice? I think it should be either debug or info maximum. Users generally don't like too verbose logs.

src/plugins/lua/external_relay.lua
```diff
@@ -182,6 +188,42 @@ strategies.hostname_map = function(rule)
...
+      return
+    end
+
+    local rcvd_hdrs = task:get_received_headers()
+    local num_rcvd = #rcvd_hdrs
+    -- Try find sending IP in Received headers
+    for i, rcvd in ipairs(rcvd_hdrs) do
+      if rcvd.real_ip then
+        local rcvd_ip = rcvd.real_ip
+        if rcvd_ip:is_valid() and (not ip_map:get_key(rcvd_ip) or i == num_rcvd) then
```
vstakhov, line 217: This logic is a bit unclear. What is the overall goal of the check `i == num_rcvd`? It seems that it might break the exisiting logic of passing smtp envelope data to Rspamd.

vstakhov (commented):
Looks good in general with some minor nits. Thank you!

moesoha:
I added a lower priority to EXTERNAL_RELAY_COUNT to fix the test cases. This is in a separate commit so this PR should not be squashed.

src/plugins/lua/external_relay.lua
moesoha: everywhere in this module uses `errx` for logging such message. Of course I can change all such conditions if you want.

src/plugins/lua/external_relay.lua
moesoha: I am wondering if a message is originated from a relay host, the real Received header decision will be wrong if we remove this.

PR #4538, opened 2023-07-11
[Feature] Detect Apple Mail and remove scores for their known common behavior

dragoangel:
Both Apple Mail and Apple iOS Mail always use `Mime-Version` instead of `MIME-Version` leading to `MV_CASE`, so it need to be excluded for them. Specifically Apple iOS Mail not add `plain-text` version of email leading to `MIME_MA_MISSING_TEXT` which need to be excluded as well. 

Note about `plain-text`: it will be added by Apple iOS Mail __only__ on replies, in the case when the email to which the reply is made has `plain-text` part. As such behavior can't be captured, we just can drop  `MIME_MA_MISSING_TEXT` generally for Apple iOS Mail

### DIFF
--- a/rules/regexp/headers.lua
+++ b/rules/regexp/headers.lua
@@ -975,7 +975,7 @@ reconf['APPLE_MAILER'] = {
 -- Detect Apple iPhone/iPad Mail
 -- Apple iPhone/iPad Mail X-Mailer contains iOS build number, e. g. 9B206, 16H5, 18G5023c
 -- https://en.wikipedia.org/wiki/IOS_version_history
-local apple_ios_x_mailer = [[i(?:Phone|Pad) Mail \((?:1[AC]|[34][AB]|5[ABCFGH]|7[A-E]|8[ABCEFGHJKL]|9[AB]|\d{2}[A-Z])\d+[a-z]?\)]]
+local apple_ios_x_mailer = [[i(?:Phone|Pad) Mail \(\d{1,2}[A-Z]\d{1,4}[a-z]?\)]]
 reconf['APPLE_IOS_MAILER'] = {
   description = 'Sent with Apple iPhone/iPad Mail',
   re = string.format('X-Mailer=/^%s/{header}', apple_ios_x_mailer),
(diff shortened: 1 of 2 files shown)


rules/regexp/headers.lua
```diff
@@ -963,6 +963,26 @@ reconf['OLD_X_MAILER'] = {
...
+  description = 'Sent with Apple Mail',
+  re = string.format('X-Mailer=/^%s/{header}', apple_x_mailer),
+  score = 0.0,
+  group = 'headers',
+}
+
+-- Detect Apple iPhone/iPad Mail
+-- Apple iPhone/iPad Mail X-Mailer contains iOS build number, e. g. 9B206, 16H5, 18G5023c
+-- https://en.wikipedia.org/wiki/IOS_version_history
+local apple_ios_x_mailer = [[i(?:Phone|Pad) Mail \((?:1[AC]|[34][AB]|5[ABCFGH]|7[A-E]|8[ABCEFGHJKL]|9[AB]|\d{2}[A-Z])\d+[a-z]?\)]]
```
vstakhov, line 978: Can we constrain the last `\d+` to avoid slow performance?

vstakhov (approved):
LGTM with a minor nit

rules/regexp/headers.lua
dragoangel: It's an existing regex btw, I just moved it to top to be near detect rule

vstakhov: I know, but it doesn't mean it that this RE is good :) If we look for a version it is definitely constrained by 1-5 digits (`{1,5}`)

dragoangel: its a mix of digits and uppercase letters :)

dragoangel: it is to cover this (18G5023c) version for example. If we simplify it - the quality of `FORGED_X_MAILER` will suffer - it is a regular from it: 
https://github.com/rspamd/rspamd/blob/0729c58cb03befdb31b7a1ca56bd052abfd66477/rules/regexp/headers.lua#L1002-L1008

We can leave the complex regex in `FORGED_X_MAILER`, and insert this symbol using a simple regex, only if there no `FORGED_X_MAILER`?

dragoangel: I updated regex, please check

PR #4541, opened 2023-07-15
replace ECDSA_sign_setup/ECDSA_sign_ex with ECDSA_sign

sthen:
There's no need to use ECDSA_sign_setup/ECDSA_sign_ex separately, a single call to ECDSA_sign will handle everything that's needed here.

Avoids breakage with a soon-to-be-committed change to LibreSSL to remove ECDSA_sign_setup/ex from libcrypto.

Diff from @botovq


vstakhov:
Sadly, `ECDSA_sign` is also deprecated by mad changes in OpenSSL 3 API. However, I think this change is still good anyway.

PR #4542, opened 2023-07-15
[Feature] Use PTR records in hfilter

bobobo1618:
My MTA (Axigen) doesn't pass the hostname in its Milter `connect()` call, which leads to consistent `HFILTER_HOSTNAME_UNKNOWN`.

This change updates hfilter so that when there is no hostname but there is an IP, it will do the PTR lookup itself.

### DIFF
--- a/src/plugins/lua/hfilter.lua
+++ b/src/plugins/lua/hfilter.lua
@@ -429,17 +429,23 @@ local function hfilter_callback(task)
     end
   end
 
-
-  local function hostname_cb()
+  local function hostname_processed_cb()
     local weight_hostname = 0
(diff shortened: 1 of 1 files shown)


fatalbanana:
This would better go in a prefilter which could be conditionally enabled by those who might need it. Such a prefilter should perform forward-confirmation of the hostname.

bobobo1618:
> This would better go in a prefilter which could be conditionally enabled by those who might need it. Such a prefilter should perform forward-confirmation of the hostname.

Very good suggestion, hadn't thought of that. Closing this in favour of https://github.com/rspamd/rspamd/pull/4543.

PR #4543, opened 2023-07-15
Add a new module that ensures a hostname is present

bobobo1618:
Follow-up from #4542.

My MTA (Axigen) doesn't pass the hostname in its Milter connect() call, which leads to consistent HFILTER_HOSTNAME_UNKNOWN.

This change adds a new pre-filter module that can add a missing hostname based on a PTR lookup.

### DIFF
--- a/conf/modules.d/ensure_hostname.conf
+++ b/conf/modules.d/ensure_hostname.conf
@@ -1,0 +1,19 @@
+# Please don't modify this file as your changes might be overwritten with
+# the next update.
+#
+# You can modify 'local.d/ensure_hostname.conf' to add and merge
+# parameters defined inside this section
+#
+# You can modify 'override.d/ensure_hostname.conf' to strictly override all
+# parameters defined inside this section
+#
+# See https://rspamd.com/doc/faq.html#what-are-the-locald-and-overrided-directories
+# for details
+#
+# Module documentation can be found at  https://rspamd.com/doc/modules/ensure_hostname.html
+
+
+ensure_hostname {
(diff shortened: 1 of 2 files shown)


vstakhov:
I have checked the original [ticket](https://www.axigen.com/community/t/does-axigen-pass-hostnames-to-milter-filters/3482) and I think that this change is not needed at all. Rspamd indeed, tries to set the hostname from the milter connect stage. However, if there is no information about hostname at the connect stage Rspamd tries to get that from the `{client_name}` macro that must be announced by MTA. So I think that the problem is clearly on the MTA side: it specifiesthe IP address as a hostname, that leads to various issues. Hence, I think it is better to fix the MTA behaviour than to add this plugin as a patchwork.

bobobo1618:
Ioan at Axigen responded on that thread and seems to disagree. His main sticking point seems to be:

> As mentioned before, we consider that the milter client should not be enforced to make any extra DNS checks and present that information to the milter server.

Would you be up for discussing with him directly there? I'm happy to keep acting as a proxy if you don't want to make an account but it might be more efficient.

But I totally understand if you don't think it's worth your time. It looks to me like they consider this too much work for them so they're trying to push it on the milters.

For my part, this custom module seems to work but I may be better off just switching to Postfix/Dovecot like everyone else, as they play nicely with the open-source ecosystem,.

indreias:
Actually I could comment here as well.
What I am saying is that I need some arguments to put on our Dev team regrading why our current implementation is not working nice.

Issue #4318, opened 2022-10-25
[BUG] history_redis stopped working w/password if specified in history_redis.conf


Player701:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
This is my configuration for the redis history module (```/etc/rspamd/local.d/history_redis.conf```):

```
servers = "127.0.0.1";
password = "<password hidden>";
key_prefix = "rs_history";
nrows = 1000;
compress = true;
subject_privacy = true;
```

Used to work perfectly until a few weeks ago, I suppose that's when rspamd was updated to 3.3. Since then, I've been finding messages of the following form in the log:

```Oct 25 17:26:39 mail.mydomain.com rspamd[2831172]: <4niw6e>; lua; history_redis.lua:277: history_redis: plugin is misconfigured: field "password": expected { "write_servers" = type "string", or array of type "string", "read_servers" = type "string", or array of type "string" }, { "servers" = type "string", or array of type "string" }, or { "server" = type "string", or array of type "string" }```

Moving the redis settings to ```/etc/rspamd/local.d/redis.conf``` appeared to fix the issue. Disregarding that some of the history was inevitably lost (unless it's still on the disk somewhere can be moved to Redis somehow, but I probably won't bother), the module now loads again, and the web UI appears to reflect that - I was wondering why I suddenly wasn't seeing detailed scores and from/to addresses anymore.

I'm not sure if the greylisting module is also affected since my settings do not let it trigger often. But the documentation implies that Redis servers must be specifically defined in ```/etc/rspamd/local.d/greylist.conf``` for it to work, so I guess it means that it cannot use global settings from ```/etc/rspamd/local.d/redis.conf```. Might be worth to check it also.

[...]

PR #4562, opened 2023-08-04
fix: redis schema parsing

Conni2461:
Attempt to fix https://github.com/rspamd/rspamd/issues/4318

This issue is still not fixed, at least for us (@ajs124 and me) so he asked me if i could look into it.
I think the problem is a misuse of the tableshape library, also see https://github.com/leafo/tableshape/issues/17

The rewritten code now merges tables rather than using on `extra_fileds` because looking at this source code all remaining keys, which includes `read_server`, `write_server` and `password` e.g. need to have the correct type specified in `extra_fields`, also see https://github.com/leafo/tableshape/blob/master/tableshape/init.lua#L1509-L1515
`password` being a string will then result in what ajs124 saw here: https://github.com/rspamd/rspamd/issues/4318#issuecomment-1453502529

There are two possible options on solving this issue, either we extend tableshape to support this use case of we update the usage here. I decided to do the last one because it seemed simpler.

Its a rather simple solution and i'd like to write some test but struggled to get the lua testing runtime to work. So if you have some pointers on how to get that running, i'd love to add a couple of tests

I validated it with the minimal reproduce

```lua
package.path = package.path .. ";./?.lua;"
local ts = require("contrib.lua-tableshape.tableshape").types

local function table_merge(t1, t2)
  for k, v in pairs(t2) do
    t1[k] = v
  end
  return t1
end

local common = { key_prefix = ts.string, nrows = ts.number, compress = ts.boolean }

local settings_schema = ts.one_of {
  ts.shape(common),
  ts.shape(table_merge({
    read_servers = ts.string + ts.array_of(ts.string),
    password = ts.string:is_optional(),
  }, common)),
  ts.shape(table_merge({
    write_servers = ts.string + ts.array_of(ts.string),
    password = ts.string:is_optional(),
  }, common)),
  ts.shape(table_merge({
    write_servers = ts.string + ts.array_of(ts.string),
    read_servers = ts.string + ts.array_of(ts.string),
    password = ts.string:is_optional(),
  }, common)),
}

local obj1 = {
  key_prefix = "rs_history",
  nrows = 1000,
  compress = true,

  -- read_servers = "127.0.0.1:6381",
  write_servers = "127.0.0.1:6381",
  password = "<password hidden>",
}

local res, err = settings_schema:transform(obj1)
print(res, err)
```

### DIFF
--- a/lualib/lua_redis.lua
+++ b/lualib/lua_redis.lua
@@ -39,25 +39,39 @@ local common_schema = ts.shape {
   sentinel_password = ts.string:is_optional(),
 }
 
-local config_schema = -- Allow separate read/write servers to allow usage in the `extra_fields`
-ts.shape({
+local read_schema = lutil.table_merge({
   read_servers = ts.string + ts.array_of(ts.string),
-}, { extra_fields = common_schema }) +
-    ts.shape({
-      write_servers = ts.string + ts.array_of(ts.string),
-    }, { extra_fields = common_schema }) +
-    ts.shape({
-      read_servers = ts.string + ts.array_of(ts.string),
-      write_servers = ts.string + ts.array_of(ts.string),
-    }, { extra_fields = common_schema }) +
-    ts.shape({
-      servers = ts.string + ts.array_of(ts.string),
-    }, { extra_fields = common_schema }) +
-    ts.shape({
-      server = ts.string + ts.array_of(ts.string),
-    }, { extra_fields = common_schema })
+}, common_schema)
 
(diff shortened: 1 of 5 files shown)


vstakhov:
Perhaps, manual merging is the best option in this case. I'm not quite sure if `tableshape` is an active project, so more control from our side seems like a good idea. Thank you for looking into this issue!

vstakhov:
Just in case: this change has broken many plugins, specifically `reputation`. I'm working on fix.

PR #4572, opened 2023-08-14
Add new exporter to push JSON to a raw TCP socket

lingfish:
This is for example useful with Graylog.

Although Graylog has various HTTP inputs, they are for GELF and others, which would need additional formatters on the `rspamd` side.

Using the raw input results in IO timeouts from the `rspamd` side, because Graylog's raw input doesn't "speak" HTTP.

This resolves the issue.  Combined with an appropriate Graylog pipeline to split into fields, this works well.

Example Graylog pipeline:

```
rule "JSON to fields"
when
  starts_with(to_string($message.message), "{") && ends_with(to_string($message.message), "}")
then
  let json = parse_json(to_string($message.message));
  let map = to_map(json);
  set_fields(map, "rspamd_");
  set_field("rspamd_score", to_double(map.score));
end
```

### DIFF
--- a/src/plugins/lua/metadata_exporter.lua
+++ b/src/plugins/lua/metadata_exporter.lua
@@ -357,10 +357,10 @@ local pushers = {
       timeout = rule.timeout or settings.timeout,
     }, formatted, sendmail_cb)
   end,
-  json_raw = function(task, formatted, rule)
-    local function json_raw_callback(err, code)
+  json_raw_tcp = function(task, formatted, rule)
+    local function json_raw_tcp_callback(err, code)
       if err then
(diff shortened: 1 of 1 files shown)


src/plugins/lua/metadata_exporter.lua
```diff
@@ -356,6 +357,23 @@ local pushers = {
       timeout = rule.timeout or settings.timeout,
     }, formatted, sendmail_cb)
   end,
+  json_raw = function(task, formatted, rule)
```
vstakhov, line 360: I think, it should have the name like `json_raw_tcp` to reflect what it does.

lingfish: Gack.. typo in the next commit, apologies!

PR #4575, opened 2023-08-16
[Feature] Support feed exclusions in phishing module

dragoangel:
Currently, if you use `openfish`, `phishtank` or `generic_service` lists for checking URLs existing `exceptions` or `phishing_exceptions` not applied for them. This lists cover only logic for checking if URL just looks as phishing, but not excluding them from datafeed checks in mentioned services. This PR created a separate exclusion list for this case.

To economy compute units (and DNS requests in case of  `phishtank`) hosts from the exclusion list will be not checked, excluded symbol will be added even it not appear in  `openfish`, `phishtank` or `generic_service`.

### DIFF
--- a/src/plugins/lua/phishing.lua
+++ b/src/plugins/lua/phishing.lua
@@ -69,17 +69,21 @@ local function is_host_excluded(exclusion_map, host)
 end
 
 local function phishing_cb(task)
-  local function check_phishing_map(map, exclusion_map, url, phish_symbol, excl_symbol)
+  local function check_phishing_map(table)
+    local phishing_data = {}
+    for k,v in pairs(table) do
+      phishing_data[k] = v
+    end
+    local url = phishing_data.url
     local host = url:get_host()
 
-    local excl_host = is_host_excluded(exclusion_map, host)
-    if excl_host then
-      task:insert_result(excl_symbol, 0.0, host)
+    if is_host_excluded(phishing_data.exclusion_map, host) then
+      task:insert_result(phishing_data.excl_symbol, 1.0, host)
       return
     end
 
     if host then
-      local elt = map[host]
+      local elt = phishing_data.map[host]
       local found_path = false
       local found_query = false
       local data = nil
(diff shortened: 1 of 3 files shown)


src/plugins/lua/phishing.lua
```diff
@@ -399,6 +425,26 @@ local function insert_url_from_string(pool, tbl, str, data)
...
+  local rspamd_mempool = require "rspamd_mempool"
+  local pool = rspamd_mempool.create()
+
+  local function phishing_feed_exclusion_elt_parser(cap)
+    if insert_url_from_string(pool, new_data, cap, nil) then
+      nelts = nelts + 1
+    end
+  end
+
+  rspamd_str_split_fun(string, '\n', phishing_feed_exclusion_elt_parser)
```
vstakhov, line 440: It's better to use a function from `lua_util` than this legacy global variable.

src/plugins/lua/phishing.lua
```diff
@@ -144,6 +164,12 @@ local function phishing_cb(task)
     local path = url:get_path()
     local query = url:get_query()
 
+    local excl_host = is_host_excluded(exclusion_map, host)
+    if excl_host then
+      task:insert_result(excl_symbol, 0.0, host)
```
vstakhov, line 169: If you add a symbol with dynamic score `0.0` it won't be possible to set any score to it. So a rule of thumb is to add all scores with dynamic score `1.0` and set the static score to `0.0` if the final result is expected to be zero. This allows uses to configure scores as they want eventually.

src/plugins/lua/phishing.lua
```diff
@@ -144,6 +164,12 @@ local function phishing_cb(task)
     local path = url:get_path()
     local query = url:get_query()
 
+    local excl_host = is_host_excluded(exclusion_map, host)
```
vstakhov, line 167:
```suggestion
    if is_host_excluded(exclusion_map, host) then
```

vstakhov: Also it is good to move this check earlier before extracting path and query.

src/plugins/lua/phishing.lua
```diff
@@ -198,15 +224,15 @@ local function phishing_cb(task)
...
       end
 
       if openphish_hash then
-        check_phishing_map(openphish_data, url, openphish_symbol)
+        check_phishing_map(openphish_data, phishing_feed_exclusion_data, url, openphish_symbol, phishing_feed_exclusion_symbol)
       end
 
       if phishtank_enabled then
-        check_phishing_dns(phishtank_suffix, url, phishtank_symbol)
+        check_phishing_dns(phishtank_suffix, phishing_feed_exclusion_data, url, phishtank_symbol, phishing_feed_exclusion_symbol)
```
vstakhov, line 235: Hum, I don't like this amount of positional arguments. Maybe convert them all to a single table and reuse?

vstakhov (changes requested):

src/plugins/lua/phishing.lua
dragoangel: Yes, I know, but exclude list should allow only exclusion, so always should be zero? 😊, but I'm okay to put `1.0` there.

src/plugins/lua/phishing.lua
dragoangel: This [lua_util.str_split(text, delimiter)](https://rspamd.com/doc/lua/lua_util.html#f59ff1) one? I would gladly get a recommendation how to use the existing phishing exclude list or an option to use multiply `[list,list2,...]` of exclusions.

src/plugins/lua/phishing.lua
dragoangel: I will check syntax for it, will try.

src/plugins/lua/phishing.lua
dragoangel: btw, not quite clear what to do with func (3rd element) in case of using str_split?

vstakhov: Yes, that's my fault - it is a specific function from this module. Please ignore.

src/plugins/lua/phishing.lua
vstakhov: It is always better to put 1.0 from the safety/configurability point of view.

dragoangel:
@vstakhov Hi, I updated PR with mentioned changes, can you please check if it's okay now?

vstakhov (approved):

PR #4581, opened 2023-09-02
[Feature] Add ICAP Filename and Content-Type from TODO List

lbahtarliev:
Hello guys,

First of all let me congratulate you for the good job done on this project. Well done!
I found it while I was particularly searching for an ICAP Client implementation for SMTP, due to some internal company regulations and security tightening we need to implement. I noticed that in the TODO list there is an entry for developing dynamic Filename/Content-Type, rather than going with the static req_fake_url. 
I banged my head for a couple of days until I`ve found the approach I`m suggesting. It may not be the cleanest solution though, so your comments and thoughts will be much appreciated.
I was thinking that instead of providing content as a result of get_content() userdata type, provide it as a mime_part from which it can be easily fetched with get_content, but that will ease the implementation I did. Then one could just use part:get_filename() part:get_detected_type()

PS: Bear with me, this is my first pull request ever on github, so If I did something out of the ordinary, or missed some rule/guideline do not flame me a lot :)

Yours,
Lyuben


lualib/lua_scanners/icap.lua
```diff
@@ -240,10 +240,21 @@ local function icap_check(task, content, digest, rule, maybe_part)
 
       local function get_req_headers()
 
+        local in_filename, in_type, in_subtype, in_client_ip = nil
+        local parts = task:get_parts()
+        for _, part in pairs(parts) do
```
vstakhov, line 245: I don't understand this logic. There could be multiple parts in a message with different filenames and content types. With this approach we just select the first one.

vstakhov (changes requested):

lualib/lua_scanners/icap.lua
lbahtarliev: I found out, that icap_check is called multiple times for each attachment passed as content. Because task variable is always the same and returns all the parts, and content variable doesn't have get_detected_type() method. So, on each call I get all the parts from task and check against content variable to see which one is actually the current one. 
As you wrote, now I can imagine how expensive get_parts can be, especially with larger attachments. 
I will try the "maybe_part" route as it seems the most logical one.
Just a few questions - if maybe_part is actually a mime part object, will it have the get_detected_type() and get_filename() methods? and as content is coming as a userdata without any methods can it be converted to mime part object?

Thanks,
Lyuben

lualib/lua_scanners/icap.lua
```diff
@@ -240,10 +240,21 @@ local function icap_check(task, content, digest, rule, maybe_part)
 
       local function get_req_headers()
 
+        local in_filename, in_type, in_subtype, in_client_ip = nil
+        local parts = task:get_parts()
```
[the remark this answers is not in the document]
lbahtarliev: Just checked this option. Unfortunately, maybe_part is always nil when icap_check is called with an attachment as a content... 
It seems in order to keep it optimized, content variable is the way to go, but it is just a data in the memory without the needed methods. And I`m not quite sure how I can convert that to a mime part, if possible at all.

lbahtarliev (commented):
It turns out external_services.lua plugin does not submit maybe_part argument at all. 
So a more elegant solution is to hook in the plugin where it calls icap scanner for each mime part and get filename, content type, content subtype there, then pass them in place of maybe_part (which is never used, or at least I couldn't find its usage in regards to icap services)

lbahtarliev:
I will create a clean one.

PR #4585, opened 2023-09-06
[Minor] Remove unwanted debug logging

fatalbanana:
I expect it's unwanted anyway. I thought about setting a better descriptive text but nothing came to mind.


vstakhov:
In fact, I had some linter in my `overcommit` hooks to prevent this to come to the production repo, however, it seems I have lost those on git cleanup.

PR #4586, opened 2023-09-06
[Minor] Improve Afrikaans language detection

fatalbanana:
- Added diacritics flag for Afrikaans language (not clear it helped something)
 - Added Afrikaans stop-words
 - Removed ambiguous Dutch stop-words

Perhaps it might be worthwhile. Number of Afrikaans speakers is a little under half the number of Dutch speakers.


fatalbanana:
Not sure about this - probably it calls for a bit more attention that I'm ready to give it - I want to take special care to avoid harming Dutch ....

PR #4587, opened 2023-09-06
[Fix] DMARC reporting: fix reporting for subdomains

fatalbanana:


vstakhov (approved):

vstakhov:
Is that ready for merge?

fatalbanana:
> Is that ready for merge?

It should be I hope though I'd still like to have some tests for DMARC reporting in general ...

PR #4588, opened 2023-09-07
Fix regression in invokation of rspamd-test

ln5:
A space seems to have sneaked into the setting of LUA_PATH environment.


arkamar:
Just for info, it happened again, see #4560. It looks like a linter issue.

PR #4590, opened 2023-09-08
[Fix] Correct format pattern for RE tree tempfile name

ln5:
Triggered by -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -DENABLE_CLANG_PLUGIN=ON -DENABLE_HYPERSCAN=ON.

### DIFF
--- a/src/libserver/re_cache.c
+++ b/src/libserver/re_cache.c
@@ -1952,7 +1952,7 @@ rspamd_re_cache_compile_timer_cb(EV_P_ ev_timer *w, int revents)
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


vstakhov:
Hum, I really don't understand how has it passed our ci then, so I need to check that. Thank you for a good finding!

ln5:
FWIW, the build-clang target in .drone.yml doesn't enable hyperscan. No idea if that's the file driving your CI though.

vstakhov:
I see, it should be enabled then indeed.

PR #4595, opened 2023-09-11
[Feature] Add ICAP Content-Type and Filename from TODO List

lbahtarliev:
It turns out, that external_services plugin is not submitting maybe_part parameter at all when calling icap services check method. In order to avoid expensive operations in icap.lua scanner, I am just getting the filename and content-type/subtype in the external_services plugin and submitting that as maybe_part parameter. Of course, a new parameter can be introduced should maybe_part is used, but I wasn't able to find any usage of it at least related to the ICAP functionality.

### DIFF
--- a/lualib/lua_scanners/icap.lua
+++ b/lualib/lua_scanners/icap.lua
@@ -241,13 +241,15 @@ local function icap_check(task, content, digest, rule, maybe_part)
       local function get_req_headers()
 
         local in_client_ip = task:get_from_ip()
-        lua_util.debugm(rule.name, task, 'URL: http://%s/%s | Content-Type: %s/%s',
-            in_client_ip, maybe_part[1], maybe_part[2], maybe_part[3])
-
         local req_hlen = 2
-        table.insert(req_headers, string.format('GET http://%s/%s HTTP/1.0\r\n', in_client_ip, maybe_part[1]))
         table.insert(req_headers, string.format('Date: %s\r\n', rspamd_util.time_to_string(rspamd_util.get_time())))
-        table.insert(http_headers, string.format('Content-Type: %s/%s\r\n', maybe_part[2], maybe_part[3]))
+        if maybe_part then
+          table.insert(req_headers, string.format('GET http://%s/%s HTTP/1.0\r\n', in_client_ip, maybe_part:get_filename()))
+          table.insert(http_headers, string.format('Content-Type: %s/%s\r\n', maybe_part:get_detected_type()))
+        else
+          table.insert(req_headers, string.format('GET %s HTTP/1.0\r\n', rule.req_fake_url))
+          table.insert(http_headers, string.format('Content-Type: application/octet-stream\r\n'))
+        end
         if rule.user_agent ~= "none" then
           table.insert(req_headers, string.format("User-Agent: %s\r\n", rule.user_agent))
         end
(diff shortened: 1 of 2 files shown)


src/plugins/lua/external_services.lua
```diff
@@ -181,12 +181,19 @@ local function add_scanner_rule(sym, opts)
       fun.each(function(p)
         local content = p:get_content()
         if content and #content > 0 then
-          cfg.check(task, content, p:get_digest(), rule)
+          local in_fname = p:get_filename()
+          local in_type, in_stype = p:get_detected_type()
+          local part_info = {in_fname,in_type,in_stype}
+          cfg.check(task, content, p:get_digest(), rule, part_info)
```
vstakhov, line 187: I think the original idea was just to pass `mime_part` object (hence `p`) and not to split it into some array.

src/plugins/lua/external_services.lua
```diff
@@ -181,12 +181,19 @@ local function add_scanner_rule(sym, opts)
...
         end
       end, common.check_parts_match(task, rule))
 
     else
-      cfg.check(task, task:get_content(), task:get_digest(), rule)
+      local in_fname = "mail"
+      local in_type = "application"
+      local in_stype = "octet-stream"
+      local part_info = {in_fname,in_type,in_stype}
+      cfg.check(task, task:get_content(), task:get_digest(), rule, part_info)
```
vstakhov, line 196: If we don't have part, then we must just pass `nil` here and reconstruct the necessary information afterwards.

vstakhov (changes requested):

src/plugins/lua/external_services.lua
lbahtarliev: You are absolutely right. How I couldn't think of that... I found that maybe_part is always nil, but it didn't occur to me to just pass p object instead of doing all these variable assignments here.
However, will that mean that the extraction of content type and subtype, as well as filename should go in icap.lua in lua_scanners?

src/plugins/lua/external_services.lua
lbahtarliev: Maybe if it's nil we can stick with the already available req_fake_url? Otherwise, I am not sure how I can reconstruct the missing information in the lua_scanner itself... May you can point me in the right direction?

lbahtarliev (commented):
Another attempt.

lualib/lua_scanners/icap.lua
```diff
@@ -240,10 +240,19 @@ local function icap_check(task, content, digest, rule, maybe_part)
 
       local function get_req_headers()
 
+        local in_client_ip = task:get_from_ip()
+        lua_util.debugm(rule.name, task, 'URL: http://%s/%s | Content-Type: %s/%s',
```
vstakhov, line 244: This will crash the plugin when `maybe_part` is nil. Lua does not do lazy evaluations. Also, that will call two C functions twice (get_filename and get_detected_type).

vstakhov (changes requested):

lualib/lua_scanners/icap.lua
lbahtarliev: Didn't notice I am missing a check before the debug message. Is it OK to just remove it?

lbahtarliev: Just removed the debug message

PR #4596, opened 2023-09-11
[Feature] rbl: support checking numeric URLs in isolation

fatalbanana:


vstakhov:
Why don't we just enable this flag by default then?

fatalbanana:
Usually URLs include numeric URLs; we have `no_ip` toggle to suppress this behaviour (for lists that only contain domains).

The problem this feature tries to address is how to check URLs against lists which contain IP addresses only. [Spamhaus' DQS config](https://github.com/spamhaus/rspamd-dqs/blob/master/3.x/rbl.conf#L148-L156) wants to do it but I expect it sends unwanted queries as configured there.

PR #4602, opened 2023-09-15
[Minor] Revisit numeric_urls

fatalbanana:
Make `numeric_urls` behave like `urls`:

* `images` and `content_urls` are not included by default
* Including these does not exclude anything.

I'm open to suggestions about how it should work / what configuration toggles may be wanted.


vstakhov:
Very good, thank you!

PR #4605, opened 2023-09-18
Update configwizard.lua

Dani778:
FIXED dkim key generation in configwizard, update from old api

On current configwizard.lua I was getting such error:
call to rspamadm lua script failed (2): /usr/share/rspamd/lualib/rspamadm//configwizard.lua:382: attempt to call field 'generate_keypair' (a nil value); trace: [1]:{/usr/share/rspamd/lualib/rspamadm//configwizard.lua:784 - <unknown> [Lua]};
So I modified it basing on rspamd cryptobox docs and dkim_keygen.lua

### DIFF
--- a/lualib/rspamadm/configwizard.lua
+++ b/lualib/rspamadm/configwizard.lua
@@ -382,7 +382,10 @@ local function setup_dkim_signing(cfg, changes)
     end
     printf("--")
   end
-
+  local function print_public_key(pk)
+    local base64_pk = tostring(rspamd_util.encode_base64(pk))
+    printf('v=DKIM1; k=rsa; p=%s\n', base64_pk)
+  end
   repeat
     if has_domains then
       print_domains()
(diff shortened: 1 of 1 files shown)


lualib/rspamadm/configwizard.lua
```diff
@@ -406,22 +409,12 @@ local function setup_dkim_signing(cfg, changes)
...
-        local content = f:read("*all")
-        f:close()
-        print("To make dkim signing working, you need to place the following record in your DNS zone:")
-        print(content)
+        local rsa = require "rspamd_rsa"
+        local sk, pk = rsa.keypair(2048)
+        sk:save(privkey_file, 'pem')
+        print("You need to chown private key file to rspamd user!!")
+        print("To make dkim signing working, to place the following record in your DNS zone:")
+        print_public_key(tostring(pk))
```
vstakhov, line 417: Why do we need `tostring` here?

vstakhov (approved):

lualib/rspamadm/configwizard.lua
Dani778: Generally speaking, function `print_public_key()` is slightly modified function from dkim_keygen.lua, and when you don't use `tostring`, function simply throws an error. I think because `pk` is an object and later on in `public_print_key()` its treated like a string by function `rspamd_util.encode_base64(pk)`

vstakhov: `rspamd_util.encode_base64` should clearly accept `rspamd_text` as well as a `string`.

vstakhov: Ah, but `pk` is not a text, it is another userdata type. I see then.

PR #4612, opened 2023-09-25
[Test] Rudimentary tests for known_senders plugin

fatalbanana:
Suggestions about things to test and such welcome.


vstakhov:
Are we sure that the test cases are executed in that specific order?

fatalbanana:
> Are we sure that the test cases are executed in that specific order?

Yes, they will be executed completely synchronously.

PR #4622, opened 2023-09-30
[Fix] Add url_encode function and encode filename in ICAP Request headers, which if contains space invalidate header

lbahtarliev:
Add url_encode (and url_decode) functions to lua_util, as url encode function was not found. Add url_encode to ICAP REQ Headers filename, becuase if raw data contain spaces header becomes invalid.

### DIFF
--- a/lualib/lua_util.lua
+++ b/lualib/lua_util.lua
@@ -1621,6 +1621,39 @@ local function join_path(...)
 end
 exports.join_path = join_path
 
+---[[[
+--- @function lua_util.url_encode(str)
+-- URL encode string according to RFC 3986
+--
+-- @param str The string to URL encode.
+-- @return The URL encoded string.
+--
+---]]]
+
+local function url_encode (str)
+   str = string.gsub (str, "([^0-9a-zA-Z! '&/:()*._~-])", -- locale independent
+      function (c) return string.format ("%%%02X", string.byte(c)) end)
+   str = string.gsub (str, " ", "+")
+   return str
+end
+exports.url_encode = url_encode
+
+---[[[
+--- @function lua_util.url_decode(str)
+-- URL decode string according to RFC 3986
+--
+-- @param str The string to URL decode.
+-- @return The URL decoded string.
+--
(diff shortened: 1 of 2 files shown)


lbahtarliev:
Force pushed to fix an error in automated build for reaching maximum line length

lualib/lua_util.lua
```diff
@@ -1621,6 +1621,39 @@ local function join_path(...)
...
+--- @function lua_util.url_encode(str)
+-- URL encode string according to RFC 3986
+--
+-- @param str The string to URL encode.
+-- @return The URL encoded string.
+--
+---]]]
+
+local function url_encode (str)
+   str = string.gsub (str, "([^0-9a-zA-Z! '&/:()*._~-])", -- locale independent
```
vstakhov, line 1634:
```suggestion
   str = string.gsub(str, "([^%w _%%%-%.~])",
```

lualib/lua_util.lua
```diff
@@ -1621,6 +1621,39 @@ local function join_path(...)
...
+-- @return The URL decoded string.
+--
+---]]]
+
+local function url_decode (str)
+   str = string.gsub (str, "+", " ")
+   str = string.gsub (str, "%%(%x%x)", function(h) return string.char(tonumber(h,16)) end)
+   return str
+end
+exports.url_decode = url_decode
```
vstakhov, line 1655: This function exists in `rspamd_util` C library: `rspamd_util.decode_url`.

vstakhov (commented):
It looks good in general, but I'd appreciate some unit tests for encoding/decoding sanity (including utf8 stuff).

lualib/lua_util.lua
lbahtarliev: I am not good at C, otherwise I would have proposed the encode function in the same library. I will do some research and upgrade myself, then will propose another commit with amendment to the C library, as well as unit tests for both. 
Thanks, Vsevolod! Always a pleasure to learn from you.

PR #4623, opened 2023-09-30
maps: Replace password in URL with wildcards

yo000:
Hello,

This PR replaces password with *** when logging http map name or uri, so passwords don't lies in log files

### DIFF
--- a/src/libserver/maps/map.c
+++ b/src/libserver/maps/map.c
@@ -212,6 +212,10 @@ http_map_error(struct rspamd_http_connection *conn,
 {
 	struct http_callback_data *cbd = conn->ud;
 	struct rspamd_map *map;
+	gchar *uri;
+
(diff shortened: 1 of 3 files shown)


vstakhov:
Hum, I really don't like the idea to encode password inside URL. It seems to be quite obsoleted and Rspamd should support HTTP auth for maps.

yo000:
I prepared an answer about what structure/how could it be done, and saw you are currently working on this.
That's awesome!

I just add regarding my PR, all replacements of "bk->uri" at log time could be replaced with the use of "bk->name" in which credentials would be masked only once at map declaration time.

fatalbanana:
We have support for [credentials outside of map URLs](https://docs.rspamd.com/configuration/maps/#authentication-for-http-maps) now.

PR #4627, opened 2023-10-05
[Fix] Prevent DNSWL sabotage

mdierksen:
When exceeding the query limit for DNSWL it can happen that instead of the returncode 127.0.0.255, that according to documentation (https://www.dnswl.org/?page_id=15) indicates a block, the returncode 127.0.10.3 is returned for all queries.

According to documentation (https://www.dnswl.org/?page_id=15) the 127.0.10.3 returncode indicates the highest level of trustworthiness that should never be blocked and a category of 'some special cases'.

As it turns out that documentation is a lie and that 127.0.10.3 returncode is used by DNSWL to intentionally sabotage email security by marking all sending servers as highly trustworthy (https://www.dnswl.org/?p=120).


vstakhov:
That doesn't sound nice indeed. Perhaps they have their reasons but this sort of behaviour is not something that should be treated easily...

PR #4648, opened 2023-10-17
[Test] Banish luacheck from functional tests

fatalbanana:
- Use luacheck from GitLab


vstakhov:
I have that in my commit hooks btw :)

PR #4652, opened 2023-10-19
[Feature] External_relay: support regex maps for hostnames/usernames

fatalbanana:


vstakhov:
Cannot we just specify the map augmentation in the existing logic?

fatalbanana:
> Cannot we just specify the map augmentation in the existing logic?

Yes ofcourse. It works...

PR #4656, opened 2023-10-22
Tighten rspamd's attachment policy

twesterhever:
This merge request seeks to clean up the current list of bad attachments in order to make it clear that these decision have been incorporated from third parties whenever possible, adds a couple of additional bad attachment types from [Microsoft's Outlook attachment handling policy](https://support.microsoft.com/en-us/office/blocked-attachments-in-outlook-434752e1-02d3-4e90-9124-8b81e49a8519?ui=en-us&rs=en-us&ad=us), as well as a couple of attachment types never observed in legitimate e-mails.

The latter was triggered by a recent onslaught in phishing e-mails leveraging MHTML.


src/plugins/lua/mime_types.lua
```diff
@@ -53,7 +53,7 @@ local settings = {
 
   bad_extensions = {
     cue = 2,
-    exe = 1,
+    exe = 2,
```
moisseev, line 56: IMO 2 is too much for default configuration as it will affect some legitimate messages.

twesterhever: Thank you, I just reverted the offending commit.

src/plugins/lua/mime_types.lua
```diff
@@ -52,29 +52,34 @@ local settings = {
...
-    lnk = 4,
-    scr = 4,
     -- In contrast to HTML MIME parts, dedicated HTML attachments are considered harmful
     htm = 1,
     html = 1,
     shtm = 1,
     shtml = 1,
     -- Have you ever seen that in legit email?
     ace = 4,
+    app = 1,
```
fatalbanana, line 66: It's already listed.

~~~
+ luacheck -q --no-color .
Checking src/plugins/lua/mime_types.lua           1 warning

    src/plugins/lua/mime_types.lua:66:5: (W314) value assigned to field 'app' is overwritten on line 110 before use
~~~

twesterhever: Grr, my bad. Fixed, thank you!

PR #4657, opened 2023-10-24
[Minor] rbl: support use of different matchers for return codes

fatalbanana:
The idea is to:

- [X] get rid of un-needed usage of Lua patterns
- [X] deprecate unqualified use of Lua patterns (prompt users to set `matcher = "luapattern"`)
- [X] add different options for matching, like regexp
- [x] and radix...
- [x] and glob...


lualib/plugins/rbl.lua
```diff
@@ -199,6 +200,21 @@ local function convert_checks(rule)
     rule.from = true
   end
 
+  if rule.returncodes and not rule.matcher then
+    for _, v in pairs(rule.returncodes) do
+      for _, e in ipairs(v) do
+        if e:find('%', 1, true) then
+          rspamd_logger.warnx(rspamd_config, 'implicitly enabling luapattern matcher for rule %s', rule.symbol)
```
vstakhov, line 207: Maybe that should be on `messagex` or `infox` level. Btw, you no longer need `x` suffix in logging methods :)

vstakhov (approved):
Looks very good, thank you! Please update docs if possible to match these new options.

PR #4678, opened 2023-11-02
[Fix] rbl: unbreak `exclude_local` / `exclude_private_ips`

fatalbanana:


src/plugins/lua/rbl.lua
```diff
@@ -151,9 +151,7 @@ local function gen_check_rcvd_conditions(rbl, received_total)
     end
     if ((rh.real_ip:get_version() == 6 and rbl.ipv6) or
         (rh.real_ip:get_version() == 4 and rbl.ipv4)) and
-        ((rbl.exclude_private_ips and not rh.real_ip:is_local()) or
-            not rbl.exclude_private_ips) and ((rbl.exclude_local_ips and
-        not is_excluded_ip(rh.real_ip)) or not rbl.exclude_local_ips) then
+        (rbl.exclude_local and (not rh.real_ip:is_local() or is_excluded_ip(rh.real_ip))) then
```
fatalbanana, line 154: I have doubts about this. Everywhere `exclude_local` is used to exclude by sending IP and not by what is to be looked up but here it is unfortunately both ...

fatalbanana: OK, so `exclude_local` was supposed to be for the local exclusion map and not local/private IPs, currently it is a bit muddled & the documentation is similarly in two minds about it.

vstakhov: Yes, it's very confusing. I think we should not treat `private` ips anyhow special, they should be a default for local but overrideable by a user.

fatalbanana: Yes, that's much better, let's move forward with it...

PR #4680, opened 2023-11-03
[Minor] Treat *.zpaq attachments as archives and harmful

twesterhever:
Rationale: https://dshield.org/diary/Malware+Dropped+Through+a+ZPAQ+Archive/30366/


vstakhov:
I think we should move that to rspamd/maps some day. @fatalbanana could you please take a look when you have some time?

PR #4689, opened 2023-11-07
[Feature] Allow disabling of GTUBE rejection

fatalbanana:


vstakhov:
One thing that might be better is to have a single option of gtube patterns instead of 2 separate ones.

PR #4693, opened 2023-11-09
[Fix] known_senders: fix config handling

fatalbanana:
Reported by Bensch on Telegram


vstakhov:
In fact, we should have some basic schema that will be applied for all modules. This sort of errors are very nasty and hard to check. Perhaps, we need to rewrite `ts` module by ourselves, as the error reporting is currently just awful. It's something I have in my mind tbh... We need something very simple and something that can be easily transformed into documentation, examples, json schema etc.

Issue #4690, opened 2023-11-07
[BUG] task_get_headers(true) not returning ARC headers 


JasonStephenson:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
task:get_headers(true) not returning headers created by arc signing module if no ARC headers existed previously

### Steps to Reproduce

1. Rspamd configured to run via Postfix + milter
2. arc module configured to do signing
3. aws_s3 module configured to do writing
4. Receive an email which contains no ARC headers (ARC-Message-Signature, ARC-Authentication-Results, ARC-Seal)
5. Observe arc module correctly signing 
6. Observe `task:get_headers(true)` not returning the ARC headers
7. Observe `task:get_header_full` can return the ARC headers
8. Receive an email which **does** already contain ARC headers (ex:google)
9. Observe `task:get_headers` correctly returning ARC headers

**Expected behavior**

- `task:get_headers(true)` should return ARC headers created by arc module regardless of whether existing ARC headers are present
- raw attribute should consistantly start with the MIME header (see addtional notes)
- raw attribute should consistantly end with \r\n (see additional notes)

### Versions

- 3.2

### Additional notes
Also seeing differences between ARC headers created by rspamd and ARC headers already present in the email.

Already present:
```
    {
      "empty_separator": false,
      "order": 7,
      "separator": " ",
[...]

PR #4716, opened 2023-11-23
Fix issues with the raw header on header insertion/modification

vstakhov:
- [x] Fix raw headers values
- [x] Unfold value for inserted headers
- [x] Alter `headers_order` when adding new headers

Issue: #4690


vstakhov:
It should be finished by now.

PR #4734, opened 2023-12-08
rspamd-3.7: fix systemd logging

fatalbanana:


vstakhov:
Do we check 3.7 branch in ci btw?

Issue #4741, opened 2023-12-12
[BUG] Returncode matching in RBL module


mdierksen:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

### Problem

I ran into a weird problem with the returncode matching in the RBL module.

I setup Spamhaus DQS using https://github.com/spamhaus/rspamd-dqs and tried to verify the proper function using the Spamhaus Blocklist Tester tool (https://blt.spamhaus.com/) and checking the Rspamd symbols.

The 'zrd-dqs-ehlo' and 'zrd-dqs-from' tests both test the hostname 'zrd-dqs.blt.spamhaus.net' against the 'zrd.dq.spamhaus.net' list which returns 127.0.2.2.

With the configured 'spamhaus_zrd' list:
```
spamhaus_zrd {
    rbl = "XXX.zrd.dq.spamhaus.net";
    helo = true;
    rdns = true;
    dkim = true;
    disable_monitoring = true;
    returncodes {
        RBL_ZRD_VERY_FRESH_DOMAIN = "127.0.2.[2-4]+";
        RBL_ZRD_FRESH_DOMAIN = ["127.0.2.[5-9]+", "127.0.2.1[0-9]+", "127.0.2.2[0-4]+"];
        RBL_ZRD_DONT_QUERY_IPS = "127.0.2.255";
    }
}
```
this returncode should yield a RBL_ZRD_VERY_FRESH_DOMAIN symbol.

However all I get is the symbol
`SPAMHAUS_ZRD(0.00)[zrd-dqs.blt.spamhaus.net:helo:127.0.2.2];`

When I change the configuration to
`RBL_ZRD_VERY_FRESH_DOMAIN = "127.0.2.2";`
[...]

PR #4743, opened 2023-12-13
[Minor] rbl: try harder to identify lua patterns

fatalbanana:
#4741

Try harder to identify unambiguous Lua patterns (maybe there's no need to worry so much about ambiguity in fact).

Also shout louder about auto-correction: user may have intended regex or just have no idea about the matter on hand.


fatalbanana:
Another approach could be to verify that what we have is indeed an IP address.

PR #4764, opened 2023-12-29
Fix an overzealous s/NULL/nullptr/g

jvoisin:
Reported to me by an irc friend.
You might want to ask for a CVE I think.

### DIFF
--- a/src/libserver/cfg_rcl.cxx
+++ b/src/libserver/cfg_rcl.cxx
@@ -2049,7 +2049,7 @@ rspamd_rcl_config_init(struct rspamd_config *cfg, GHashTable *skip_sections)
 									   rspamd_rcl_parse_struct_string,
 									   G_STRUCT_OFFSET(struct rspamd_config, ssl_ciphers),
 									   0,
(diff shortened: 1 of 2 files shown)


vstakhov:
I don't see any way how this could be exploited, but it is a good catch anyway!

jvoisin:
From the [documentation](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html):

> The cipher suites offering no authentication. This is currently the anonymous DH algorithms and anonymous ECDH algorithms. These cipher suites are vulnerable to “man in the middle” attacks and so their use is discouraged. These are excluded from the DEFAULT ciphers, but included in the ALL ciphers. Be careful when building cipherlists out of lower-level primitives such as kDHE or AES as these do overlap with the aNULL ciphers. When in doubt, include !aNULL in your cipherlist.

It seems that including `!aNULL` doesn't do anything in this case, you're right.

vstakhov:
Yes, it's a good example how a simple `s//` can lead to a potential vulnerability. Luckily, not this time but still good to fix :)

PR #4769, opened 2024-01-02
Update greylist.conf

ShaiMagal:
Hi, there should be **numeric** value.
Or it die with error and another processing after this error.
`2024-01-03 00:16:47 #682935(normal) <7e95eb>; task; lua_metric_symbol_callback: call to (GREYLIST_SAVE) failed (2): /usr/share/rspamd/plugins/greylist.lua:447: bad argument #1 to 'toint' (number expected, got string); trace: [1]:{[C]:-1 - toint [C]}; [2]:{/usr/share/rspamd/plugins/greylist.lua:447 - <unknown> [Lua]};`

Easily to test for example here:
https://www.tutorialspoint.com/execute_lua_online.php
Sample code with the same error:
```
local toint = math.ifloor or math.floor
print(toint("1 day"))
```

image

### DIFF
--- a/conf/modules.d/greylist.conf
+++ b/conf/modules.d/greylist.conf
@@ -19,7 +19,7 @@ greylist {
     "$LOCAL_CONFDIR/local.d/maps.d/greylist-whitelist-domains.inc",
   ];
 
-  expire = 1d; # 1 day by default
+  expire = 86400; # 86400 (1 day by default)
   timeout = 5min; # 5 minutes by default
   key_prefix = "rg"; # default hash name
   max_data_len = 10k; # default data limit to hash


vstakhov:
I don't understand the proposed change. In ucl, the value `expire = 1d` is exactly the same as `expire = 86400`, and they are both treated as numbers not strings. Of course, `expire = "1d"` will not work in this context.

ShaiMagal:
@vstakhov 
If I have "expire = 1d;" it give the error to the log "bad argument #1 to 'toint' (number expected, got string)" (like I wrote in my PR)

If I change it to "expire = 86400;" - no error and working well.

I don't use expire = "1d", If I revert changes back to expire = 1d; Again I got same error.

I tried now again a problem persist. Tested 2 versions of rspamd. Latest 3.7.5 

Try this LUA code:
local toint = math.ifloor or math.floor
print(toint("1 day"))
It's same, what is used in rspamd and it gave you same error.

fatalbanana:
Are you not indeed quoting this setting somewhere in configuration (you should not)? What build of Rspamd?

~~~
$ docker run -u root -ti rspamd/rspamd:3.7.5 bash
root@ab81ecea7cf8:/# echo 'expire = "2d";' > /etc/rspamd/local.d/greylist.conf # this is broken
root@ab81ecea7cf8:/# rspamadm configdump greylist.expire
*** Section greylist.expire ***
"2d"
*** End of section greylist.expire ***
root@ab81ecea7cf8:/# echo "expire = 2d;" > /etc/rspamd/local.d/greylist.conf # this works
root@ab81ecea7cf8:/# rspamadm configdump greylist.expire
*** Section greylist.expire ***
172800
*** End of section greylist.expire ***
~~~

ShaiMagal:
- version: 3.7.5, ubuntu 22, apt packages.
Tested again, same problem.
file /etc/rspamd/local.d/greylist.conf:
A) `expire = 86400;`
WORKING
B) `expire = 1d;`
NOT WORKING

Why you didn't try my example?

There is clean code to test it without rspamd depencency, and there you can see same error.
```
local toint = math.ifloor or math.floor
print(toint("1 day"))
```

fatalbanana:
> Please, try my example in first post.

There is no expectation that that should work (rather the expectation is that something like `1d` (unquoted) reaches Lua as a number).

`1 day` OTOH is not a valid variant and will be a string.

ShaiMagal:
@fatalbanana yes, but this my example is from real problem at all of our servers with rspamd... I don't understand, how we can be only one with this problem... (Tested on more than 10 servers - everywhere same problem) - Ubuntu 16-22, rspamd 2.x-3.x

fatalbanana:
What does `rspamadm configdump greylist.expire` show?

Is `/etc/rspamd/modules.d/greylist.conf` integral?

ShaiMagal:
@fatalbanana sorry for late reply, I miss it:

Result of command.
```
rspamadm configdump greylist.expire
*** Section greylist.expire ***
86400
*** End of section greylist.expire ***
```


but in every case, like I wrote more times, if I use:
`expire = 1d;`
I got error everytime:
`#682935(normal) <7e95eb>; task; lua_metric_symbol_callback: call to (GREYLIST_SAVE) failed (2): /usr/share/rspamd/plugins/greylist.lua:447: bad argument #1 to 'toint' (number expected, got string); trace: [1]:{[C]:-1 - toint [C]}; [2]:{/usr/share/rspamd/plugins/greylist.lua:447 - <unknown> [Lua]};`


If I use:
`expire = 86400;`
No problems in logs.

fatalbanana:
> Result of command.

Is that with the apparently bad setting in the configuration or the literal number?

> but in every case, like I wrote more times, if I use:

We don't see that such a problem should exist & can't reproduce that.

Issue #4515, opened 2023-06-19
[Feature] Allow opt-out of DMARC reports for DMARC reports


plinss:
## Summary
Say there are two domains, a.com and b.com, both have DKIM/DMARC configured and both send DMARC reports.

A user from a.com sends a single email to b.com. b.com sends a DMARC report to a.com's DMARC reporting address, which happens to be @a.com. a.com then, having received an email from b.com, sends a DMARC report to b.com's DMARC reporting address which happens to be @b.com. Now a.com and b.com will be sending DMARC reports to each other for the rest of time in response to each other's prior DMARC report... even tough users from either domain never send each other another email.

There needs to be a way to break this cycle. One thought is to allow a list of specific addresses to opt-out of generating DMARC reports. So for example, messages sent to dmarc@a.com would not generate a DMARC report to the sender.

## Motivation

Prevent an endless series of DMARC reports.

## Describe alternatives you've considered

Ideally, one or both domains are configured to have their DMARC reports sent to another domain entirely. But that's not always the case.

reinob:
The common solution (AFAIK) to this is to use send your DMARC reports using an address with a subdomain having no DMARC reporting, e.g. in my case my DMARC reports are sent with an address like "report@dmarc.example.org", and the TXT record _dmarc.dmarc.example.org has "v=DMARC1; p=none;", so that no reports are requested.

So if your DMARC reports are sent with that address, the recipients won't sent a report-for-the-report.

MexHigh:
Related: #4095

PR #4772, opened 2024-01-10
[Feature] Support suppressing DMARC reporting for particular recipients

fatalbanana:
Existing possibility could be to disable DMARC completely for some particular recipients.

Suggested by @plinss in [#4515](https://github.com/rspamd/rspamd/issues/4515)


PHPGangsta:
This is an important feature! Currently I try to block those recipients in Postfix, but then the data has been collected and a report generated, only to be discarded by Postfix. So skipping to collect the data is the best solution!

There already is a setting "no_reporting_domains", but I like to block the recipient email addresses a lot more.

I like that it's a map that can be grabbed from an http-URL for example.

Does it make sense that rspamd provides a (basic) list with the most common entries? Like the other provided basic lists here:
https://github.com/rspamd/maps/tree/master/rspamd

Or will someone start a public one (somewhere on Github) that we can include? It would be nice if not everyone has to create his/her own local list, crowdsourcing makes sense here I guess.

PS: I added a small sentence to the documentation of rspamd about the new `exclude_recipients` feature :
https://github.com/rspamd/rspamd.com/pull/721

PR #4774, opened 2024-01-11
Rewrite redis_cache logic in statistics

vstakhov:
The current statistics has a serious flaw mentioned by @dragoangel - learned ids never expire. After analysing the current code I have decided to do a complete refactoring by moving the logic to the Redis scripts side (similarly to the statistics module itself).


lualib/redis_scripts/bayes_cache_learn.lua
```diff
@@ -0,0 +1,61 @@
...
+    local prefix = conf.cache_prefix .. string.rep("X", i)
+    local exists = redis.call('EXISTS', prefix)
+
+    if exists then
+      if expired then
+        redis.call('DEL', prefix)
+        redis.call('HSET', prefix, cache_id, is_spam)
+
+        -- Do not expire anything else
+        expired = true
```
jeremy, line 51: Bug. This branch is unreachable since `expired = false` initially but is only set to `true` within an `if expired` condition.

The result is that key won't be added when the caches are full. The next key add will succeed.

The consequence is that once the caches are full (50k keys by default) then one spam learn will be dropped per cache rotation, 1 in 10k keys by default.

vstakhov: That's a very good catch, thank you!

Issue #4784, opened 2024-01-20
[BUG] rspamd v3.8.0 fails to start with lua 5-1 (works with lua 5-3 and lua 5-4)


erialor:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
rspamd fail to start when using lua 5-1 with:

mail ~ # /etc/init.d/rspamd start
rcl parse error: cannot init lua file /usr/share/rspamd/rules/rspamd.lua: error loading module 'lua_maps' from file '/usr/share/rspamd/lualib/lua_maps.lua':
        /usr/share/rspamd/lualib/lua_maps.lua:564: nesting of [[...]] is deprecated near '['; trace: [1]:{[C]:-1 - <unknown> [C]}; [2]:{[C]:-1 - require [C]}; [3]:{/usr/share/rspamd/lualib/global_functions.lua:21 - <unknown> [main]}; [4]:{[C]:-1 - require [C]}; [5]:{rspamd.lua:19 - <unknown> [main]};
 * ERROR: rspamd failed to start
mail ~ #

### Steps to Reproduce

1. Make sure your using Lua 5.1
2. Compile rspamd 3.8.0
3. Start rspamd 3.8.0

**Expected behavior**
After changing to Lua 5.3 or Lua 5.4 it *just* starts:
mail ~ # /etc/init.d/rspamd start
 * Caching service dependencies ...                                                                                                                       [ ok ]
implicitly enabling luapattern returncodes_matcher for rule nil
implicitly enabling luapattern returncodes_matcher for rule nil
 * Starting rspamd ...
[...]

PR #4785, opened 2024-01-20
[Fix] Replace comment types

arkamar:
Rspamd fails to start with lua 5-1 without this change.

Bug: https://bugs.gentoo.org/922522
Fixes: b189c9fea633 ("[Minor] lua_maps docs: apply formatting")
Fixes: https://github.com/rspamd/rspamd/issues/4784

### DIFF
--- a/lualib/lua_maps.lua
+++ b/lualib/lua_maps.lua
@@ -561,9 +561,9 @@ exports.rspamd_maybe_check_map = rspamd_maybe_check_map
 --     optional = true,
 --   }
 -- }
--- --[[ Then this function will look for opts.my_map parameter and try to replace it's with
--- a map with the specific type, description but not failing if it was empty.
--- It will also set options.my_map_orig to the original value defined in the map --]]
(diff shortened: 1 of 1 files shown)


arkamar:
Hmmm, I am not sure if the fix is correct as it most probably breaks docs formatting, right? Anyway, nested comments are not allowed, see [3rd point, section 7.1, lua 5.1 manual](https://www.lua.org/manual/5.1/manual.html#7.1). But it seems we can use `[=[` syntax in outer comments to fix this. What do you think?

fatalbanana:
If we would double up those comments it would still be rendered as a comment in docs (`./doc/doxydown/doxydown.pl -l lua < lualib/lua_maps.lua`).

~~~diff
+-- -- Then this function will look for opts.my_map parameter and try to replace it's with
~~~

arkamar:
@fatalbanana good idea, thanks! I did the proposed change, plus one small typo fix.

PR #4789, opened 2024-01-24
[Test] statistics: per-user

fatalbanana:


vstakhov:
I think it should work now.

Issue #4793, opened 2024-01-27
Kubernetes - History Lost on Restart


cfis:
### Prerequisites

* [x ] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
When running rspamd on kubernetes, every time its pod it is restarted history "appears" to be lost although it is stored in a separate Redis container. This is because rspamd incorporates the hostname into the history key. For example one run might be `rs_historydocker-mailserver-59cc9899d4-xmmgd_zst` while the next could be `rs_historydocker-mailserver-677cb6cd79-4xn48_zst`. 

The history is not actually lost, it still in Redis, but they key has been orphaned since rspamd loses track of it.

Note this happens in docker too, see https://github.com/Mailu/Mailu/pull/1837.

Any chance this could be changed? Unlike docker, Kubernetes does not let you specify the host name. So some ideas:

* Don't use hostname in the key
* Add an ENV variable like RSPAMD_HISTORY_HOSTNAME that can be optionally set or RSPAMD_HISTORY_KEY
* Query redis for an existing key and reuse it
* Etc.

### Steps to Reproduce

1. Deploy rspamd in Kubernetes. Get an email, look at the history in the UI
2. Restart the pod
3. Go back to the UI, there will no longer by any records under the history tab

**Expected behavior**
History is not lost since it still actually exists in a redis instance running in the cluster

### Versions
rspamd --version
Rspamd daemon version 3.7.5

k0s version
v1.28.5+k0s.0

Container OS is:
FROM docker.io/debian:12-slim AS stage-base

Host is:
cat /etc/redhat-release
Fedora release 39 (Thirty Nine)

Host CPU:
13th Gen Intel(R) Core(TM) i9-13900T

PR #4801, opened 2024-01-29
[Feature] Allow to add templates to redis history prefix

vstakhov:
We allow two things for now - `HOSTNAME` and `COMPRESS`. Hence, the default prefix will be the same, but the configuration default is now different to match the existing behaviour: `rs_history{{HOSTNAME}}{{COMPRESS}}`.

Issue: #4793
Closes: #4793


cfis:
Cool - thanks!

PR #4803, opened 2024-01-31
[Minor] Add more returnbits to surbl configuration

fatalbanana:


fatalbanana:
I wonder about scoring but I see no hits on these symbols in my own traffic ...

fatalbanana:
> I wonder about scoring but I see no hits on these symbols in my own traffic ...

I've seen it in combination with other bits. Maybe this is fine as it is ...

PR #4825, opened 2024-02-21
[Fix] Only run all the tests if luajit is enabled

a16bitsysop:
lua tests in `rspamd-test` fail with missing modules if luajit is not used


vstakhov:
I'm not sure that it's a good approach. Tests should work fine with plain Lua assuming that `lua-ffi` is installed.

a16bitsysop:
ah okay, it looks like that package is missing for alpine.  I will investigate further.

PR #4857, opened 2024-03-05
[Project] Optimization of userdata hashing

vstakhov:
Lua userdata class tag checking consumes a significant portion of CPU time, and there is a way for improvement. Instead of using static strings like "rspamd{task}" we can define a static constant, so we can just use it's address as a hash table key and avoid expensive hashing. It also simplifies the architecture as we have the single definition of each class being used.


vstakhov:
Drone, come on, lazy bastard...

PR #4888, opened 2024-03-24
Implement DNSBL checks for Message-ID RHS

twesterhever:
This merge request adds the functionality to the RBL plugin to check the right hand side of the `Message-ID` header against DNSBLs as well. Sometimes, the FQDN present in this header differs from any other FQDN in a messages' headers, caused by sloppy spam setups and the like.

Since a `Message-ID` header must only be present once per message, enabling this DNSBL check by default will add exactly one additional DNS query per configured DNSBL per message. Therefore, this merge request proposes to do so by default for all DNSBL operators that commonly list domains; Spamhaus, for example, explicitly recommends doing so [here](https://docs.spamhaus.com/datasets/docs/source/40-real-world-usage/smtp/020-SMTP-Checks.html#headers-checks).

### DIFF
--- a/conf/modules.d/rbl.conf
+++ b/conf/modules.d/rbl.conf
@@ -205,7 +205,7 @@ rbl {
     "SURBL_MULTI" {
       ignore_defaults = true;
       rbl = "multi.surbl.org";
-      checks = ['emails', 'dkim', 'helo', 'rdns', 'replyto', 'mid', 'urls'];
+      checks = ['emails', 'dkim', 'helo', 'rdns', 'replyto', 'urls'];
       emails_domainonly = true;
       exclude_users = false;
 
@@ -241,7 +241,7 @@ rbl {
     "URIBL_MULTI" {
       ignore_defaults = true;
       rbl = "multi.uribl.com";
-      checks = ['emails', 'dkim', 'helo', 'rdns', 'replyto', 'mid', 'urls'];
+      checks = ['emails', 'dkim', 'helo', 'rdns', 'replyto', 'urls'];
       emails_domainonly = true;
       exclude_users = false;
 
@@ -274,7 +274,7 @@ rbl {
       ignore_defaults = true;
       rbl = "dbl.spamhaus.org";
       no_ip = true;
-      checks = ['emails', 'dkim', 'helo', 'rdns', 'replyto', 'mid', 'urls'];
+      checks = ['emails', 'dkim', 'helo', 'rdns', 'replyto', 'urls'];
       emails_domainonly = true;
       exclude_users = false;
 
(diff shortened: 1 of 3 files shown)


vstakhov:
A quick question: why cannot we use selector for this purpose?

twesterhever:
> A quick question: why cannot we use selector for this purpose?

Oh, I took the `Reply-To` RBL check as an example. Happy to rework this using selectors, if you prefer.

vstakhov:
I think that the selectors usage is more appreciated here.

twesterhever:
> I think that the selectors usage is more appreciated here.

Roger that, I'll rework this PR accordingly.

twesterhever:
> I think that the selectors usage is more appreciated here.

Roger that, but I am afraid I need some help with this: Having worked through https://rspamd.com/doc/configuration/selectors.html and https://rspamd.com/doc/modules/rbl.html#selectors, it is not really clear to me how I must rework this PR exactly. In particular, I fail to see how I can go for selector usage in `rbl.conf` only, but without losing the information that a domain was extracted from the `Message-ID` header, or introducing new DNSBL symbols.

Any guidance on this would be appreciated. TIA! :-)

fatalbanana:
We could add something like:

~~~
selector = 'header(Message-Id).regexp("@([^\.]+\.[^>]+)").last'
~~~

to configuration rather than modifying `checks`.

twesterhever:
Above all, apologies for my belated response, and thanks for the hint. I was able to do the requested changes, and entirely go for selectors. However, I noticed that this will change the symbol hits. Before, they would look like this:
```
Symbol: DBL_SPAM (6.50)[getnada.com:mid]
```
When using selectors without `symbols_prefixes`, it will look like this:
```
Symbol: DBL_SPAM (6.50)[getnada.com:selector]
```
Without looking at the source code, it is not obvious which FQDN in a message triggered a DNSBL hit. Adding a symbols prefix is possible, but that will cause DNSBL hits based on Message-ID RHS to go under different symbols (`DBL_SPAM_MID` or something like that), and I am not sure if this is a good idea.

I'll push the selectors changes to this branch regardless. However, personally, I'd prefer the `checks`-based implementation, to keep DNSBL symbol hits transparent and avoid discrepancies from other DNSBL hits, such as PTRs or HELOs. :-)

fatalbanana:
> Without looking at the source code, it is not obvious which FQDN in a message triggered a DNSBL hit.

We could use something like

```
selector = {
  mid = 'header...';
}
```

vstakhov:
@fatalbanana is it the existing feature or do we need to implement that?

fatalbanana:
> @fatalbanana is it the existing feature or do we need to implement that?

It's an existing feature

fatalbanana:
> > @fatalbanana is it the existing feature or do we need to implement that?
> 
> It's an existing feature

... It should probably documented (it's not)

twesterhever:
Ah, _now_ I finally got the selectors usage for this. Apologies for my belated understanding; PR https://github.com/rspamd/rspamd/pull/4955 fixes the selector usage to preserve the Message-ID information for DNSBL hits.

PR #4895, opened 2024-03-28
Upgraded replies and known senders modules

LeftTry:
Added local and global replies sets to store verified senders.
Added functional to the known senders module to check if sender is verified.


src/plugins/lua/known_senders.lua
```diff
@@ -53,6 +53,11 @@ local settings = {
   symbol = 'KNOWN_SENDER',
   symbol_unknown = 'UNKNOWN_SENDER',
   redis_key = 'rs_known_senders',
+  sender_prefix = 'rsrk',
+  rsrk_privacy = false,
+  rsrk_privacy_alg = 'blake2',
+  rsrk_privacy_prefix = 'obf',
+  rsrk_privacy_length = 16,
 }
```
vstakhov, line 61: I don't like the idea of prefixing these settings with `rsrk` - it looks obscure. Maybe it is better to get the meaningful prefix, such as `reply_sender_privacy` for these options?

src/plugins/lua/known_senders.lua
```diff
@@ -72,6 +77,23 @@ local function make_key(input)
...
 
+local function make_key_replies(goop, sz, prefix)
+  local h = rspamd_cryptobox_hash.create()
+  h:update(goop)
+  local key
+  if sz then
+    key = h:base32():sub(1, sz)
+  else
+    key = h:base32()
+  end
```
vstakhov, line 88:
```suggestion
  local key = prefix or '' .. h:base32():sub(1, sz)
```

As `string.sub('hui', 1, nil) == 'hui'`

src/plugins/lua/known_senders.lua
```diff
@@ -72,6 +77,23 @@ local function make_key(input)
...
+
+  if prefix then
+    key = prefix .. key
+  end
```
vstakhov, line 92: Also not needed.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +219,57 @@ local function known_senders_callback(task)
   end
 end
 
+local function check_known_incoming_mail_callback(task)
+  local sender = task:get_reply_sender()
+  if not sender then
+    rspamd_logger.errx(task, 'Couldn\'t get sender')
```
vstakhov, line 225: It should not be logged as an error, maybe better just add a `info` or even `debugm` here.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +219,57 @@ local function known_senders_callback(task)
...
+    list_of_senders = data
+    rspamd_logger.infox(task, 'Successfully got list: %s of verified senders', data)
+
+    lua_util.debugm(N, task, 'List of senders: %s', list_of_senders)
+
+    -- searching for sender in global replies set
+    if list_of_senders then
+      for _, sndr in ipairs(list_of_senders) do
+        if sndr == sender then
+          task:insert_result('INC_MAIL_KNOWN', 1.0,
```
vstakhov, line 254: `INC_MAIL_KNOWN` must be configurable or set as a static constant. Reason: what if we want to change this symbol - it's not good to grep all it's mentions.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +219,57 @@ local function known_senders_callback(task)
...
+                  string.format('Incoming mail and it\'s sender is known'))
```
vstakhov, line 255: I would suggest to add `sender` name here, or at least remove useless `format` invocation for a string literal.

src/plugins/lua/known_senders.lua
```diff
@@ -230,6 +306,13 @@ if opts then
       augmentations = { string.format("timeout=%f", redis_params.timeout or 0.0) }
     })
 
+    rspamd_config:register_symbol({
+      name = 'INC_MAIL_KNOWN',
```
vstakhov, line 310: Same - must be set in one place.

src/plugins/lua/replies.lua
```diff
@@ -44,6 +45,10 @@ local settings = {
   cookie_is_pattern = false,
   cookie_valid_time = '2w', -- 2 weeks by default
   min_message_id = 2, -- minimum length of the message-id header
+  rsrk_privacy = false,
+  rsrk_privacy_alg = 'blake2',
+  rsrk_privacy_prefix = 'obf',
+  rsrk_privacy_length = 16,
```
vstakhov, line 51: Same - readable options are preferred.

src/plugins/lua/replies.lua
```diff
@@ -69,7 +74,7 @@ local function replies_check(task)
   local in_reply_to
   local function check_recipient(stored_rcpt)
     local rcpts = task:get_recipients('mime')
-
+    rspamd_logger.infox(task, 'recipients: %s', rcpts)
```
vstakhov, line 77: It should be a debugm not info.

src/plugins/lua/replies.lua
```diff
@@ -81,7 +86,12 @@ local function replies_check(task)
         return rcpt.addr or ''
       end, rcpts)) then
         lua_util.debugm(N, task, 'reply to %s validated', in_reply_to)
-        return true
+
+        --storing only addr of rcpt
+        for i = 1, #rcpts, 1 do
```
vstakhov, line 91: Why do we need a step here:
```suggestion
        for i = 1, #rcpts do
```

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+
+    -- getting last score of recipients
+    local function redis_zrange_cb(err, data)
+      if err ~= nil then
+        rspamd_logger.errx(task,
+                'redis_zrange_cb error when reading zrange withscores from global replies set with error: %s', err)
+        return
+      end
+      last_score = tonumber(data[#data])
+      rspamd_logger.infox(task, 'last score %s of global replies set was received', last_score)
```
vstakhov, line 142: `debugm` please

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+
+      -- if last score wasn't found
+      if last_score == -1 or last_score == nil then
+        rspamd_logger.errx(task, 'have not found any senders in global replies set, considering last score as 0')
```
vstakhov, line 146: `debugm` please

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+    local task_time = task:get_timeval(true)
+
+    local params = {}
+    -- making params out of recipients list for replies set
+    local j = 1
+    for i = 1, #recipients * 2, 2 do
+      table.insert(params, i, tostring(task_time))
+      table.insert(params, i + 1, tostring(recipients[j]))
+      j = j + 1
+    end
```
vstakhov, line 183:
```suggestion
    -- making params out of recipients list for replies set
    local task_time_str = tostring(task_time)
    for rcpt in ipairs(recipients) do
      table.insert(params, task_time_str)
      table.insert(params, tostring(rcpt))
    end
```

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+        rspamd_logger.errx(task, 'adding to %s failed with error: %s', sender_key, err)
+        return
+      end
+
+      table.remove(params, 1)
+
+      rspamd_logger.infox(task, 'added data: %s to sender: %s with code: %s', params, sender_key, data)
+
+      update_global_replies_set(params, sender_key)
+
```
vstakhov, line 203: Extra empty newlines.

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+    local function zadd_cb(err, data)
+      if err ~= nil then
+        rspamd_logger.errx(task, 'adding to %s failed with error: %s', sender_key, err)
+        return
+      end
+
+      table.remove(params, 1)
+
+      rspamd_logger.infox(task, 'added data: %s to sender: %s with code: %s', params, sender_key, data)
```
vstakhov, line 200: `debugm` please.

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+    local _, conn, _ = lua_redis.redis_make_request(task, -- making local replies set (sender - recipients)
+            redis_params, -- connect params
+            sender_key, -- hash key
+            true, -- is write
+            zadd_cb, --callback
+            'ZADD', -- command
+            params -- arguments
+    )
+    -- adding expiration to the local replies set
+    conn:add_cmd('EXPIRE', { sender_key, tostring(math.floor(settings['expire'] * 1000)) })
```
vstakhov, line 215: Not sure why do we multiply it by 1000. Time values in Rspamd are usually stored as plain ucl variables, such as `1h` or `1min`.

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
   end
 
   local function redis_get_cb(err, data, addr)
     if err ~= nil then
       rspamd_logger.errx(task, 'redis_get_cb error when reading data from %s: %s', addr:get_addr(), err)
       return
     end
-    if data and type(data) == 'string' and check_recipient(data) then
+    local recipients = check_recipient(data)
+    if data and type(data) == 'string' and recipients then
```
vstakhov, line 228:
```suggestion
    if type(data) == 'string' and recipients then
```

No need to check `data` for nil, as `type` does that for you.

src/plugins/lua/replies.lua
```diff
@@ -318,7 +449,7 @@ if opts then
       group = "replies"
     })
     rspamd_config:register_symbol({
-      name = settings['symbol'],
+      name = 'REPLY',
```
vstakhov, line 452: Why?

vstakhov (changes requested):
It looks very good overall, thanks! Some corrections are still needed though.

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+  local function add_to_replies_set(recipients)
+    local sender = task:get_reply_sender()
+
+    local task_time = task:get_timeval(true)
+
+    local params = {}
+    -- making params out of recipients list for replies set
+    local j = 1
+    for i = 1, #recipients * 2, 2 do
+      table.insert(params, i, tostring(task_time))
```
fatalbanana, line 180: This `tostring` could be outside of the loop

src/plugins/lua/replies.lua
```diff
@@ -91,16 +101,133 @@ local function replies_check(task)
...
+      table.insert(params, i + 1, tostring(recipients[j]))
+      j = j + 1
+    end
+
+    local sender_string = lua_util.maybe_obfuscate_string(tostring(sender), settings, settings.sender_prefix)
+    local sender_key = make_key(sender_string:lower(), 8)
+
+    lua_util.debugm(N, task, 'Adding  recipients %s to sender %s local replies set', params, sender_key)
```
fatalbanana, line 188: Double-whitespace; `params` contains weird things?

fatalbanana (commented):
Could we get a human-readable description of the changes please? I heard something about it but can't particularly remember ...

src/plugins/lua/known_senders.lua
```diff
@@ -36,7 +36,7 @@ known_senders {
   # Maximum time to live (when not using bloom filters)
   max_ttl = 30d;
   # Use bloom filters (must be enabled in Redis as a plugin)
-  use_bloom = false;
+  use_bloom = falsereply_sender_;
```
vstakhov, line 39: This wasn't intentional, I suppose.

src/plugins/lua/known_senders.lua
```diff
@@ -52,12 +52,13 @@ local settings = {
...
   redis_key = 'rs_known_senders',
   sender_prefix = 'rsrk',
-  rsrk_privacy = false,
-  rsrk_privacy_alg = 'blake2',
-  rsrk_privacy_prefix = 'obf',
-  rsrk_privacy_length = 16,
+  reply_sender_privacy = false,
+  reply_sender_privacy_alg = 'blake2',
+  reply_sender_privacy_prefix = 'obf',
+  reply_sender_privacy_length = 16,
```
vstakhov, line 61: 👍

src/plugins/lua/known_senders.lua
```diff
@@ -251,8 +242,8 @@ local function check_known_incoming_mail_callback(task)
     if list_of_senders then
       for _, sndr in ipairs(list_of_senders) do
         if sndr == sender then
-          task:insert_result('INC_MAIL_KNOWN', 1.0,
-                  string.format('Incoming mail and it\'s sender is known'))
+          task:insert_result(settings.symbol_check_mail, 1.0,
+                  string.format('Incoming mail with sender %s is known', sndr))
```
vstakhov, line 246: `          task:insert_result(settings.symbol_check_mail, 1.0, sndr)`

Symbol options should be kept short where possible.

src/plugins/lua/replies.lua
```diff
@@ -91,16 +91,127 @@ local function replies_check(task)
...
+    table.insert(params, 1, 'rsrk_verified_recipients')
+
+    lua_redis.redis_make_request(task, -- making global replies set for verified recipients
+            redis_params, -- connect params
+            sender_key, -- hash key
+            true, -- is write
+            zadd_global_set_cb, --callback
+            'ZADD', -- command
+            params -- arguments
+    )
```
vstakhov, line 117: Hum, I also don't see any limits for this key, it must have not more than X (configurable parameter - default should be like 20-30 at max) and `zrembyrank` command must be emitted after any successful write.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +210,57 @@ local function known_senders_callback(task)
...
+
+  -- making sender key
+  lua_util.debugm(N, task, 'Sender: %s', sender)
+
+  local sender_string = lua_util.maybe_obfuscate_string(tostring(sender), settings, settings.sender_prefix)
+  local sender_key = make_key_replies(sender_string:lower(), 8)
+
+  lua_util.debugm(N, task, 'Sender key: %s', sender_key)
+
+  local list_of_senders = {}
```
vstakhov, line 228: What is the purpose of this variable, counting that we just assign it to `data` in the callback afterwards?

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +210,57 @@ local function known_senders_callback(task)
...
+
+    list_of_senders = data
+    rspamd_logger.infox(task, 'Successfully got list: %s of verified senders', data)
+
+    lua_util.debugm(N, task, 'List of senders: %s', list_of_senders)
+
+    -- searching for sender in global replies set
+    if list_of_senders then
+      for _, sndr in ipairs(list_of_senders) do
+        if sndr == sender then
```
vstakhov, line 244: In Lua, this comparison is O(1), so this code makes sense. In other languages, it is better to convert `list_of_senders` to a set and check that value in set should the length of `list_of_senders` > around 50-100 elements in general. Not a change request, just FYI.

src/plugins/lua/replies.lua
LeftTry: Wdym I need to execute `zrembyrank` command after successful write? I just added recipients and I need to remove them or what? I don't really understand.

vstakhov: You need to maintain the number of elements in the lists. Otherwise, they will grow infinitely and blow up the Redis instance. I would suggest to use the redis scripts here: they can do zadd/zcard/zrembyrank/expire all at once in the atomic execution pipeline.

src/plugins/lua/replies.lua
```diff
@@ -94,18 +98,42 @@ local function replies_check(task)
...
+      local redis_script_zadd_global = [[
+      local global_size = redis.call('ZCARD', KEYS[1])
+      if global_size > {= mgs =} then
+        redis.call('ZREMRANGEBYRANK', KEYS[1], '{= mgs =}', global_size)
+      end
+      ]]
+
+      local set_script_zadd_global = lua_util.jinja_template(redis_script_zadd_global,
+              { mgs = settings['max_global_size'] })
+      local zadd_id = lua_redis.add_redis_script(set_script_zadd_global, redis_params)
```
vstakhov, line 128: This code part must be done in the configuration part of the module, not in the callback. I would also suggest to remove `ZADD` invocation and do all things: `ZCARD` -> `ZREM` -> `ZADD` in one Redis script.

src/plugins/lua/replies.lua
```diff
@@ -180,6 +207,16 @@ local function replies_check(task)
...
 
+    local global_key = make_key(settings.sender_key_global, settings.sender_key_size, settings.sender_prefix)
+
+    local function zrem_cb(err, data)
+      if err ~= nil then
+      rspamd_logger.errx(task, 'failed to remove extra recipients from %s replies set with error: %s', sender_key, err)
+      return
+      end
+      rspamd_logger.infox(task, 'if necessary, additional recipients were deleted from the %s replies set', sender_key)
+    end
```
vstakhov, line 218: NIT: indentation issues.

src/plugins/lua/replies.lua
```diff
@@ -200,7 +255,7 @@ local function replies_check(task)
             params -- arguments
     )
     -- adding expiration to the local replies set
-    conn:add_cmd('EXPIRE', { sender_key, tostring(math.floor(settings['expire'])) })
+    --conn:add_cmd('EXPIRE', { sender_key, tostring(math.floor(settings['expire'])) })
```
vstakhov: We still need expirty - maybe in Redis script?

vstakhov (changes requested):
It gets to a better shape now, some few bits are still needed.

src/plugins/lua/replies.lua
```diff
@@ -91,16 +97,146 @@ local function replies_check(task)
...
+        redis.call('ZREMRANGEBYRANK', KEYS[1], '{= mls =}', local_replies_set_size)
+      end
+      -- adding recipients for the local replies set
+      redis.call('ZADD', KEYS[1], ARGV[1], ARGV[2])
+      -- setting expire for local replies set
+      redis.call('EXPIRE', KEYS[1], tostring(math.floor('{= exp =}')))
+    ]]
+    local set_script_zadd_local = lua_util.jinja_template(redis_script_zadd_local,
+      { exp = settings['expire'], mls = settings['max_local_size'] })
+    table.insert(script_ids, lua_redis.add_redis_script(set_script_zadd_local, redis_params))
```
vstakhov, line 128: I'd recommend to have `script_ids` being indexed not by a numeric index but by some reasonable string index.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +212,55 @@ local function known_senders_callback(task)
...
+
+  lua_util.debugm(N, task, 'Making redis request to global replies set')
+  lua_redis.redis_make_request(task,
+          redis_params, -- connect params
+          sender_key, -- hash key
+          false, -- is write
+          redis_zrange_cb, --callback
+          'ZRANGE', -- command
+          { global_key, '0', '-1' } -- arguments
+  )
```
vstakhov, line 260: Why do we need this all logic in the first place? Why not just check `global_key` for `sender`? Getting all values does not make a lot of sense to me.

src/plugins/lua/replies.lua
```diff
@@ -91,16 +97,146 @@ local function replies_check(task)
...
+    lua_util.debugm(N, task, 'Getting recipients withscores from global replies set to get last score')
+
+    -- getting scores of recipients in global replies set
+    lua_redis.redis_make_request(task,
+            redis_params,
+            sender_key,
+            false,
+            redis_zrange_cb,
+            'ZRANGE',
+            {global_key, '0', '-1', 'WITHSCORES'}
```
vstakhov, line 182: I don't understand this logic as well: why do we get **all** the keys for each check?

src/plugins/lua/known_senders.lua
vstakhov: Also, I don't see any logic to verify **local** senders set. We need to check both, and the symbols might be different for each of the keys - local and global.

vstakhov (changes requested):
I have still some questions about the checking logic, please ensure that it is correct.

src/plugins/lua/known_senders.lua
LeftTry: I don't quite understand what you mean by checking `global_key` and verifying the local set. What should be verified in the local set and why do I need to check the `global_key` for the `sender`? This logic was created in order to get the highest score in the `global set` and sum it with the scores of incoming recipients to be sure that their scores are higher than those of new ones. It does not check anything. So, could you please explain what you mean?

vstakhov: https://redis.io/docs/latest/commands/zscore/ ?

src/plugins/lua/known_senders.lua
```diff
@@ -212,6 +213,67 @@ local function known_senders_callback(task)
...
+  local recipients = task:get_recipients('mime')
+
+  for i, rcpt in pairs(recipients) do
+    if i ~= #recipients then
+      lua_redis.redis_make_request(task,
+              redis_params,
+              sender_key,
+              false,
+              redis_zscore_local_cb,
+              'ZSCAN',
```
vstakhov, line 261: I really don't understand why do we need `ZSCAN` here. It works as O(N) and I get no glue why can't we just use `ZSCORE <sender_key> <rcpt.addr>`.

vstakhov: Erm, we don't even need this loop as well - there is `ZMSCORE` command in Redis. Probably, it will not work on ancient Redis though. We also need to limit number of recipients to be checked (10-15 at max). Redis calls are costly, so we cannot overuse them. If `ZMSCORE` is not good from the terms of portability, we can emulate it via a simple Redis script that calls for `ZSCORE` several times. I would anyway avoid extra calls to Redis and, for sure, O(N) commands such as `ZSCAN ... PATTERN` unless absolutely needed.

vstakhov (changes requested):

src/plugins/lua/replies.lua
```diff
@@ -91,16 +98,146 @@ local function replies_check(task)
...
+
+    local sender_string = lua_util.maybe_obfuscate_string(tostring(sender), settings, settings.sender_prefix)
+    local sender_key = make_key(sender_string:lower(), 8)
+
+    lua_util.debugm(N, task,
+            'Adding recipients %s to sender %s local replies set', recipients, sender_key)
+
+    local global_key = make_key(settings.sender_key_global, settings.sender_key_size, settings.sender_prefix)
+
+    configure_redis_scripts()
```
fatalbanana, line 203: Could we move this out of symbol callback similar to how `ratelimit` module is doing it?

src/plugins/lua/replies.lua
```diff
@@ -91,16 +98,146 @@ local function replies_check(task)
...
+
+      -- if last score wasn't found
+      if last_score == nil then
+        lua_util.debugm(N, task,
+                'have not found any senders in global replies set, considering last score as 0')
+        last_score = 0
+      end
+
+      -- updating params considering last score of existing sender
+      task_time = task_time + last_score
```
fatalbanana, line 169: Is there a reason we don't rather increment it by 1 each time?

Could we please get a more detailed description for this PR if possible to help me understand the intentions?

fatalbanana: Probably it is so newer data would be retained rather than older data, that's reasonable ...

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +214,98 @@ local function known_senders_callback(task)
...
+local function verify_local_replies_set(task)
+  local sender = task:get_reply_sender()
+  if not sender then
+    lua_util.debugm(N, task, 'Could not get sender')
+    return nil
+  end
+  local recipients = task:get_recipients('mime')
+
+  local sender_string = lua_util.maybe_obfuscate_string(tostring(sender), settings, settings.sender_prefix)
+  local sender_key = make_key_replies(sender_string:lower(), 8)
```
fatalbanana, line 226: Could we assign all keys some or other prefix please?

src/plugins/lua/replies.lua
```diff
@@ -91,16 +98,146 @@ local function replies_check(task)
...
+    lua_util.debugm(N, task, 'Adding recipients %s to global replies set', recipients)
+
+    local function zadd_global_set_cb(err, data)
+      if err ~= nil then
+        rspamd_logger.errx(task, 'failed to add recipients %s to global replies set with error: %s', recipients, err)
+        return
+      end
+      rspamd_logger.infox(task, 'added recipients %s to global set', recipients)
+    end
+    for _, rcpt in ipairs(recipients) do
```
fatalbanana, line 142: Might we avoid the need to call the script for each recipient?

LeftTry:
Maybe it is better to rewrite scripts to files so the code in the replies and known_senders plugins will look better?

src/plugins/lua/known_senders.lua
```diff
@@ -72,6 +82,49 @@ local function make_key(input)
...
+    local replies_recipients_addrs = {}
+    replies_recipients_addrs = ARGV
+    if replies_recipients_addrs ~= nil then
+
+      for _, rcpt in ipairs(replies_recipients_addrs) do
+        local score = redis.call('ZSCORE', KEYS[1], rcpt)
+
+        if type(score) == 'boolean' then
+          score = nil
+          table.insert(results, score)
```
fatalbanana, line 107: This is a no-op.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +250,93 @@ local function known_senders_callback(task)
...
+
+  local replies_sender_string = lua_util.maybe_obfuscate_string(tostring(replies_sender), settings, settings.sender_prefix)
+  local replies_sender_key = make_key_replies(replies_sender_string:lower(), 8)
+
+  local function redis_zscore_script_cb(err, data)
+    if err ~= nil then
+      rspamd_logger.errx(task, 'Could not verify %s local replies set %s', replies_sender_key, err)
+    end
+    if data[1] ~= 1 then
+      rspamd_logger.infox(task, 'Recipients was not verified')
```
fatalbanana, line 270: `were not`

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +250,93 @@ local function known_senders_callback(task)
...
+    else
+      rspamd_logger.infox(task, 'Recipients was verified')
```
fatalbanana, line 272: Could be moved out of the conditional to avoid duplication or removed if this is misplaced.

src/plugins/lua/replies.lua
```diff
@@ -34,42 +34,102 @@ local settings = {
...
+      end
+      local params = {}
+      params = ARGV
+      if params ~= nil then
+          local task_time = params[1]
+          table.remove(params, 1)
+
+          for _, rcpt in ipairs(params) do
+            -- adding recipients for the local replies set
+            redis.call('ZADD', KEYS[1], task_time, rcpt)
```
fatalbanana, line 114: We could add more arguments to `ZADD` rather than using `for` loop (but it's a little unpleasant to do).

src/plugins/lua/known_senders.lua
LeftTry: I don't quite understand what the problem is with this `if`. Could you please explain what you mean by duplication?

src/plugins/lua/known_senders.lua
```diff
@@ -72,6 +82,49 @@ local function make_key(input)
...
+local function configure_scripts(_, _, _)
+  -- script checks if given recipients are in the local replies set of the sender
+  local redis_zscore_script = [[
+    local results = {}
+    local replies_recipients_addrs = {}
+    replies_recipients_addrs = ARGV
+    if replies_recipients_addrs ~= nil then
+
+      for _, rcpt in ipairs(replies_recipients_addrs) do
+        local score = redis.call('ZSCORE', KEYS[1], rcpt)
```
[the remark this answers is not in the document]
LeftTry: This may not work with previous versions of redis and does not work on ubuntu. So I decided to keep this in the for loop instead of using the 'ZMSCORE' command.

src/plugins/lua/known_senders.lua
fatalbanana: I was reading the same log message on both of these lines somehow ....

src/plugins/lua/known_senders.lua
```diff
@@ -72,6 +82,43 @@ local function make_key(input)
...
+      end
+      -- first number in return statement is stands for the success/failure code
+      -- where success code is 1 and failure code is 0
+      return 1
+    else
+    -- 0 is a failure code
+      return 0
+    end
+  ]]
+  local zscore_script = lua_util.jinja_template(redis_zscore_script, {  })
```
vstakhov, line 117: Why do we need `jinja_template` with no environment here?

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +244,94 @@ local function known_senders_callback(task)
...
+  local replies_sender_string = lua_util.maybe_obfuscate_string(tostring(replies_sender), settings, settings.sender_prefix)
+  local replies_sender_key = make_key_replies(replies_sender_string:lower(), 8)
+
+  local function redis_zscore_script_cb(err, data)
+    if err ~= nil then
+      rspamd_logger.errx(task, 'Could not verify %s local replies set %s', replies_sender_key, err)
+    end
+    rspamd_logger.infox(task, 'DATA: %s', data)
+    if data ~= 1 then
+      rspamd_logger.infox(task, 'Recipients were not verified')
```
vstakhov, line 265: `debugm` here please.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +244,94 @@ local function known_senders_callback(task)
...
+      return
+    end
+    rspamd_logger.infox(task, 'Recipients were verified')
```
vstakhov, line 268: Again, it is debug.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +244,94 @@ local function known_senders_callback(task)
...
+    task:insert_result(settings.symbol_check_mail_local, 1.0, replies_sender_key)
+  end
+
+  local replies_recipients_addrs = {}
+
+  -- assigning addresses of recipients for params and limiting number of recipients to be checked
+  for i, rcpt in ipairs(replies_recipients) do
```
vstakhov, line 275:
```suggestion
  local maxn = math.min(settings.max_recipients, #replies_recipients)
  for i = 1, maxn do
```

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +244,94 @@ local function known_senders_callback(task)
...
+
+  local function redis_zscore_global_cb(err, data)
+    if err ~= nil then
+      rspamd_logger.errx(task, 'Couldn\'t find sender %s in global replies set. Ended with error: %s', replies_sender, err)
+      return
+    end
+
+    --checking if zcore have not found score of a sender
+    if data ~= nil and data ~= '' and type(data) ~= 'userdata' then
+      rspamd_logger.infox(task, 'Sender: %s verified. Output: %s', replies_sender, data)
```
vstakhov, line 313: It is also debug.

src/plugins/lua/known_senders.lua
```diff
@@ -197,6 +244,94 @@ local function known_senders_callback(task)
...
+      task:insert_result(settings.symbol_check_mail_global, 1.0, replies_sender)
+    else
+      rspamd_logger.infox(task, 'Sender: %s was not verified', replies_sender)
```
vstakhov, line 316: It is debug.

src/plugins/lua/replies.lua
```diff
@@ -34,42 +34,107 @@ local settings = {
...
+      data = redis.call('ZRANGE', KEYS[1], '-1', '-1', 'WITHSCORES')
+      local score = tonumber(data[#data])
+      if score == nil then
+        score = 0
+      end
 
-  return key
+      local global_size = redis.call('ZCARD', KEYS[1])
+      if global_size > {= mgs =} then
+        redis.call('ZREMRANGEBYRANK', KEYS[1], '{= mgs =}', global_size)
```
vstakhov, line 82: We can just call this function all the time in something like `ZREMRANGEBYRANK key 0 -{= mgs =}` and remove `ZCARD` call completely, as we don't need it at all. Please look how other modules utilise this function as an example.

src/plugins/lua/replies.lua
```diff
@@ -34,42 +34,107 @@ local settings = {
...
+        local params = {}
+        --table.insert(params, KEYS[1])
+        for _, rcpt in ipairs(recipients_addrs) do
+          -- score is using here as a one-based numeration just like in lua
+          score = score + 1
+          table.insert(params, score)
+          table.insert(params, tostring(rcpt))
+        end
+        -- adding recipients to the global replies set
+        redis.call('ZADD', KEYS[1], unpack(params))
```
vstakhov, line 96: I'm sorry, but I'm lost in this code. It looks incorrect to me I'm afraid. What we need to do is to call `ZINCRBY key recipient 1` for each recipient. We must also remove `ZRANGE` call, as it has no point here. If you want to have not LFU (least frequently used) logic but LRU (least recently used), then just use task timestamp as a value, hence, replacing all `ZINCRBY` with `ZADD key recipient <task_timestamp>`. However, LFU logic seems to be more preferrable here to my opinion (I'd also address this question to @fatalbanana).

src/plugins/lua/replies.lua
```diff
@@ -34,42 +34,107 @@ local settings = {
...
+  local set_script_zadd_global = lua_util.jinja_template(redis_script_zadd_global,
+          { mgs = settings['max_global_size'] })
+  rspamd_logger.debugm(N, rspamd_config, 'added add to global replies set script %s', set_script_zadd_global)
+  global_replies_set_script =  lua_redis.add_redis_script(set_script_zadd_global, redis_params)
+
+  local redis_script_zadd_local = [[
+      local local_replies_set_size = redis.call('ZCARD', KEYS[1])
+      if local_replies_set_size > {= mls =} then
+        redis.call('ZREMRANGEBYRANK', KEYS[1], '{= mls =}', local_replies_set_size)
+      end
```
vstakhov, line 108: Same here: please do expiration properly.

src/plugins/lua/replies.lua
```diff
@@ -34,42 +34,107 @@ local settings = {
...
+          local task_time = given_params[1]
+          table.remove(given_params, 1)
+          -- passing_params is a table that will be passed to the redis
+          local passing_params = {}
+          for _, rcpt in ipairs(given_params) do
+            -- adding recipients for the local replies set
+            table.insert(passing_params, task_time)
+            table.insert(passing_params, rcpt)
+          end
+          redis.call('ZADD', KEYS[1], unpack(passing_params))
```
vstakhov, line 120: I see LRU logic here, good. Probably, we can use LFU for a global table and LRU here for a personal one, that looks sensible.

src/plugins/lua/replies.lua
```diff
@@ -34,42 +34,107 @@ local settings = {
...
+
+          -- setting expire for local replies set
+          redis.call('EXPIRE', KEYS[1], tostring(math.floor('{= exp =}')))
+      end
+    ]]
+  local set_script_zadd_local = lua_util.jinja_template(redis_script_zadd_local,
+          { exp = settings['expire'], mls = settings['max_local_size'] })
```
vstakhov, line 127:
```suggestion
          { exp = settings.expire, mls = settings.max_local_size })
```

NIT: use `.` to access constant literal named elements in a table.

vstakhov: Also, neither `mls` nor `exp` do not look like a good variable names to me. Why not use `expire` and `max_local_size` - it looks much more readable.

src/plugins/lua/known_senders.lua
LeftTry: In which cases should I use `debugm` instead of `infox`, so that I know this and do not make such a mistake?

vstakhov: In all cases aside of the cases when there is some important information that must attract user's attention.

src/plugins/lua/replies.lua
vstakhov: Then use another Redis Lua script for these purposes.

LeftTry: Sorry, what do you mean by using another Redis Lua script? What is the problem with using it inside this redis script?

LeftTry:
Due to the fact that the logic of adding recipients to global replies set is still being discussed, it has been left unchanged.

vstakhov (approved):
LGTM, good job! @fatalbanana could you also please take one last pass on it?

src/plugins/lua/known_senders.lua
```diff
@@ -230,6 +360,20 @@ if opts then
...
+      type = 'normal',
+      callback = verify_local_replies_set,
+      score = 1.0
+    })
+
+    rspamd_config:register_symbol({
+      name = settings.symbol_check_mail_global,
+      type = 'normal',
+      callback = check_known_incoming_mail_callback,
+      score = 1.0
```
fatalbanana, line 374: We should rather have negative scores for these symbols I suppose

vstakhov: We can set it in the config later maybe?

LeftTry: Which negative scores should we have?

moisseev:
After this change configdump fails if the known_senders module is enabled.

```hcl
# rspamadm configdump known_senders
cannot parse known_senders options: expected { "symbol_unknown" = optional type "string", "redis_key" = optional type "string", "domains" = type "string", array of type "string", or { "upstreams" = optional type "string", or array of type "string", "data" = optional array of type "string", "selector_alias" = optional type "string", "url" = optional type "string", or array of type "string", "name" = optional type "string", "timeout" = type "number", "description" = optional type "string" }, or { "external" = equivalent to true, "method" = "body", "header", or "query", "encode" = optional "json", or "messagepack", "timeout" = 

--SNIP--

equivalent to true, "method" = "body", "header", or "query", "encode" = optional "json", or "messagepack", "timeout" = optional type "number", or type "string", "backend" = type "string" }, "username" = Username, "sentinel_password" = Sentinel password, "sentinel_watch_time" = Sentinel watch time, "max_senders" = optio
*** Section known_senders ***
domains = "https://maps.rspamd.com/freemail/free.txt.zst";
use_bloom = false;
enabled = true;
max_senders = 100000;
max_ttl = 2592000;

*** End of section known_senders ***
```

Issue #4908, opened 2024-04-05
[BUG] greylist doesn't use dynamic/webui score threshold


Beuc:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
The greylist module appears to only use the static configuration score threshold, and ignore the one redefined in the webui.

### Steps to Reproduce

1. Keep default configuration (`actions.conf` -> `greylist = 4;`)
2. Lower greylist in the webui#configuration to `1`.
3. Check that `/var/lib/rspamd/rspamd_dynamic` is updated:  `"name": "greylist", "value": 1.0`
4. Notice that e-mails with a score around 2.x don't get greylisted (no action)
5. Modify logging in greylist.lua to print out the actual values (l.331): `rspamd_logger.infox(task, 'Score too low (score %s < greylist_min_score %s) - skip greylisting', score, settings.greylist_min_score)`
6. Send test mail.
7. Check the logs: `Score too low (score -0.20000000000000004 < greylist_min_score 4) - skip greylisting`
8. Notice that the check use the file configuration score (4) not the one redefined in the webui (1).

**Expected behavior**
greylist takes the score threshold overriden from the webui into account, here `1.0`.

### Versions

```
# rspamd --version
Rspamd daemon version 3.4
# lsb_release -a
Description:	Debian GNU/Linux 12 (bookworm)
```

### Additional Information

I have no greylist_min_score in my configuration, so I believe it's computed through:
```
    local greylist_threshold = rspamd_config:get_metric_action('greylist')
    if greylist_threshold then
      settings['greylist_min_score'] = greylist_threshold
    end
```
when loading the module.

[...]

PR #4910, opened 2024-04-05
Update greylisting threshold on each scan as it can be easily dynamic

vstakhov:
Issue: #4908
Closes: #4908


vstakhov:
Ok, so it's now proven by tests, thanks @fatalbanana !

PR #4912, opened 2024-04-08
Rework grow factor

vstakhov:
Currently, `grow_factor` setting is totally bogus:

* It is increased with every symbol with positive score being added
* As symbols are added in undefined order, the weight of symbol X might be different depending if there were different amount of **other** symbols being inserted before
* Symbols with low score (like 0.01) are as significant in grow factor calculations as symbols with high scores

This PR fixes all the issues above. `grow_factor` is just a value that will be used to multiply weights of **all** positive score symbols after the postfilters phase. This factor will be calculated by checking each individual symbol contribution to the final result, so symbols with 1/2 of rejection (or any other highest score metric threshold) score will multiply grow_factor by `1.0 + (grow_factor - 1.0) * 0.5`, and symbol with 0.1 of rejection threshold will increase grow_factor by `1.0 + (grow_factor - 1.0) * 0.1`. After all symbols are examined, the adjusted grow factor is applied to all symbols with positive scores.

This logic will provide consistent and reasonable (to me) result.

### DIFF
--- a/src/libmime/scan_result.c
+++ b/src/libmime/scan_result.c
@@ -231,7 +231,7 @@ insert_metric_result(struct rspamd_task *task,
 					 bool *new_sym)
 {
 	struct rspamd_symbol_result *symbol_result = NULL;
-	double final_score, *gr_score = NULL, next_gf = 1.0, diff;
+	double final_score, *gr_score = NULL, diff;
 	struct rspamd_symbol *sdef;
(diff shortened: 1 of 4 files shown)


fatalbanana (approved):
It looks good to me; it's easier to reason about relative to how it was and the results seem comparable; I don't find issue with it; will work on adding tests.

PR #4925, opened 2024-04-14
[Minor] Add rspamd filtering server to clickhouse

lucasRolff:
This is a minor improvement to add the rspamd filtering server to the `rspamd` clickhouse table.

It allows for (easier) filtering when doing queries, e.g. if you were to run separate inbound and outbound rspamd servers, and want to be able to query based on this.

### DIFF
--- a/src/plugins/lua/clickhouse.lua
+++ b/src/plugins/lua/clickhouse.lua
@@ -20,8 +20,10 @@ local lua_util = require "lua_util"
 local lua_clickhouse = require "lua_clickhouse"
 local lua_settings = require "lua_settings"
 local fun = require "fun"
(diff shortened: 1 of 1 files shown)


vstakhov:
That looks like a useful addition, however, I would prefer to keep the current structure of `rspamd` table as flat and simple as possible. The behaviour you propose could be achieved, for example, by using of `extra_columns` and selectors. Unfortunately, there is no pre-built selector extractor for the hostname but it should not be a big deal to add one (even cached).

lucasRolff:
I'll look into adding a selector instead, see if I can figure that out 👌 PR will probably arrive some time tomorrow (hopefully)

PR #4926, opened 2024-04-15
[Minor] Add selector to get rspamd_hostname

lucasRolff:
Instead of extending the clickhouse `rspamd` table directly, this allows using `rspamd_hostname` as a selector with `extra_columns`, such as:

```lua
extra_columns = {
        RspamdServer = {
                selector = "rspamd_hostname";
                type = "String";
        }
}
```

### DIFF
--- a/lualib/lua_selectors/extractors.lua
+++ b/lualib/lua_selectors/extractors.lua
@@ -17,6 +17,7 @@ limitations under the License.
 local fun = require 'fun'
 local meta_functions = require "lua_meta"
 local lua_util = require "lua_util"
+local rspamd_util = require "rspamd_util"
 local rspamd_url = require "rspamd_url"
 local common = require "lua_selectors/common"
 local ts = require("tableshape").types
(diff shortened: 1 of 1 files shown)


lucasRolff:
@vstakhov - I tried to see if there was a decent way of caching the `rspamd_util.get_hostname()` within rspamd, to then only read the hostname once until restarting rspamd (or cache it for 1 hour).

I see some config options use a mempool to do this (if I understand the code correctly :-D), but I'm unsure how to utilize the mempool to not create a new pool every time.

And caching it in Redis seemed slightly overkill.

Do you have any suggestions/pointers to how to possibly cache it "globally" within rspamd context? There's a bunch of places (mainly exporters) that simply use `rspamd_util.get_hostname()` directly currently, so if it could be cached overall within rspamd, it could possibly be changed across the exporters to use the cached value.

fatalbanana:
> Do you have any suggestions/pointers to how to possibly cache it "globally" within rspamd context?

We can do `local HOSTNAME = rspamd_util.get_hostname()` in the main body of this script and `return HOSTNAME` inside the method.

vstakhov:
Thank you!

Issue #4928, opened 2024-04-16
[BUG] Re-learning bayes not work properly after migration to LUA scripts in Redis


dragoangel:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * [X] Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * [X] Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * [X] Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * [ ] Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * [X] Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * [X] Checked that there is not already an experimental package or master branch

**Describe the bug**
Re-learning bayes not work properly after migration to LUA scripts in Redis due to missing action to swap flag in `learned_ids` (`learned_idsX`, etc), allowing to learn email in opposite direction as much as you want, which should not be the case.

### Steps to Reproduce

1. Learn eml as spam
2. Learn eml as spam second time - get `status code 208` already learned
3. Learn eml as ham - get `status code 200`
4. Learn eml as ham multiply time (f.e.: 3x) - get `status code 200`
5. Check redis `RS` key, where element `learns_spam` will be negative value which is wrong and bayes will not work at all from this point.
6. Check redis `learned_ids` key for this eml - flag will be still set as "spam" - `1`, when should be swapped to `-1` (as old, pre 3.8 behavior) or `0` (as 3.8 key for ham)

This works same in both directions, so if swap spam to ham in above steps - results will be same.

**Expected behavior**
Re-learning should change flag on relearning in opposite direction and **move** eml hash to "fresh" `learned_ids` slot (key). By *move* I mean:
1. check if we have key in any of `learned_ids`
4. if there "no learns" of previous learn has "opposite flag" from what we tries to learn proceed >>
5. if previous learn is in latest `learned_ids` - replace it with new "flag"
6. if previous learn not in latest `learned_ids` - add it to latest `learned_ids` with new flag + replace element key name in "older" `learned_ids` to value that will be never correct "hash"
3. learn bayes for needed flag.

[...]

Issue #3600, opened 2021-01-07
[Feature] unlearn ham/spam


jnorell:
## Summary

Please add the capability to unlearn bayes tokens.

## Motivation

Moving from spamassassin to rspamd, we have a training system in place that I'm converting, and rspamd seems to be missing a bayes "unlearn" capability.

We let users put mail into ham/spam training folders, and knowing users are not very adept at the task, the training messages are subsequently reviewed by a moderator - if there's a clear error, the moderator can simply correct the classification, but sometimes it is genuinely difficult to know (eg. user said "spam" but it looks legit, though not 100% positive) and in those cases the moderator puts the message in an "ignore" folder, so it won't influence the scanner.

With spamassassin you simply run "sa-learn --forget" for those messages to undo the influence the previous user classification had.  With rspamd it appears you can only learn as ham or spam, and to unlearn I presume you would need to wipe the entire bayes db and rebuild from stored message corpus.

## Describe alternatives you've considered

Alternatives considered:  wipe the bayes db and rebuild periodically (not yet tested for performance).  Rspamd's current functionality isn't suited for a moderated training folder setup (described above) because there is no way to undo the influence a message had in bayes from previous training.

## Additional context

n/a

jnorell:
silly stale bot....

ktetzlaff:
Here's another case which seems to ask for an unlearn feature. A SPAM message which is accidentally learned as HAM but also matches existing messages learned as SPAM cannot be removed from HAM:

1. message is classified as HAM:
```
# rspamc learn_ham '1626159428.M693251P477904.rakaposhi,S=54220,W=55107:2,S'
Results for file: 1626159428.M693251P477904.rakaposhi,S=54220,W=55107:2,S (0.003 seconds)
error = "<1626153864-2031512-1394241877-22ba2823aa1b9b4376448add3578dbcc-1e44cf@bounces.bettermail-news.de> has been already learned as ham, ignore it";
filename = "1626159428.M693251P477904.rakaposhi,S=54220,W=55107:2,S";
scan_time = 0.004000;
```

2. try to relearn as SPAM doesn't work:
```
[...]

PR #4930, opened 2024-04-18
[Fix] Relearn messages in bayes filter

aduernberger:
Relearning a message should have the same result as learning a message the first time. #4928, #3600
The following should have the same result:
- learn as spam
- learn as ham, then relearn as spam

The same should be true for the opposite direction:

- learn as ham
- learn as spam, then relearn as ham

With these changes, we were able to successfully relearn a message in both ways. The score always shows the same result, regardless of whether it is a relearn or first learn.

### DIFF
--- a/lualib/redis_scripts/bayes_cache_learn.lua
+++ b/lualib/redis_scripts/bayes_cache_learn.lua
@@ -15,7 +15,8 @@ for i = 0, conf.cache_max_keys do
   local have = redis.call('HGET', prefix, cache_id)
 
   if have then
-    -- Already in cache
+    -- Already in cache, but is_spam changes when relearning
+    redis.call('HSET', prefix, cache_id, is_spam)
     return false
   end
 end
--- a/lualib/redis_scripts/bayes_learn.lua
+++ b/lualib/redis_scripts/bayes_learn.lua
@@ -26,7 +26,7 @@ redis.call('HSET', prefix, 'version', '2') -- new schema
 redis.call('HINCRBY', prefix, learned_key, is_unlearn and -1 or 1) -- increase or decrease learned count
 
 for i, token in ipairs(input_tokens) do
-  redis.call('HINCRBY', token, hash_key, 1)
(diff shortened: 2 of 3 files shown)


lualib/redis_scripts/bayes_cache_learn.lua
```diff
@@ -15,7 +15,8 @@ for i = 0, conf.cache_max_keys do
   local have = redis.call('HGET', prefix, cache_id)
 
   if have then
-    -- Already in cache
+    -- Already in cache, but is_spam changes when relearning
+    redis.call('HSET', prefix, cache_id, is_spam)
```
vstakhov, line 19: 👍

lualib/redis_scripts/bayes_learn.lua
```diff
@@ -26,7 +26,7 @@ redis.call('HSET', prefix, 'version', '2') -- new schema
 redis.call('HINCRBY', prefix, learned_key, is_unlearn and -1 or 1) -- increase or decrease learned count
 
 for i, token in ipairs(input_tokens) do
-  redis.call('HINCRBY', token, hash_key, 1)
+  redis.call('HINCRBY', token, hash_key, is_unlearn and -1 or 1)
```
vstakhov, line 29: I should have used that in the first place, thanks!

src/libstat/learn_cache/redis_cache.cxx
```diff
@@ -170,7 +170,7 @@ rspamd_stat_cache_checked(lua_State *L)
 						  (task->flags & RSPAMD_TASK_FLAG_LEARN_SPAM) ? "spam" : "ham");
 			task->flags |= RSPAMD_TASK_FLAG_ALREADY_LEARNED;
 		}
-		else if (val != 0) {
+		else {
```
vstakhov, line 173: Agreed, we will have compatibility with the previous (broken) version then!

vstakhov (approved):
All very nice, thank you!

PR #4946, opened 2024-04-26
Fix neural check in lua_cfg_transform.lua

mykeenu:
check for neural.enabled to see if it's enabled. only checking for neural is always true unless someone messed with config files.


vstakhov:
We might also check if there are any rules defined there

vstakhov:
Tests are failing because `enabled` can be missing of course, so the check must be changed to something like `if cfg.neural and (type(cfg.neural.enabled) ~= 'boolean' or cfg.neural.enabled)`

PR #4948, opened 2024-04-27
Improve pyzor with min and halfscore threshold settings

mykeenu:
For some e-mails, pyzor reports a very low number of both blacklist and whitelist hits. This pull request adds a mechanism to only set the pyzor symbol if there is a minimum amount of total reports. Further, it adds another setting that allows you to only set half of the calculated pyzor score if the total report count is lower than this second threshold setting.

Theoretically this could be further improved by making a linear formula between the settings instead of the two steps here.

Both new variables are set to zero by default in order to not break existing deployments.

This can be used by adding something like this to `local.d/external_services.conf`:
```
pyzor {
  #servers = "127.0.0.1:5953";
  default_score = 2.5;
  min_threshold = 5;
  halfscore_threshold = 20;
}
```

### DIFF
--- a/lualib/lua_scanners/pyzor.lua
+++ b/lualib/lua_scanners/pyzor.lua
@@ -43,6 +43,8 @@ local function pyzor_config(opts)
     message = '${SCANNER}: Pyzor bulk message found: "${VIRUS}"',
     default_score = 1.5,
     action = false,
+    min_threshold = 0, -- zero score if whitelist+report count smaller than this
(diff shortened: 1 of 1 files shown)


vstakhov:
It looks good but I think we can improve it slightly by adding a table to match thresholds and multipliers. For example, something like this:

```
thresholds = {
   5 = 0.0,
   10 = 0.6,
   20 = 0.7,
   30 = 1.0,
}
```

That will make `min_threshold` and other things redundant.

PR #4950, opened 2024-04-28
Add Content-Description header rules

twesterhever:


twesterhever:
Implemented both change requests; thank you!

vstakhov:
Thank you!

PR #4953, opened 2024-05-02
[Fix] SUBJ_ALL_CAPS is overkill with some unicase scripts

ikedas:
For example, a subject header field with following content is determined to be all uppercase.
```
ツイスター、世界王者がAIロボットに完敗　得意技「関節外し」も及ばず
```
Ideographs and letters in unicase scripts which have Unicode general category `Lo` should be considered being not uppercase.

### DIFF
--- a/src/lua/lua_util.c
+++ b/src/lua/lua_util.c
@@ -1421,7 +1421,7 @@ lua_util_is_uppercase(lua_State *L)
 	LUA_TRACE_POINT;
 	int32_t i = 0;
 	UChar32 uc;
-	unsigned int nlc = 0, nuc = 0;
+	bool is_upper = false, is_lower = false, is_other = false;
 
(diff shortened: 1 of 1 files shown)


src/lua/lua_util.c
```diff
@@ -1421,7 +1421,7 @@ lua_util_is_uppercase(lua_State *L)
 	LUA_TRACE_POINT;
 	int32_t i = 0;
 	UChar32 uc;
-	unsigned int nlc = 0, nuc = 0;
+	unsigned int nuc = 0, nlc = 0, noc = 0;
```
vstakhov, line 1424: In fact, these should be booleans in this case.

vstakhov:
I think it looks very good, thank you!

src/lua/lua_util.c
ikedas: Make sense.  I re-pushed.
Thanks for suggestion!

PR #4960, opened 2024-05-13
[FIX] Arc regex to no longer match on dmarc=

### LINKED ISSUE #4959: [BUG] Incorrect ARC-Seal generated when reusing authentication results
**Describe the bug**
Incorrect  `Arc-Seal` header when configured with `reuse_auth_results = true;`

### Steps to Reproduce

1. Have an email with `Authentication-Results` header that contains `dmarc=` before `arc=`, or have one with no `arc=` line present
2. Ask rspamd to ARC sign the email while configured with `reuse_auth_results = true;`

**Expected behavior**
The `cv=` in the `ARC-Seal` header should not return the `dmarc=` results. It should only return `arc=` results

### Versions
All

JasonStephenson:
Fixes #4959
new_regex_correct_match
new_regex_correct_nomatch
old_regex_wrong_match

### DIFF
--- a/src/plugins/lua/arc.lua
+++ b/src/plugins/lua/arc.lua
@@ -642,7 +642,6 @@ local function prepare_arc_selector(task, sel)
          return v
         end
       end
-      
       return nil
     end
 
--- a/src/plugins/lua/arc.lua
+++ b/src/plugins/lua/arc.lua
@@ -635,11 +635,22 @@ local function prepare_arc_selector(task, sel)
       end
     end
 
+    local function arc_result_from_ar(ar_header)
+      ar_header = ar_header or ""
+      for k, v in string.gmatch(ar_header, "(%w+)=(%w+)") do
+        if k == 'arc' then
+         return v
+        end
+      end
+      
+      return nil
+    end
+
     if settings.reuse_auth_results then
       local ar_header = task:get_header('Authentication-Results')
 
       if ar_header then
-        local arc_match = string.match(ar_header, '[; ]arc=(%w+)')
+        local arc_match = arc_result_from_ar(ar_header)
 
         if arc_match then
           if arc_match == 'none' or arc_match == 'pass' then
--- a/src/plugins/lua/arc.lua
+++ b/src/plugins/lua/arc.lua
@@ -639,7 +639,7 @@ local function prepare_arc_selector(task, sel)
       local ar_header = task:get_header('Authentication-Results')
 
       if ar_header then
-        local arc_match = string.match(ar_header, 'arc=(%w+)')
+        local arc_match = string.match(ar_header, '[; ]arc=(%w+)')
 
         if arc_match then
           if arc_match == 'none' or arc_match == 'pass' then


src/plugins/lua/arc.lua
```diff
@@ -639,7 +639,7 @@ local function prepare_arc_selector(task, sel)
       local ar_header = task:get_header('Authentication-Results')
 
       if ar_header then
-        local arc_match = string.match(ar_header, 'arc=(%w+)')
+        local arc_match = string.match(ar_header, '[; ]arc=(%w+)')
```
vstakhov, line 642: It won't match a string where `arc=` is the first element...

vstakhov: I don't see any possible choices to match that thing with Lua `match` I'm afraid. With PCRE it's trivial.

JasonStephenson: aaah pants. I'll have another go at this soon

JasonStephenson: Currently toying with the idea of the pattern matching below, which allows `arc=` to be detected regardless of where in the header value it appears and regardless of the prescence of `dmarc=`.

```
 s = "spf=softfail (sender IP is 3.212.108.9) \r\nsmtp.mailfrom=gmail.com;  arc=TESTING; dkim=fail (body hash did not verify)\r\n header.d=gmail.com;dmarc=fail action=none header.from=gmail.com;"
 for k, v in string.gmatch(s, "(%w+)=(%w+)") do
   if k == 'arc' then
    print(v)
   end
 end
```

However https://datatracker.ietf.org/doc/html/rfc7601#section-2.2 states CFWS may be present. Would rspamd have unfolded the headers during its parsing? Or is this something I need to factor in?

vstakhov: Well, in fact that require an additional attention then. Rspamd provides headers unfolding for all headers but not SMTP comments removing if that means anything in this context.

JasonStephenson: I've changed my MR based on my previous comment and I believe it is sufficiently better to consider merging:

- This may not be a total final solution (accounting for CFWS comments) but is superior to the current code which incorrectly matches `dmarc=`
- I have a few email corpus, in which I can see no examples of CFWS in problematic areas (i.e between `arc=pass`)

For example:
I cannot see an examples of:
```
Authentication-Results: example.com;
  dkim ( DKIM check) = pass (signature was verified) header.d=example.com;
  spf ( SPF check) = pass (sender IP is authorized) smtp.mailfrom=example.com;
  dmarc ( DMARC check) = pass (p=reject, dis=none) header.from=example.com;
  arc ( ARC check) = pass (instance=1) example.com
```
Which would not be detected in my, nor the original code.

but i see examples of:
```
Authentication-Results: example.com;
  dkim=pass (signature was verified) header.d=example.com;
  spf=pass (sender IP is authorized) smtp.mailfrom=example.com;
  dmarc=pass (p=reject, dis=none) header.from=example.com;
  arc=pass (i=1) example.com
```
Which would have incorrect behavior in the original code (matching `dmarc=`) but correct behavior in mine

PR #4964, opened 2024-05-13
[Minor] Decrease XM_UA_NO_VERSION score to 0.0

ishisora:
My rspamd is always getting a score due to XM_UA_NO_VERSION. That's because Thunderbird's default settings send a minimal user-agent.

* [Thunderbird Release Notes](https://www.thunderbird.net/en-US/thunderbird/115.0/releasenotes/)
> Thunderbird will now send a minimal user-agent header by default

The disclosure of the sender's OS and Thunderbird version to the recipient and any intermediaries raises security and privacy concerns.

Based on the above, the score should be reduced from 0.01 to 0.0. Setting it to 0.0 will allow detection of XM_UA_NO_VERSION without affecting actions.

### DIFF
--- a/rules/regexp/headers.lua
+++ b/rules/regexp/headers.lua
@@ -973,7 +973,7 @@ reconf['XM_UA_NO_VERSION'] = {
       'X-Mailer=/^[^0-9]+$/H',
       'User-Agent=/^[^0-9]+$/H'),
   description = 'X-Mailer/User-Agent header has no version number',
-  score = 0.01,
+  score = 0.0,
   group = 'experimental'
 }
 


moisseev:
1. Can a score of 0.01 have a real impact on actions?
2. If so, the correct approach would be making an exclusion for `User-Agent: Mozilla Thunderbird` instead of setting neutral score for all MUAs.

ishisora:
> 1. Can a score of 0.01 have a real impact on actions?
> 2. If so, the correct approach would be making an exclusion for `User-Agent: Mozilla Thunderbird` instead of setting neutral score for all MUAs.

Thank you for your response. Because 0.01 is a very small score, I don't think it directly affects actions. However, since positive scores are displayed in red when checking scores on a web page, it catches my attention every time.

PR #4972, opened 2024-05-15
Add bayes_expiry.conf to provide ability to set custom settings

dragoangel:


moisseev:
In my opinion, there's no necessity to redefine module defaults. Instead, I suggest commenting parameters out and leaving them as examples.

Short comments on their meanings would be useful as well. ;)

dragoangel:
I just do it same way I see it in other modules, it has a place in other conf. Also maybe instead of adding comments I better expand rspamd docs?

dragoangel:
Removed settings duplication. Will create PR to docs separately

dragoangel:
@moisseev https://github.com/rspamd/rspamd.com/pull/748 please check, if this is okay, can you approve both PRs please?

PR #4980, opened 2024-05-23
[Rules] Added rules for detecting likely malware

fatalbanana:
- Punish exe-laden archives in general
- Harsh punishment for exe-laden archives with suspicious filenames or subjects
- Harsh punishment for fantastically well compressed UDF images
- Punish misidentified RARs & misidentified RARs bearing executables


vstakhov:
@moisseev @dragoangel I would appreciate if you take a quick look on this.

rules/archives.lua
```diff
@@ -0,0 +1,152 @@
...
+          if fn then
+            if not rar_re:match(fn) then
+              task:insert_result('MISIDENTIFIED_RAR', 1.0)
+              misidentified_rar = true
+            end
+          end
+        end
+
+        local files = arc:get_files_full()
+	local max_check = #files
```
vstakhov, line 48: nit: formatting

rules/archives.lua
```diff
@@ -0,0 +1,152 @@
...
+        if max_check > 10 then
+          max_check = 10
+        end
```
vstakhov, line 51:
```suggestion
max_check = math.min(max_check, 10)
```

rules/archives.lua
```diff
@@ -0,0 +1,152 @@
...
+          return
+        end
+      end
+    end
+  end,
+  name = 'CHECK_ARCHIVES',
+  type = 'callback',
+}
+
+rspamd_config:register_symbol{
```
vstakhov, line 92: Please add descriptions of possible.

dragoangel:
Hi @vstakhov @fatalbanana , my view on this:

1. Join @moisseev that we should not punish `.exe` and other thing, as of:
I assume such files most administrators already ban via `mime_types.lua` or multimaps with `prefilter`. Maybe this module will be okay for people who want accepting `exe`s or `img`, but I not sure if naming is enough to ban `exe`, this too much dynamic thing. Compressed `img` - same, not better just block it and that's all? Usually "valid" img files takes more then 100mb, correct me if I'm wrong? Malware can skip compressing img at all, but greatly obfuscate and compress malware itself. Plus there more extensions that used for same thing: `dmg` & `iso` at minimum. Saw them in spam campaigns and blocked at `mime_types.lua`.
2. About `misidentified RARs` - nice thing, but if we can detect in Rspamd content extensions inside of such archives - I think better to still punish them by `mime_types.lua`+this module via composite. Maybe I missed something and this not possible, but it looks more optimal.

Issue #5010, opened 2024-06-10
[BUG] Lua API issue: `task:inject_url` could also perform URL lookup inside the injected URL query


vstakhov:
Some URLs extracted can have another URLs encoded in their queries, e.g. `http://example.com?redir=http://another.com` (probably url encoded). Rspamd already has logic to find urls in queries (see `rspamd_url_query_callback`) but it is not implemented for injected urls, for example those that are extracted from PDF documents.

FYI: @citrin

PR #5021, opened 2024-06-18
Fixed bug with task:inject() function according to the issue #5010

LeftTry:


src/lua/lua_task.c
```diff
@@ -2644,22 +2644,48 @@ lua_task_inject_url(lua_State *L)
...
-		mpart = *((struct rspamd_mime_part **) rspamd_lua_check_udata_maybe(L,
-																			3, rspamd_mimepart_classname));
+		mpart = *((struct rspamd_mime_part **)
+							   rspamd_lua_check_udata_maybe(L,3,rspamd_mimepart_classname));
 	}
-
 	if (task && task->message && url && url->url) {
 		if (rspamd_url_set_add_or_increase(MESSAGE_FIELD(task, urls), url->url, false)) {
+			struct rspamd_mime_text_part *text_part;
+			text_part = rspamd_mempool_alloc0(task->task_pool,
```
vstakhov, line 2653: Why do we need a fake text part here? Please check how it's done in HTML urls extraction: https://github.com/rspamd/rspamd/blob/cd92d8bb1d412abc4aaaea57efb55145f9f49b8b/src/libserver/html/html.cxx#L749

src/lua/lua_task.c
```diff
@@ -2644,22 +2644,48 @@ lua_task_inject_url(lua_State *L)
...
+				text_part->parsed.begin = mpart->parsed_data.begin;
+				text_part->parsed.len = mpart->parsed_data.len;
+				text_part->flags |= flags;
+
+				/* Adding mime_text_part to mime_part */
+				g_ptr_array_add(MESSAGE_FIELD(task, text_parts), text_part);
+				mpart->part_type = RSPAMD_MIME_PART_TEXT;
+				mpart->specific.txt = text_part;
+			} else {
+				/* Creating empty mime_part if no mime_part was got from lua_state */
```
vstakhov, line 2671: No, please don't do it!

src/lua/lua_url.c
```diff
@@ -855,18 +855,20 @@ lua_url_create(lua_State *L)
...
+		}
 	}
 
 	if (pool == NULL) {
-		return luaL_error(L, "invalid arguments: mempool is expected as the second argument");
-	}
-
-	if (t == NULL) {
-		return luaL_error(L, "invalid arguments: string/text is expected as the first argument");
+		return luaL_error(L, "invalid arguments: mempool is expected as the first argument");
```
vstakhov, line 871: 👍

vstakhov (changes requested):

LeftTry:
Updated the code in accordance with the requested changes.

src/lua/lua_task.c
```diff
@@ -2644,22 +2691,22 @@ lua_task_inject_url(lua_State *L)
...
+		mpart = *((struct rspamd_mime_part **)
+							   rspamd_lua_check_udata_maybe(L,3,rspamd_mimepart_classname));
 	}
-
 	if (task && task->message && url && url->url) {
 		if (rspamd_url_set_add_or_increase(MESSAGE_FIELD(task, urls), url->url, false)) {
-			if (mpart && mpart->urls) {
-				/* Also add url to the mime part */
-				g_ptr_array_add(mpart->urls, url->url);
+			if(mpart == NULL) {
```
vstakhov, line 2699: We don't need a fake mpart, why do we?

LeftTry: So, we also do not need to extract mime part form Lua State?

src/lua/lua_task.c
```diff
@@ -2644,22 +2691,18 @@ lua_task_inject_url(lua_State *L)
...
 	}
-
 	if (task && task->message && url && url->url) {
 		if (rspamd_url_set_add_or_increase(MESSAGE_FIELD(task, urls), url->url, false)) {
-			if (mpart && mpart->urls) {
-				/* Also add url to the mime part */
-				g_ptr_array_add(mpart->urls, url->url);
-			}
+			GPtrArray* part_urls = g_ptr_array_new();
+			inject_url_query(task, url->url, part_urls);
```
vstakhov, line 2700: I'm sorry but I don't understand that. Why don't we use `mpart->urls` here? We will clearly 1) miss urls from part; 2) leak memory.

src/lua/lua_task.c
```diff
@@ -2644,22 +2691,18 @@ lua_task_inject_url(lua_State *L)
...
 		}
 	}
 	else {
 		return luaL_error(L, "invalid arguments");
 	}
-
```
vstakhov, line 2662: Please leave this empty line.

src/lua/lua_task.c
```diff
@@ -2644,22 +2691,18 @@ lua_task_inject_url(lua_State *L)
 
 	if (lua_isuserdata(L, 3)) {
 		/* We also have a mime part there */
-		mpart = *((struct rspamd_mime_part **) rspamd_lua_check_udata_maybe(L,
-																			3, rspamd_mimepart_classname));
+		mpart = *((struct rspamd_mime_part **)
+							   rspamd_lua_check_udata_maybe(L,3,rspamd_mimepart_classname));
```
vstakhov, line 2695: Please add spaces after commas.

src/lua/lua_task.c
LeftTry: Because if maprt is NULL, it will cause a segmentation fault, so in order not to catch this error, we will need to create a fake mpart, but as mentioned earlier, we should not fake the mpart. Also, if I understand correctly, the mime part has already parsed all its URLs to `taks->urls`.

vstakhov: No, if mpart is null, then you don't need to insert a new url there, that's it.

vstakhov (approved):

vstakhov:
Could you please revert the last commit [8c4e951](https://github.com/rspamd/rspamd/pull/5021/commits/8c4e951b81a60b86fc9fc32674ff4d15badc8222) from this branch and use a separate branch and PR for this work?

LeftTry:
Yes. Sorry for this. I forgot that this branch was not merged. Wait a few minutes.

PR #5030, opened 2024-06-25
Exclude MIME_BAD_UNICODE false positive

dragoangel:
Resolve cases from https://github.com/rspamd/rspamd/discussions/4893

### DIFF
--- a/conf/composites.conf
+++ b/conf/composites.conf
@@ -204,7 +204,7 @@ composites {
     description = "Message contains redirector, anonymous or IPFS gateway URL and is marked by fuzzy/bayes/SURBL/RBL";
   }
   MIME_BAD_UNICODE_GENERAL_PUNCTUATION {
-    expression = "^MIME_BAD_UNICODE[/^0x(20(0|2|6)[8-9a-f]|feff)d after /i]";
+    expression = "^MIME_BAD_UNICODE[/^0x(20[026][8-9a-f]|feff)d after /i]";
     description = "Exclude legit general punctuation unicode characters from MIME_BAD_UNICODE";
     score = 0.0;
   }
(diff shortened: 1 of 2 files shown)


conf/composites.conf
```diff
@@ -203,6 +203,16 @@ composites {
     policy = "leave";
     description = "Message contains redirector, anonymous or IPFS gateway URL and is marked by fuzzy/bayes/SURBL/RBL";
   }
+  MIME_BAD_UNICODE_NNBSP {
+    expression = "^MIME_BAD_UNICODE[/^0x202fd after /i]";
+    description = "Exclude NNBSP (narrow no-break space) from MIME_BAD_UNICODE which used by MacOS Mail.app";
```
moisseev, line 208:
```diff
-description = "Exclude NNBSP (narrow no-break space) from MIME_BAD_UNICODE which used by MacOS Mail.app";
+description = "Exclude NNBSP (narrow no-break space) from MIME_BAD_UNICODE which is used by MacOS Mail.app";
```

dragoangel: @moisseev I also found that there other cases, in total it's quite big list:
```
200bd
200cd
200dd
200ed
200fd
2028d
2029d
202ad
202bd
202cd
202ed
202fd
2068d
2069d
206ad
206cd
206ed
206fd
feffd
```

All them are special chars if strip last "d", reference: https://www.compart.com/en/unicode/U+202e

dragoangel:
`MIME_BAD_EXTENSION(9.10){exe;},MIME_BAD_UNICODE(8.00){0x202ed after scan;},MIME_OBFUSCATED_ARCHIVE(8.00){obfuscated archive;scan�s�x�l�x�.rar;` - case of evil usage of 202e -LRM, I think the real proper fix would be to set `MIME_BAD_UNICODE` and `MIME_OBFUSCATED_ARCHIVE` symbol to `1.0` if not `0.1`, and give them score via composite over `MIME_BAD_EXTENSION`, because there is too much cases when people can name files like `Müller.rar`, `החתלתול האהוב עליי.jpg` or `قطتي المفضلة.jpg` and so on...

PR #5032, opened 2024-06-27
Add GPT plugin

vstakhov:


vstakhov:
@fatalbanana could you please take a look at this? :)

src/plugins/lua/gpt.lua
```diff
@@ -0,0 +1,333 @@
...
+  end
+
+  return spam_score
+end
+
+local function openai_gpt_check(task)
+  local ret, content = settings.condition(task)
+
+  if not ret then
+    lua_util.info(N, task, "skip checking gpt as the condition is not met: %s", content)
```
fatalbanana, line 190:
```suggestion
    rspamd_logger.info(task, "skip checking gpt as the condition is not met: %s", content)
```

src/plugins/lua/gpt.lua
```diff
@@ -0,0 +1,333 @@
...
+  top_p = 0.9,
+  timeout = 10,
+  prompt = nil,
+  condition = nil,
+  autolearn = false,
+  url = 'https://api.openai.com/v1/chat/completions',
+}
+
+-- Exclude checks if one of those is found
+local symbols_to_except = {
```
fatalbanana, line 73: Might we put it in configuration?

conf/modules.d/gpt.conf
```diff
@@ -0,0 +1,46 @@
...
+
+gpt {
+  # Supported types: openai
+  type = "openai";
+  # Your key to access the API (add this to enable this plugin)
+  #api_key = "xxx";
+  # Model name
+  model = "gpt-3.5-turbo";
+  # Maximum tokens to generate
+  max_tokens = 100;
```
fatalbanana, line 23: It doesn't match module-default & config example

fatalbanana: Also, this is punny because it is both response tokens & number of words sent in request

src/plugins/lua/gpt.lua
```diff
@@ -0,0 +1,336 @@
...
+  if nwords > settings.max_tokens then
+    -- We need to truncate words (sometimes get_words_count returns a different number comparing to `get_words`)
+    local words = sel_part:get_words('norm')
+    nwords = #words
+    if nwords > settings.max_tokens then
+      -- Trim something that does not fit
+      for i = nwords, settings.max_tokens, -1 do
+        rawset(words, i, nil)
+      end
+      return true, table.concat(words, ' ')
```
fatalbanana, line 145: How about `table.concat(words, ' ', 1, math.min(settings.max_tokens, nwords))`

vstakhov: That looks better, thanks.

dragoangel:
I think adding headers like subject and from in addition to email would help better classify it.

Also I think we need exclude by default `REPLIES` and `BOUNCE`.

PR #5036, opened 2024-07-02
Draft PR for OpenSSL 3.0 support issue

LeftTry:
This PR already has edits for this issue, but they need to be tested.


vstakhov:
Please check output of https://github.com/rspamd/rspamd/actions/runs/9755846790/job/26933504019?pr=5036

There are some issues introduced:

* Incompatible pointers passed
* Warnings in dkim.c (maybe somewhere else as well)

Could you please address those and make build warnings free where possible?

src/libcryptobox/cryptobox.c
```diff
@@ -630,7 +645,7 @@ void rspamd_cryptobox_sign(unsigned char *sig, unsigned long long *siglen_p,
 		g_assert(EVP_DigestSignInit(sha_ctx, &pctx, EVP_sha512(),
 									ENGINE_F_ENGINE_NEW, pkey) == 1);
 
-		g_assert(EVP_DigestSignFinal(sha_ctx, sig, &diglen) == 1);
+		g_assert(EVP_DigestSignFinal(sha_ctx, sig, (size_t *) &diglen) == 1);
```
vstakhov, line 648: It's not safe to cast pointers to integers of different size. I would say, that either `diglen` should be changed to `size_t` or there should be a dedicated local variable of `size_t` type for that purposes (if the first option is not possible only).

src/libcryptobox/cryptobox.c
```diff
@@ -697,7 +712,8 @@ bool rspamd_cryptobox_verify(const unsigned char *sig,
 		g_assert(EVP_DigestSignInit(sha_ctx, &pctx, EVP_sha512(),
 									ENGINE_F_ENGINE_NEW, pkey) == 1);
 
-		g_assert(EVP_DigestSignFinal(sha_ctx, sig, &siglen) == 1);
+		unsigned char non_const_sig = *sig;
+		g_assert(EVP_DigestSignFinal(sha_ctx, &non_const_sig, &siglen) == 1);
```
vstakhov, line 716: On the contrary, it is **possible** to convert `const char *` to `unsigned char *`, and with your change the whole signatures checking will be broken, as apparently signature is not a 1 character string (it also will cause OOB reading which is a potential vulnerability).

src/libcryptobox/cryptobox.c
```diff
@@ -360,9 +357,19 @@ void rspamd_cryptobox_keypair(rspamd_pk_t pk, rspamd_sk_t sk,
 		ec_pub = EC_POINT_new(group);
 		g_assert(ec_pub != NULL);
 
+		unsigned char *buf = NULL;
+		EVP_PKEY_get_raw_private_key(pkey, buf, &len);
```
vstakhov, line 361: EVP_PKEY_get_raw_private_key() fills the buffer provided by priv with raw private key data. The size of the priv buffer should be in *len on entry to the function, and on exit *len is updated with the number of bytes actually written. If the buffer priv is NULL then *len is populated with the number of bytes required to hold the key. The calling application is responsible for ensuring that the buffer is large enough to receive the private key data. This function only works for algorithms that support raw private keys. Currently this is: EVP_PKEY_HMAC, EVP_PKEY_POLY1305, EVP_PKEY_SIPHASH, EVP_PKEY_X25519, EVP_PKEY_ED25519, EVP_PKEY_X448 or EVP_PKEY_ED448.

Hence, this will not allocate `buf` at all, it will only populate `len` required for it. We definitely need tests for these changes, as OpenSSL API is way too confusing and dangerous.

LeftTry:
There are some places in code such as` lua_rsa.c` and parts of `dkim.c` and `lua_cryptobox.c` that operate with completely deprecated structures such as `RSA`, `EC_KEY` and `HMAC`, all functions of which were deprecated since OpenSSL 3.0. So in this case what should we do with this structures? Just make them operate with older versions of OpenSSL?

test/rspamd_cxx_unit_cryptobox.hxx
```diff
@@ -0,0 +1,182 @@
...
+		rspamd_cryptobox_keypair_sig(pk, sk, mode);
+	}
+*/
+	TEST_CASE("rspamd_cryptobox_hash")
+	{
+		rspamd_cryptobox_hash_state_t p;
+		const unsigned char *key = reinterpret_cast<const unsigned char *>("keys");
+		gsize keylen = sizeof(key);
+
+		ottery_rand_bytes(&p, rspamd_cryptobox_HASHSTATEBYTES);
```
vstakhov, line 58: Why do we do it? Hash init will reset state anyway.

test/rspamd_cxx_unit_cryptobox.hxx
```diff
@@ -0,0 +1,182 @@
...
+		const unsigned char* data = reinterpret_cast<const unsigned char *>("data");
+		gsize len = sizeof(data);
+
+		rspamd_cryptobox_hash_update(&p, data, len);
+
+		unsigned char out1[rspamd_cryptobox_HASHSTATEBYTES];
+		memset(out1, 0, rspamd_cryptobox_HASHSTATEBYTES);
+
+		rspamd_cryptobox_hash_final(&p, out1);
+		CHECK(*out1 == 'd');
```
vstakhov, line 71: Why do we check one byte and not the whole hash?

test/rspamd_cxx_unit_cryptobox.hxx
```diff
@@ -0,0 +1,182 @@
...
+		rspamd_cryptobox_fast_hash_state_s *st = rspamd_cryptobox_fast_hash_new();
+
+		uint64_t seed = 10;
+
+		rspamd_cryptobox_fast_hash_init(st, seed);
+
+		const unsigned char* data = reinterpret_cast<const unsigned char *>("data");
+		gsize len = sizeof(data);
+
+		rspamd_cryptobox_fast_hash_update(st, data, len);
```
vstakhov, line 91: We should do IUF in pieces to test something meaningful. We also need to check that our hash is matching our expectations.

test/rspamd_cxx_unit_cryptobox.hxx
```diff
@@ -0,0 +1,182 @@
...
+		unsigned int complexity = 10;
+		enum rspamd_cryptobox_pbkdf_type type = RSPAMD_CRYPTOBOX_PBKDF2;
+
+
+		CHECK(rspamd_cryptobox_pbkdf(pass, pass_len, &salt, salt_len, &key, key_len, complexity, type) == true);
+		CHECK(key == 60);
+
+		type = RSPAMD_CRYPTOBOX_CATENA;
+		CHECK(rspamd_cryptobox_pbkdf(pass, pass_len, &salt, salt_len, &key, key_len, complexity, type) == true);
+		CHECK(key == 204);
```
vstakhov, line 121: Again, this is not enough: we need to ensure that pbkdf works correctly for a given password/salt pair.

test/rspamd_cxx_unit_cryptobox.hxx
```diff
@@ -0,0 +1,182 @@
...
+		unsigned char * sig = (unsigned char *) "sig";
+		unsigned long long siglen = sizeof(sig);
+		const unsigned char m[] = "data to be signed";
+		size_t mlen = strlen((const char*)m);
+
+		ottery_rand_bytes(sk, sizeof(sk));
+		ottery_rand_bytes(pk, sizeof(pk));
+
+		rspamd_cryptobox_sign(sig, &siglen, m, mlen, sk, mode);
+		bool check_result = rspamd_cryptobox_verify_compat(sig, reinterpret_cast<gsize>(&siglen), m, mlen, pk, sk, mode);
```
vstakhov, line 179: I'm sorry but I don't understand what you are doing here. Why do we need a **secret** key for verification??

test/rspamd_cxx_unit_cryptobox.hxx
```diff
@@ -0,0 +1,182 @@
...
+		enum rspamd_cryptobox_mode mode = RSPAMD_CRYPTOBOX_MODE_NIST;
+		rspamd_sk_t sk;
+		rspamd_pk_t pk;
+		unsigned char * sig = (unsigned char *) "sig";
+		unsigned long long siglen = sizeof(sig);
+		const unsigned char m[] = "data to be signed";
+		size_t mlen = strlen((const char*)m);
+
+		ottery_rand_bytes(sk, sizeof(sk));
+		ottery_rand_bytes(pk, sizeof(pk));
```
vstakhov, line 174: No, it won't work in this way. You **must** create a pair of secret/public key to do anything sane. Also, IIRC, for p256 you simply can't do a random secret key - you need to verify that this random sequence belongs to the curve. For ed25519 it is less concern but still there are some bit that needs to be set properly.

Public key generation is simply wrong.

src/libcryptobox/cryptobox.c
```diff
@@ -677,6 +673,59 @@ void rspamd_cryptobox_sign(unsigned char *sig, unsigned long long *siglen_p,
 	}
 }
 
+#if OPENSSL_VERSION_MAJOR >= 3
+bool rspamd_cryptobox_verify_compat(const unsigned char *sig,
+									gsize siglen,
+									const unsigned char *m,
+									gsize mlen,
+									const rspamd_pk_t pk,
+									const rspamd_sk_t sk,
```
vstakhov, line 684: No, verify should not use secret key - it is wrong.

src/libcryptobox/cryptobox.c
```diff
@@ -677,6 +673,59 @@ void rspamd_cryptobox_sign(unsigned char *sig, unsigned long long *siglen_p,
...
+
+		/* Prehash */
+		sha_ctx = EVP_MD_CTX_create();
+		g_assert(EVP_DigestInit(sha_ctx, EVP_sha512()) == 1);
+		EVP_DigestUpdate(sha_ctx, m, mlen);
+		EVP_DigestFinal(sha_ctx, h, NULL);
+
+
+		OSSL_LIB_CTX *libctx = OSSL_LIB_CTX_new();
+		EVP_PKEY_CTX *pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL);
```
vstakhov, line 708: You should not produce public key from private, you need to use public key for verifications - that's how digital signatures ever works.

src/libcryptobox/cryptobox.c
```diff
@@ -706,18 +755,18 @@ bool rspamd_cryptobox_verify(const unsigned char *sig,
 
 
 #if OPENSSL_VERSION_MAJOR >= 3
-		EVP_PKEY_CTX *pctx = EVP_PKEY_CTX_new_id(CRYPTOBOX_CURVE_NID, ENGINE_F_ENGINE_NEW);
-		g_assert(pctx != NULL);
+		OSSL_LIB_CTX *libctx = OSSL_LIB_CTX_new();
+		EVP_PKEY *pkey = EVP_PKEY_new_raw_public_key_ex(libctx, "ED25519", NULL, pk, sizeof(rspamd_sk_t));
```
vstakhov, line 761: Why `ed25519`? For this curve we use libsodium functions. Openssl is responsible for p256 operations only.

src/libcryptobox/cryptobox.c
```diff
@@ -1586,16 +1635,23 @@ unsigned int rspamd_cryptobox_signature_bytes(enum rspamd_cryptobox_mode mode)
 #else
 #if OPENSSL_VERSION_MAJOR >= 3
 		if (ssl_keylen == 0) {
-			EVP_PKEY *pkey = EVP_PKEY_new();
-			EVP_PKEY_CTX *pctx = EVP_PKEY_CTX_new(pkey, ENGINE_F_ENGINE_NEW);
-			EVP_PKEY_CTX_set_ec_paramgen_curve_nid(pctx, CRYPTOBOX_CURVE_NID);
-			g_assert(pctx != NULL);
+			OSSL_LIB_CTX *libctx = OSSL_LIB_CTX_new();
```
vstakhov, line 1640: `#define CRYPTOBOX_CURVE_NID NID_X9_62_prime256v1` - and this curve should remain the same!

vstakhov (changes requested):

src/libcryptobox/cryptobox.c
LeftTry: The `new_raw_private_key` function works only with `Ed25519`, `Ed448`, `X25519` and `X448`. Maybe I don't understand something vital about PKEY, but there is no way to create PKEY out of buffer `pk` or `sk` of type EC.

test/rspamd_cxx_unit_cryptobox.hxx
vstakhov: pbkdf returns a string that is a safe to store representation of the password. So we need to check if we can derive the same encrypted string from the same password given that we use the same salt. Also, why salt is only 1 byte?

src/libcryptobox/cryptobox.c
vstakhov: I don't get it: so if this function can't create p256 keys we need to change p256 to some other curve? No it's not acceptable, we need to create elliptic keys exactly as we did with older OpenSSL, we cannot simply change the curve. Also, I don't think that `EVP_PKEY_CTX_new` is now deprecated - why can't we leave it as is?

vstakhov: In fact, we just need to replace `EC_KEY_new_by_curve_name` with `EVP_EC_gen`

vstakhov: https://www.openssl.org/docs/manmaster/man7/EVP_PKEY-EC.html

Issue #5027, opened 2024-06-23
[BUG] `metric_exporter` sends `nil` instead of `0` for action metrics


ivanov17:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

When I try to use the `metric_exporter` module to get graphite metrics on Prometheus [graphite_exporter](https://github.com/prometheus/graphite_exporter/), the last one [returns the following errors](https://github.com/prometheus/graphite_exporter/issues/264):

```
ts=2024-06-21T16:18:44.573Z caller=main.go:85 level=info msg="Starting graphite_exporter" version_info="(version=0.15.1, branch=HEAD, revision=cf47e460f839e7e42f1d6016b1300df2e3d15ad9)"
ts=2024-06-21T16:18:44.574Z caller=main.go:86 level=info build_context="(go=go1.22.0, platform=linux/amd64, user=root@e9d9aba80dbb, date=20240322-16:31:01, tags=netgo)"
ts=2024-06-21T16:18:44.578Z caller=tls_config.go:313 level=info msg="Listening on" address=[::]:9108
ts=2024-06-21T16:18:44.578Z caller=tls_config.go:316 level=info msg="TLS is disabled." http2=false address=[::]:9108
ts=2024-06-21T16:18:45.901Z caller=collector.go:131 level=debug msg="Incoming line" line="rspamd.actions.add_header nil 1718986725"
ts=2024-06-21T16:18:45.901Z caller=collector.go:166 level=info msg="Invalid value" line="rspamd.actions.add_header nil 1718986725"
ts=2024-06-21T16:18:45.901Z caller=collector.go:131 level=debug msg="Incoming line" line="rspamd.actions.greylist nil 1718986725"
[...]

PR #5041, opened 2024-07-08
[Fix] metric_exporter: avoid sending `nil` in metric values (#5027)

fatalbanana:
https://github.com/rspamd/rspamd/issues/5027


vstakhov:
Maybe we still need to insert those values? I'm not quite sure how is it intended to work.

vstakhov:
> Maybe we still need to insert those values? I'm not quite sure how is it intended to work.

I mean why not just change `nil` to `null` as Prometheus expects?

Issue #5028, opened 2024-06-24
[BUG] Bayes autolearn relearns (reclassifies) mail


aduernberger:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

Bayes autolearn enabled.
Autolearn will always try to relearn the mail, even if it has already been learned as spam or ham.

### Steps to Reproduce

1. Mail is detected as spam and then autolearned as spam.
2. The mail is then relearned as ham (rspamc)
3. Same Mail is recieved again. But still detected as SPAM (eg. SPAM Score -> 7.0, Mail Score -> 11, Bayes Ham -> -3.0, final Score: 8.0 == SPAM)
4. Autolearn will relearn the Mail from HAM to SPAM

**Expected behavior**

A Mail that was already learned should not be relearned automatically.
rspamc should still be able to be used to relearn a mail.

### Versions

current master

### Additional Information

I am not sure if this is part of the same problem or a completely separate issue.

While debugging, I noticed that `RSPAMD_TASK_FLAG_LEARN_SPAM` or `RSPAMD_TASK_FLAG_LEARN_HAM` is always set when the Mail (detected as spam) is already learned as `SPAM`.

So the `if (task->flags &...` in `task.c` is always `true` even if the mail is already learned.
```
	case RSPAMD_TASK_STAGE_LEARN_POST:
		if (task->flags & (RSPAMD_TASK_FLAG_LEARN_SPAM | RSPAMD_TASK_FLAG_LEARN_HAM)) {
			if (task->err == NULL) {
				if (!rspamd_stat_learn(task,
```

I think that stems from `stat_process.c`
```
			if (ret) {
				/* Do not autolearn if we have this symbol already */
[...]

PR #5047, opened 2024-07-09
[Fix] Prevent reclassify on autolearn

aduernberger:
Bayes autolearn is able to reclassify a Mail from Spam to Ham and Ham to Spam. #5028

The idea behind this is to call `rspamd_stat_cache_check` before checking if autolearn is possible and after all stages that modify the score are finished.
This sets the `RSPAMD_TASK_FLAG_ALREADY_LEARNED` and `RSPAMD_TASK_FLAG_UNLEARN` flags.

`rspamd_stat_autolearn` is basically a copy of `rspamd_stat_classify`.
I tried calling `rspamd_stat_cache_check` directly the original `rspamd_stat_check_autolearn` but the flags are set in a Lua callback, so not there yet when I try to check for them. 

Essentially it adds three new stages `RSPAMD_TASK_STAGE_AUTOLEARN_PRE`, `RSPAMD_TASK_STAGE_AUTOLEARN` and `RSPAMD_TASK_STAGE_AUTOLEARN_POST` (not used, just there to be equal to the other stages). First stage calls `rspamd_stat_cache_check` to get the necessary symbols. Second stage calls the original  `rspamd_stat_check_autolearn` and does not learn if `RSPAMD_TASK_FLAG_ALREADY_LEARNED` or `RSPAMD_TASK_FLAG_UNLEARN` flag are set.


src/libstat/stat_process.c
```diff
@@ -1262,3 +1263,33 @@ rspamd_stat_statistics(struct rspamd_task *task,
...
+	st_ctx = rspamd_stat_get_ctx();
+	g_assert(st_ctx != NULL);
+
+	if (st_ctx->classifiers->len == 0) {
+		task->processed_stages |= stage;
+		return ret;
+	}
+
+	if (stage == RSPAMD_TASK_STAGE_AUTOLEARN_PRE) {
+		rspamd_stat_cache_check(st_ctx, task, task->classifier, task->flags & RSPAMD_TASK_FLAG_LEARN_SPAM, &err);
```
vstakhov, line 1283: In this case, we will check stat cache twice: once on autolearn and one on the actual learning.

src/libserver/task.h
```diff
@@ -51,13 +51,16 @@ enum rspamd_task_stage {
...
 	RSPAMD_TASK_STAGE_COMPOSITES = (1u << 9u),
 	RSPAMD_TASK_STAGE_POST_FILTERS = (1u << 10u),
-	RSPAMD_TASK_STAGE_LEARN_PRE = (1u << 11u),
-	RSPAMD_TASK_STAGE_LEARN = (1u << 12u),
-	RSPAMD_TASK_STAGE_LEARN_POST = (1u << 13u),
-	RSPAMD_TASK_STAGE_COMPOSITES_POST = (1u << 14u),
-	RSPAMD_TASK_STAGE_IDEMPOTENT = (1u << 15u),
-	RSPAMD_TASK_STAGE_DONE = (1u << 16u),
-	RSPAMD_TASK_STAGE_REPLIED = (1u << 17u)
+	RSPAMD_TASK_STAGE_AUTOLEARN_PRE = (1u << 11u),
```
vstakhov, line 54: I'm not convinced that we need 3 extra stages for that. We can probably move the whole logic to `STAGE_LEARN_PRE` and execute it unconditionally.

aduernberger: Correct me if I am wrong, but as I see it, we need 2 stages to decide if we can autolearn.

As we only get the flags from `cache_check` asynchronously, since it is a Lua callback that sets them.
There is code in `cache_check` that sets `RSPAMD_TASK_FLAG_UNLEARN` but as far as I see it can only be reached with the sqlite3 backend which has code to return  `RSPAMD_LEARN_UNLEARN`.

So if `STAGE_LEARN_PRE` does the cache check. We only get the symbols in the next stage (unless there is a way to get the Flags of the LUA callback earlier).
So the decision on whether we do autolearn or not would have to move into `STAGE_LEARN` which would also mean it has to be called unconditionally (at least for the first part of the stage).

aduernberger:
So I had a go at a solution without creating new stages. I have two different attempts that both work in my testing.

https://github.com/aduernberger/rspamd/commit/99949a0181aa8f2c1d68b42b930ac56e2d73d8f0
 - `rspamd_stat_learn`  does the autolearn check. Which requires `rspamd_stat_learn` to be called when `LEARN_AUTO` is set.
 - `rspamd_stat_learn` checks if the LEARN flags are set wherever required. (Diff looks bigger than it is because of that)
 - This means that compared to master, `rspamd_stat_learn` is always called three times (since the conditions are now inside `rspamd_stat_learn`)

https://github.com/aduernberger/rspamd/commit/b46412eea9e9d5d3f6d17fb1682fadf121dca585
 - Moves the error handling for `rspamd_stat_learn` from `task.c` to `rspamd_stat_learn` in `stat_process.c` (to make it possible to call `rspamd_stat_learn` from `RSPAMD_TASK_STAGE_LEARN_PRE` without having to copy the complete error handling)
 - Calls `rspamd_stat_learn` in `RSPAMD_TASK_STAGE_LEARN_PRE` directly in `task.c` when any of the `LEARN` flags are set
 - Calls `rspamd_stat_check_autolearn` in `RSPAMD_TASK_STAGE_LEARN` directly in `task.c` with the same condition as before.
 - After checking `rspamd_stat_check_autolearn` it falls through to `RSPAMD_TASK_STAGE_LEARN_POST` where it is handled the same as in the current master.
 - Has the advantage, that `rspamd_stat_learn` is only called once when learn is not necessary/allowed and three times as before if a mail is to be learned as spam/ham.
 - Don't like that it requires the switch case to fall through from `RSPAMD_TASK_STAGE_LEARN` to `RSPAMD_TASK_STAGE_LEARN_POST`

Would one of those options work for you?

PR #5072, opened 2024-07-24
Pull Request for OpenSSL 3.0 support

LeftTry:


vstakhov:
Thanks, it is now much more readable.

src/libserver/dkim.c
```diff
@@ -155,8 +156,8 @@ struct rspamd_dkim_key_s {
 	gsize decoded_len;
 	char key_id[RSPAMD_DKIM_KEY_ID_LEN];
 	union {
-		RSA *key_rsa;
-		EC_KEY *key_ecdsa;
+		//RSA *key_rsa;
+		//EC_KEY *key_ecdsa;
```
vstakhov, line 160: We can remove the whole union I suppose to make the code clearer.

LeftTry: We should remove whole code that is connected to `RSA` and `ECDSA` or move it to the #if #endif section?

vstakhov: I think that using EVP methods is much better. It should work with OSSL 1.x as well from what I know. We will break compatibility with OSSL < 1.0 but nobody should care about that crap.

LeftTry: Ok. I'll remove the whole support. Also, should we leave ed25519 in the union or remove it?

vstakhov: In fact, I will iterate on this PR by myself, it seems we can clean up the code massively.

vstakhov:
It seems there are many formatting changes that are not intended. Please use `clang-format` with the settings that are shipped with Rspamd: https://github.com/rspamd/rspamd/blob/master/.clang-format

vstakhov:
I have merged it occasionally. Git is such a messy thing when it comes to multiple remotes...

PR #5082, opened 2024-07-29
[Minor] milter_headers: x-virus: configurable default value

fatalbanana:
Potentially useful to make it work with other symbols which might not have options.

Inability to configure multiple instances of a routine is a bit limiting ... :(


fatalbanana:
Probably better we should rework this module...

Issue #5095, opened 2024-08-07
[Feature] Ability to encrypt X-Spamd-Result


udf2457:
Adding `X-Spamd-Result` is useful in many scenarios, however, it does introduce the potential of inadvertent "leakage".

Inspired by the fact that the Protonmail equivalent of `X-Spamd-Result` is `X-Pm-Spam` and that value is encrypted, it would be nice if `rspamd` offered the option to encrypt `X-Spamd-Result` with a configured public key.

Clearly all other rspamd related headers would remain plaintext (e.g. `X-Rspamd-Action` etc. etc.).   This feature suggestion is exclusively for  the more detailed `X-Spamd-Result`.

DarshanCode2005:
Respected @vstakhov Sir,

I hope you're doing well.

I would like to explore adding an option to encrypt the X-Spamd-Result header using a configured public key to prevent inadvertent data leakage. Could you please guide me on the starting steps for implementing this and point me to the relevant parts of the code where changes should be made?

Looking forward to your guidance. Thank you for your time!

PR #5116, opened 2024-08-25
Draft pull request for issue #5095

LeftTry:


lualib/lua_util.lua
```diff
@@ -1291,6 +1291,56 @@ exports.maybe_obfuscate_string = function(subject, settings, prefix)
   return subject
 end
 
+---[[[
+-- @function lua_util.maybe_encode_header(header, settings, prefix)
+-- Encode header with configured public key if enabled in settings.
+-- If header is not set then nil is returned. If pub_key is empty then header is returned.
+-- Supported settings:
+-- * <prefix>_encrypt = false - no need for encryption of a header
+-- * <prefix>_publickey = 'pub_key' - key that is used encode header
```
vstakhov, line 1300: This key is not public, it should be just referenced as a `key`

lualib/lua_util.lua
```diff
@@ -1291,6 +1291,56 @@ exports.maybe_obfuscate_string = function(subject, settings, prefix)
...
+-- @return encoded header
+---]]]
+exports.maybe_encode_header = function(header, settings, prefix)
+  local rspamd_secretbox = require "rspamd_cryptobox_secretbox"
+
+  if not header or header == '' then
+    return header
+  elseif settings[prefix .. '_encrypt'] then
+    local pub_key = settings[prefix .. '_publickey']
```
vstakhov, line 1309: s/pub_key/key/

lualib/lua_util.lua
```diff
@@ -1291,6 +1291,56 @@ exports.maybe_obfuscate_string = function(subject, settings, prefix)
...
+    if not pub_key or pub_key == '' then
+      return header
```
vstakhov, line 1311: We should also log error here clearly.

src/lua/lua_cryptobox.c
```diff
@@ -989,20 +989,24 @@ rspamd_lua_ssl_hmac_create(struct rspamd_lua_cryptobox_hash *h, const EVP_MD *ht
...
+	EVP_MAC *mac = EVP_MAC_fetch(NULL, "HMAC", NULL);
+	h->content.hmac_c = EVP_MAC_CTX_new(mac);
 #endif
 	h->out_len = EVP_MD_size(htype);
 
 #if OPENSSL_VERSION_NUMBER > 0x10100000L
 	if (insecure) {
 		/* Should never ever be used for crypto/security purposes! */
 #ifdef EVP_MD_CTX_FLAG_NON_FIPS_ALLOW
-		HMAC_CTX_set_flags(h->content.hmac_c, EVP_MD_CTX_FLAG_NON_FIPS_ALLOW);
```
vstakhov, line 1000: Why this is removed? It was needed for some particular appliance afair.

src/lua/lua_cryptobox.c
```diff
@@ -1452,7 +1458,10 @@ lua_cryptobox_hash_finish(struct rspamd_lua_cryptobox_hash *h)
 		memcpy(h->out, out, ssl_outlen);
 		break;
 	case LUA_CRYPTOBOX_HASH_HMAC:
-		HMAC_Final(h->content.hmac_c, out, &ssl_outlen);
+		size_t ssl_outlen_size_t = ssl_outlen;
```
vstakhov, line 1461: Why do we have this extra variable?

vstakhov (commented):
I'd also suggest not to use the word `encode/decode` for operations that are, in fact, `encrypt/decrypt`.

LeftTry:
I decided to rewrite error messages for `cryptobox:encrypt` and  `cryptobox:decrypt`, because I found them not quite informative. Also, I found and typo in the function description. I changed position of description lines of `input` and `nonce` arguments and I think it is also need to be changed in the docs at rspamd.com.

src/lua/lua_cryptobox.c
LeftTry: Because `EVP_MAC_final` has a `size_t* outl` argument while `ssl_outlen` is `unsigned int`. That is why I am converting it to the size_t using temporary variable.

vstakhov: Yes, but maybe it's betted to have `ssl_outlen` to be `size_t` then?

LeftTry: Yes, it will be probably be better. Thank you!

src/lua/lua_cryptobox.c
```diff
@@ -986,13 +986,14 @@ rspamd_lua_ssl_hmac_create(struct rspamd_lua_cryptobox_hash *h, const EVP_MD *ht
...
-	OSSL_LIB_CTX *libctx = OSSL_LIB_CTX_new();
+	char* properties = malloc(256);
+	strcpy(properties, "provider=default");
 #if OPENSSL_VERSION_NUMBER > 0x10100000L
 	if (insecure) {
 		/* Should never ever be used for crypto/security purposes! */
 #ifdef EVP_MD_CTX_FLAG_NON_FIPS_ALLOW
 		//HMAC_CTX_set_flags(h->content.hmac_c, EVP_MD_CTX_FLAG_NON_FIPS_ALLOW);
-		OSSL_PROVIDER_load(libctx, "fips");
+		strcpy(properties, "provider=fips");
```
vstakhov, line 996: `properties = "provider=fips"`

src/lua/lua_cryptobox.c
```diff
@@ -1001,10 +1002,9 @@ rspamd_lua_ssl_hmac_create(struct rspamd_lua_cryptobox_hash *h, const EVP_MD *ht
 	(defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x30500000)
 	h->content.hmac_c = g_malloc0(sizeof(*h->content.hmac_c));
 #else
-	EVP_MAC *mac = EVP_MAC_fetch(libctx, "HMAC", NULL);
+	EVP_MAC *mac = EVP_MAC_fetch(NULL, "HMAC", properties);
```
vstakhov, line 1005: Please note, that these functions are ONLY for OpenSSL 3.0+. We need to be compatible with OpenSSL 1.0 and LibreSSL, so we can't unconditionally use OpenSSL 3.0 functions - it must be guarded with `#ifdef`. And yes, OpenSSL sucks, I know.

src/lua/lua_cryptobox.c
```diff
@@ -985,24 +999,51 @@ rspamd_lua_ssl_hmac_create(struct rspamd_lua_cryptobox_hash *h, const EVP_MD *ht
...
 	h->content.hmac_c = g_malloc0(sizeof(*h->content.hmac_c));
+#else
+#if OPENSSL_VERSION_MAJOR >= 3
+	EVP_MAC* mac = EVP_MAC_fetch(NULL, "HMAC", NULL);
+	h->content.hmac_c = EVP_MAC_CTX_new(mac);
+	EVP_MAC_free(mac);
 #else
 	h->content.hmac_c = HMAC_CTX_new();
 #endif
-	h->out_len = EVP_MD_size(htype);
```
vstakhov, line 994: Isn't it needed for 1.0 code path?

src/lua/lua_cryptobox.c
```diff
@@ -1438,21 +1486,25 @@ lua_cryptobox_hash_finish(struct rspamd_lua_cryptobox_hash *h)
...
+	size_t ssl_outlen = sizeof(out);
 
 	switch (h->type) {
 	case LUA_CRYPTOBOX_HASH_BLAKE2:
 		rspamd_cryptobox_hash_final(h->content.h, out);
 		memcpy(h->out, out, sizeof(out));
 		break;
 	case LUA_CRYPTOBOX_HASH_SSL:
-		EVP_DigestFinal_ex(h->content.c, out, &ssl_outlen);
+		EVP_DigestFinal_ex(h->content.c, out, (unsigned int *) &ssl_outlen);
```
vstakhov, line 1497: We cannot cast pointer of `size_t*` to `unsigned int*`, as it will likely be corrupted. Please use a dedicated `unsigned int` variable. I know it's annoying but I see no other options.

src/lua/lua_cryptobox.c
```diff
@@ -985,24 +999,51 @@ rspamd_lua_ssl_hmac_create(struct rspamd_lua_cryptobox_hash *h, const EVP_MD *ht
 {
 	h->type = LUA_CRYPTOBOX_HASH_HMAC;
 
+#if OPENSSL_VERSION_NUMBER > 0x10100000L
+	if (insecure) {
+		/* Should never ever be used for crypto/security purposes! */
+#ifdef EVP_MD_CTX_FLAG_NON_FIPS_ALLOW
+#if OPENSSL_VERSION_MAJOR >= 3
+		OSSL_PROVIDER *fips = OSSL_PROVIDER_load(NULL, "fips");
```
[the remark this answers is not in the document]
LeftTry: From what I got from the docs, it is enough just to load provider itself and the operations that will need this provider will automatically use it.

src/lua/lua_cryptobox.c
LeftTry: This is implemented a bit below.

lualib/lua_util.lua
```diff
@@ -1291,6 +1291,84 @@ exports.maybe_obfuscate_string = function(subject, settings, prefix)
...
+    local result = false
+    local header = ''
+    if not nonce then
+      result, header = cryptobox:decrypt(encrypted_header, settings[prefix .. '_nonce'])
+    else
+      result, header = cryptobox:decrypt(encrypted_header, nonce)
+    end
+
+    if not result then
+      logger.errx(rspamd_config, "Decryption is failed with result: %s and decrypted header: %s", result, header)
```
vstakhov, line 1364:
```suggestion
      logger.infox(rspamd_config, "Decryption is failed with result: %s and decrypted header: %s", result, header)
```

vstakhov (approved):
Looks good to me, thank you!

dragoangel:
Maybe it's good to add also information how to decrypt such encrypted headers outside of rspamd, for example in python?

vstakhov:
> Maybe it's good to add also information how to decrypt such encrypted headers outside of rspamd, for example in python?

That sounds like a good idea! We need 2 things - a small `rspamadm` utility to decrypt stuff and an example in `python` + https://pypi.org/project/PyNaCl/ to do the same. @LeftTry could you please draft that as well?

LeftTry:
> > Maybe it's good to add also information how to decrypt such encrypted headers outside of rspamd, for example in python?
> 
> That sounds like a good idea! We need 2 things - a small `rspamadm` utility to decrypt stuff and an example in `python` + https://pypi.org/project/PyNaCl/ to do the same. @LeftTry could you please draft that as well?

Yes, I can. Should I do it in this pull request or should I open another one?

vstakhov:
Let's do it in another one, I'll merge this one.

PR #5132, opened 2024-09-11
Draft pull request for PR#5116

left-try:
I added the rspamadm utility to decrypt the header using key and nonce. I also added a python script with two functions for encrypting and decrypting the header and added a small test function to check the correctness of the previous functions.


src/rspamadm/commands.c
```diff
@@ -40,6 +41,7 @@ const struct rspamadm_command *commands[] = {
 	&fuzzyconvert_command,
 	&signtool_command,
 	&lua_command,
+    	&decrypt_header_command,
```
vstakhov, line 44: Please refrain from adding rspamadm utilities in C. The proper way is to do it in Lua - you can check `lualib/rspamadm` for examples.

lualib/rspamadm/maybe_decrypt_header.lua
```diff
@@ -0,0 +1,20 @@
+local util = require 'lua_util'
```
vstakhov, line 1: We need a fully documented utility here with `argparse` and all stuff around.

utils/encrypt_decrypt_header.py
```diff
@@ -0,0 +1,26 @@
+from operator import truediv
```
vstakhov, line 1: Can we have some interoperability tests? E.g. encrypt via lua and decrypt via python? Might need functional tests for that and extra dependency though.

vstakhov (changes requested):

lualib/rspamadm/decrypt_header.lua
```diff
@@ -0,0 +1,46 @@
...
+    :description "Decrypt given header with given key and nonce"
+    :help_description_margin(32)
+
+parser:option "-H --header"
+      :description("Encrypted header")
+      :argname("<header>")
+parser:option "-k --key"
+      :description("Key used to encrypt header")
+      :argname("<key>")
+parser:option "-n --nonce"
```
vstakhov, line 15: It looks good, but can we just use a concatenation of `nonce` + `encrypted data` as we will likely have? Of course we should have an ability to decrypt/encrypt with/without embedded nonce.

vstakhov (changes requested):
I would also like to have an ability to `encrypt` as well as `decrypt` in this command. So it should be probably named something like `secretbox` and operates in two modes:

`rspamadm secretbox encrypt`
`rspamadm secretbox decrypt`

Header is useless in this context, as we actually operate with just strings. Prefix is also meaningless - we should get rid of them or allow to ignore any prefix (e.g. by taking last X bytes).

LeftTry:
When I was doing compatibility tests manually, I found that secretbox in python and `rspamd_secretbox` work differently. Using the same key and a one-time value for both of them gives different results. Of course, reusing the same 'nonce' and the same 'key' in both secretboxes gives the same result as the first time. Maybe I don't understand something important about secretbox, but at the moment it seems like a big problem to me.

lualib/lua_util.lua
```diff
@@ -1322,7 +1322,8 @@ exports.maybe_encrypt_header = function(header, settings, prefix)
     else
       encrypted_header = cryptobox:encrypt(header, nonce)
     end
-    return encrypted_header, nonce
+    local encrypted_header_with_nonce = nonce .. encrypted_header
+    return encrypted_header, nonce, encrypted_header_with_nonce
```
vstakhov, line 1326: Why do we need that? API should just return a string + nonce. It is a task for a caller to do the remaining.

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,101 @@
+local util = require 'lua_util'
+local rspamd_util = require 'rspamd_util'
+local argparse = require 'argparse'
+
+local parser = argparse()
+    :name "rspamadm encrypt/decrypt text"
```
vstakhov, line 6: `rspamadm secretbox`

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,101 @@
...
+local encrypt = parser:command 'encrypt'
+                      :description 'Encrypt text with given key'
+
+encrypt:option "-t --text"
+      :description("Text to encrypt")
+      :argname("<text>")
+encrypt:option "-k --key"
+      :description("Key to encrypt text")
+      :argname("<key>")
+encrypt:option "-n --nonce"
```
vstakhov, line 35: Why do we need a nonce for encrypt? It should be a boolean flag: either concat nonce with a string or output two strings.

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,101 @@
...
+
+local function decryption_handler(args)
+    local settings = {
+        prefix = 'dec',
+        dec_encrypt = true,
+        dec_key = args.key
+    }
+    local decrypted_header = ''
+    if(args.nonce ~= 'nil') then
+         decrypted_header = util.maybe_decrypt_header(rspamd_util.decode_base64(args.text), settings, settings.prefix,
```
vstakhov, line 47: Hold on, why is it `base64`?? We should be flexible about encoding: raw, base64, base32, hex etc. Please look how other utilities are organized.

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,101 @@
...
+                rspamd_util.decode_base64(args.nonce))
+    else
+        local nonce = string.sub(args.text, 1, 32)
```
vstakhov, line 50: If that's encoded with base64 this logic is just wrong.

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,101 @@
...
+        print('The decryption failed. Please check the correctness of the arguments given.')
+    end
+end
+
+local function encryption_handler(args)
+    local settings = {
+        prefix = 'dec',
+        dec_encrypt = true,
+        dec_key = args.key,
+        dec_nonce = rspamd_util.decode_base64(args.nonce)
```
vstakhov, line 67: Same thing: we should not ask nonce for encryption.

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    local encrypted_text, _, encrypted_text_with_nonce
```
vstakhov, line 70: No, it will not work this way I think.

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,101 @@
...
+                                                    = util.maybe_encrypt_header(args.text, settings, settings.prefix)
+    if encrypted_text ~= nil then
+        print(rspamd_util.encode_base64(encrypted_text_with_nonce))
```
vstakhov, line 73: Same: we need flexible encoding options.

vstakhov (changes requested):

lualib/rspamadm/secretbox.lua
left-try: I decided to do it as a mutex of flags. I thing it will be pretty convenient to user to manage encodings in this way.

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,156 @@
...
+parser:mutex(
+    parser:flag '-R --raw'
+          :description('Encrypted text(and nonce if it is there) will be given in raw'),
+    parser:flag '-H --hex'
+          :description('Encrypted text(and nonce if it is there) will be given in hex'),
+    parser:flag '-b --base32'
+          :description('Encrypted text(and nonce if it is there) will be given in base32'),
+    parser:flag '-B --base64'
+          :description('Encrypted text(and nonce if it is there) will be given in base64')
+)
```
vstakhov, line 21: 👍

lualib/rspamadm/secretbox.lua
```diff
@@ -0,0 +1,156 @@
...
+        end))
+    end
+
+    local function tohex(str)
+        return (str:gsub('.', function (c)
+            return string.format('%02X', string.byte(c))
+        end))
+    end
+
+    local text_ = text
```
vstakhov, line 64:
```suggestion
    local output = text
```

vstakhov (commented):
It now looks quite good, thank you!

Issue #4906, opened 2024-04-04
[BUG] dmarc record parsing failure


errror:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
When parsing a DMARC DNS record for generating the DMARC report, rspamd fails if rua value contains multiple values seperated by `, ` (a comma followed by spaces). Those values appear at least in the records of `disneyplus.com` and `doctolib.de`.

### Steps to Reproduce

1. Get a mail from `disneyplus.com` or `doctolib.de`
2. Generate dmarc reports with `rspamadm dmarc_report`
3. rspamd reports `invalid rua url: "null""`
4. report is only sent to the first rua recipient, not to any secondary

**Expected behavior**
DMARC reports are sent to all recipients.

### Versions

```
Rspamd daemon version 3.8.4

CPU architecture x86_64; features: avx2, avx, sse2, sse3, ssse3, sse4.1, sse4.2, rdrand
Hyperscan enabled: TRUE
Jemalloc enabled: TRUE
LuaJIT enabled: TRUE (LuaJIT version: LuaJIT 2.1.1700008891)
ASAN enabled: FALSE
BLAS enabled: FALSE
Fasttext enabled: FALSE
```

### Additional Information

I already traced the problem down to `gen_dmarc_grammar()`. It seems, the generated grammar does not handle spaces followed by commas in rua values (most like likely other values, too) correctly. According to rfc7489, whitespace after commas in rua and ruf values are allowed.

vstakhov:
Hum, `gen_dmarc_grammar` does not work with commas at all.

vstakhov:
Oh, it works with spaces though.

spacefreak86:
Today I have found one more issue.

[...]

PR #5139, opened 2024-09-12
[Fix] Rework DMARC grammar to correctly handle spaces in values

spacefreak86:
Issue: #4906

According to [RFC7489 section 6.3](https://www.rfc-editor.org/rfc/rfc7489#section-6.3) and [RFC6376 section 3.2](https://www.rfc-editor.org/rfc/rfc6376.html#section-3.2), white space around tags is allowed.
But in particular, any white space after the "=" and any white space before the terminating ";" is not part of the value.
However, white space inside the value is significant.

This PR fixes the DMARC grammar to parse the values correctly.

### DIFF
--- a/lualib/plugins/dmarc.lua
+++ b/lualib/plugins/dmarc.lua
@@ -241,10 +241,10 @@ local function gen_dmarc_grammar()
   local space = lpeg.space ^ 0
   local name = lpeg.C(lpeg.alpha ^ 1)
   local sep = space * lpeg.S("\\;") * space
-  local value = lpeg.C(lpeg.P(lpeg.P((lpeg.space ^ 1) * lpeg.graph + lpeg.P(lpeg.graph)) - sep) ^ 1)
+  local value = lpeg.C(((lpeg.space ^ 1 * lpeg.graph + lpeg.graph) - sep) ^ 1)
   local pair = lpeg.Cg(name * "=" * space * value) * sep ^ -1
   local list = lpeg.Cf(lpeg.Ct("") * pair ^ 0, rawset)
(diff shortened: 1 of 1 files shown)


vstakhov:
The question here is simple: implementation vs RFC. I have seen so many broken DNS records for the **legit** senders, so I'm very reluctant if I'm suggested to blindly follow RFC. What problem are we solving here? AFAIU Rspamd just ignores all spaces in DMARC records and it seems to be the most safe decision if it comes to the real life Internet. However, I'm open to hear some motivation here aside of 'do as RFC states' - are there any records that are space sensitive?

spacefreak86:
Please take a look at what I have written in the comments of #4906 (last two comments).
This PR fixes exactly those two issues. My grammar just parses the values correctly, but is not more sensitive than it was before.

E.g. Rspamd fails to parse this record: "v=DMARC1; p=quarantine ;".
The resulting error message: "p tag has invalid value: quarantin".

This is wrong in my eyes because the DMARC record is 100% RFC conform and the "e" of the word quarantine gets cut off falsely.

vstakhov:
I see, it is a good thing then. Maybe we should add some asserts for the grammar or move grammar to `lualib/plugins/dmarc` and write some proper unit tests for all that stuff. No pressure on that though, I can do it by myself.

spacefreak86:
Thanks man, always a pleasure to work with you. Rspamd is a great project!

spacefreak86:
Sorry to bother you again, but I just noticed the unnecessary calls of lpeg.P and wanted to clean that up.
BTW: I always test this changes for a while on my productive system with a throughput of ~60K messages per day before I commit.

vstakhov:
Thank you very much!

Issue #5152, opened 2024-09-23
[BUG] Rspamd fails to build on ubuntu-focal after SSL changes


vstakhov:
OpenSSL are **constantly** breaking API, so it's important to have some builds in CI that test it.

```
[100%] Linking CXX executable rspamadm
../librspamd-server.so: error: undefined reference to 'EVP_PKEY_CTX_set1_rsa_keygen_pubexp'
../librspamd-server.so: error: undefined reference to 'EVP_PKEY_dup'
../librspamd-server.so: error: undefined reference to 'EVP_PKEY_get_size'
../librspamd-server.so: error: undefined reference to 'EVP_MD_CTX_get0_md'
```

PR #5153, opened 2024-09-23
Add more build targets

vstakhov:
Issue: #5152


vstakhov:
Ok, we have now all that stuff: need to fix RPM containers to have luajit and we're done.

PR #5166, opened 2024-09-30
Check if url is redirector filtered by tld in url_redirector module

dragoangel:
Align configuration with redirector multimap

### DIFF
--- a/src/plugins/lua/url_redirector.lua
+++ b/src/plugins/lua/url_redirector.lua
@@ -348,8 +348,8 @@ local function url_redirector_handler(task)
     task = task,
     limit = settings.max_urls,
     filter = function(url)
-      local host = url:get_host()
-      if settings.redirector_hosts_map:get_key(host) then
+      local tld = url:get_tld()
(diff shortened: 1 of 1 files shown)


dragoangel:
Hi @vstakhov the problem I see:

as I understand TLD is checking all known TLDs and add +1 level, f.e.:
- eample.com
- example.org.ua

`click.email4customers.com` - not a tld, as `email4customers.com` is, and `settings.redirector_hosts_map:get_key(tld)` will not match, as result it will not work

and opposite, we want think that `*.safelinks.protection.outlook.com` is redirector, but `outlook.com` - not.

Maybe better take some "glob" or something like that here and in multimap?

fatalbanana:
> Maybe better take some "glob" or something like that here and in multimap?

We could use `glob` type for map and add entries like `*.example.net` where desirable.

dragoangel:
Yeah, this what I propose, but then need to write the filter function a bit differently here + need to understand that people who not updating will use old map as it was, so it's had to be a "new" map*, with different name for transaction period

Issue #5163, opened 2024-09-30
[BUG] 3.10.0 fails to compile with gcc 14


darix:
This error is new with  3.10.0. 3.9.1 still worked fine.

```
 /home/abuild/rpmbuild/BUILD/rspamd-3.10.0/contrib/libucl/lua_ucl.c: In function ‘lua_ucl_newindex’:  
 /home/abuild/rpmbuild/BUILD/rspamd-3.10.0/contrib/libucl/lua_ucl.c:1430:63: error: passing argument 3 of ‘lua_tolstring’ from incompatible pointer type [-Wincompatible-pointer-types]
  1430 |                         const char *key = lua_tolstring(L, 2, &keylen);
       |                                                               ^~~~~~~
       |                                                               |
       |                                                               lua_Integer * {aka long long int *}
 In file included from /home/abuild/rpmbuild/BUILD/rspamd-3.10.0/contrib/libucl/lua_ucl.h:52,
                  from /home/abuild/rpmbuild/BUILD/rspamd-3.10.0/contrib/libucl/lua_ucl.c:46:
 /usr/include/lua5.4/lua.h:204:73: note: expected ‘size_t *’ {aka ‘long unsigned int *’} but argument is of type ‘lua_Integer *’ {aka ‘long long int *’}
   204 | LUA_API const char     *(lua_tolstring) (lua_State *L, int idx, size_t *len);
       |                                                                 ~~~~~~~~^~~
 /home/abuild/rpmbuild/BUILD/rspamd-3.10.0/contrib/libucl/lua_ucl.c:1543:63: error: passing argument 3 of ‘lua_tolstring’ from incompatible pointer type [-Wincompatible-pointer-types]
  1543 |                         const char *key = lua_tolstring(L, 2, &keylen);
       |                                                               ^~~~~~~
       |                                                               |
       |                                                               lua_Integer * {aka long long int *}
 /usr/include/lua5.4/lua.h:204:73: note: expected ‘size_t *’ {aka ‘long unsigned int *’} but argument is of type ‘lua_Integer *’ {aka ‘long long int *’}
   204 | LUA_API const char     *(lua_tolstring) (lua_State *L, int idx, size_t *len);
       |                                                                 ~~~~~~~~^~~
```

SoLoR1:
[...]

PR #5169, opened 2024-10-02
[Fix] Use correct type for keylen in lua_ucl_newindex

arkamar:
The keylen variable used in `lua_ucl_newindex` function should use `size_t` type instead of `lua_Integer`, because all functions that use `keylen` expect it to be of `size_t` type. This mismatch leads to incompatible pointer types, and modern versions of GCC fail to compile the code.

Fixes: 9e87597ceb05 ("[Project] Allow manipulations with opaque UCL objects")
Closes: #5163


vstakhov:
Thank you, that's the best way to fix issues! Much appreciated.

Issue #5149, opened 2024-09-22
Store most reputation full elements for `reputation` plugin


vstakhov:
We want to store top positive and negative elements for reputation. Better to have 2 separate keys for positive and negative entries with different cardinality limits.

kevinzb56:
@vstakhov 
Is this issue still open?

PR #5183, opened 2024-10-10
Store most reputation full elements for reputation plugin

left-try:
Issue: #5149


vstakhov:
Could you please describe your PRs properly and not use "PR for issue XXX", as it makes navigation particularly difficult.

Should you want to link that issue, just use something like `Issue: #5149` in the PR description. In any case, your PR description should really describe the task you are working on.

left-try:
I was working on the issue #5149 
It is about storing top positive and negative elements for reputation.

src/plugins/lua/reputation.lua
```diff
@@ -992,6 +995,19 @@ local function reputation_redis_init(rule, cfg, ev_base, worker)
...
+    local top_name = '{= neg_top =}'
+    if scores[i] >= 0 then
+      top_name = '{= pos_top =}'
+      top_size_max = {= pos_top_max =}
+    end
+    local top_size = redis.call('ZCARD', top_name)
+    if top_size > top_size_max then
+      redis.call('ZREMRANGEBYRANK', top_name, 0, -(top_size_max + 1))
+    end
+    redis.call('ZADD', KEYS[1] .. top_name, math.abs(scores[i]), '{= w.name =}')
```
vstakhov, line 1008: You add to list named `KEYS[1]..top_name` but clean and do all other stuff with `top_name`. It looks like a mistake to me.

I think we should add `KEYS[1]` to zlist named `top_name` with the appropriate score. If we have multiple windows it's a problem but from what I see that's the only way to achieve the goal.

vstakhov (changes requested):

src/plugins/lua/reputation.lua
left-try: Oh, ok. I just thought that user will know which key is he working with. I'll change that.

left-try: It might be better to keep it the way I'm doing it now, but also have two `zsets` (for negative and positive) `keys` and scores as the sum of all the window scores that are associated with this `key` that are stored in the `KEYS[1] .. top_name`. What do you think?

vstakhov (approved):

PR #5186, opened 2024-10-11
[Rework] Breaking: Actualize elastic module

dragoangel:
- actualize elastic module to support Elastic version 8 and OpenSearch version 2
- add version verification and protection from unsupported versions of elastic
- move fully to periodic tasks for handling sending data to elastic
- add limit based on size of the buffer
- change index template management from static json file to dynamic rendering inside of the module based on user preferences and distro options
- allow to disable index template management from the module
- add index policy management from the module with option to disable it
- allow disable geoip or turn off pipeline management from the module
- allow configure custom headers to catch and render them in index template automatically, if it managed
- refactor error handling to functions
- other small codes improvements
- refactor top-level local variables and settings structure

*BREAKING*: added enabled = false to the module because, before using it - users should reconfigure their setup, as default behavior is to add index policy with enabled deletion of logs older then 30 days.

Note: I deleted Kibana resources and upload logic as it's totally obsolete, I can't get it working even on 6th version of Elastic, nor manually nor automatically, not speaking about newer versions. Kibana should be managed via Kibana URL, nor Elastic API in newer versions, I don't think this should be part of module. If somebody want to create pre-made dashboards we can add them later to the docs for manual one-time import and I can re-look to automating it.


dragoangel:
Just to note: I will update docs if this will be merged.

dragoangel:
@vstakhov I would be very happy if you can advise how to shutdown the module properly and slow-down in some places (on retrying). Unfortunately I can't find a good way.

dragoangel:
If you want to test this setup, here is a quick compose to play with:
 - opensearch 2 (elastic 8 alternative, https ports: 9201 & 5602, default pass: `random_string_Z2oiYn7xxrGMn9BfYbRY`)
 - elastic 7 (http ports: 9200 & 5601, no password):
```yaml
  opensearch:
    image: opensearchproject/opensearch:2
    restart: unless-stopped
    environment:
      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_INITIAL_ADMIN_PASSWORD:-random_string_Z2oiYn7xxrGMn9BfYbRY}
      - OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m
      - discovery.type=single-node
      - bootstrap.memory_lock=true
    ports:
      - 9201:9200
    volumes:
      - opensearch-data:/usr/share/opensearch/data
    healthcheck:
      test: [ "CMD", "curl", "-f", "http://localhost:9200/_cat/health" ]
      interval: 30s
      timeout: 10s
      retries: 3
    ulimits:
      memlock:
        soft: -1
        hard: -1
      nofile:
        soft: 65536
        hard: 65536
    cap_add:
      - IPC_LOCK

  opensearch-dashboards:
    image: opensearchproject/opensearch-dashboards:2
    restart: unless-stopped
    environment:
      - OPENSEARCH_HOSTS=["https://opensearch:9200"] # array should not contain spaces
    ports:
      - 5602:5601


  elastic:
    image: elasticsearch:7.17.6
    restart: unless-stopped
    environment:
      - xpack.security.enabled=false
      - discovery.type=single-node
      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
    volumes:
      - elasticsearch-data:/usr/share/elasticsearch/data
    healthcheck:
      test: [ "CMD", "curl", "-f", "http://localhost:9200/_cat/health" ]
      interval: 30s
      timeout: 10s
      retries: 3
    ulimits:
      memlock:
        soft: -1
        hard: -1
      nofile:
        soft: 65536
        hard: 65536
    cap_add:
      - IPC_LOCK

  kibana:
    image: kibana:7.17.6
    restart: unless-stopped
    environment:
      - ELASTICSEARCH_HOSTS=http://elastic:9200
    ports:
      - 5601:5601

volumes:
  opensearch-data:
    driver: local
  elasticsearch-data:
    driver: local

networks:
  default:
    ipam:
      config:
        - subnet: 192.168.28.0/24
```

```

File: src/plugins/lua/elastic.lua [hunk 1]
```diff
@@ -239,257 +635,918 @@ local function get_general_metadata(task)
...
+  local parts = task:get_text_parts()
+  local lang_t = {}
+  if parts then
+    for _, part in ipairs(parts) do
+        local l = part:get_language()
+        if l and not contains(lang_t, l) then
+          table.insert(lang_t, l)
+        end
+    end
+    if table.getn(lang_t) > 0 then
```
vstakhov, line 649:
```suggestion
    if #lang_t > 0 then
```

src/plugins/lua/elastic.lua
```diff
@@ -239,257 +635,918 @@ local function get_general_metadata(task)
...
+      r.language = lang_t
+    else
+      r.language = empty
+    end
+    if table.getn(lang_t) == 1 and lang_t[1] == 'en' then
```
vstakhov, line 654:
```suggestion
    if #lang_t == 1 and lang_t[1] == 'en' then
```

src/plugins/lua/elastic.lua
```diff
@@ -239,257 +635,918 @@ local function get_general_metadata(task)
...
+  local fuzzy_hashes = task:get_mempool():get_variable('fuzzy_hashes', 'fstrings')
+  if fuzzy_hashes and #fuzzy_hashes > 0 then
+    local l = {}
+    for _, h in ipairs(fuzzy_hashes) do
+      table.insert(l, h)
+    end
+    r.fuzzy_hashes = l
+  else
+    r.fuzzy_hashes = empty
+  end
```
vstakhov, line 670:
```suggestion
r.fuzzy_hashes = fuzzy_hashes or empty
```

src/plugins/lua/elastic.lua
```diff
@@ -239,257 +635,918 @@ local function get_general_metadata(task)
...
+
+  r.received_delay = 0
```
vstakhov, line 672: It's an interesting option - maybe we need it in the Clickhouse. Probably we should just share this code between modules?

src/plugins/lua/elastic.lua
```diff
@@ -239,257 +635,918 @@ local function get_general_metadata(task)
...
+  if user then -- calculate received_delay only for incoming traffic
+    local recieved_hop = 2
+    local received_headers = task:get_received_headers()
+    if received_headers[recieved_hop] then
+      if received_headers[recieved_hop]['timestamp'] then
+        r.received_delay = math.floor(rspamd_util.get_time()) - received_headers[recieved_hop]['timestamp']
```
vstakhov, line 678: I'm not sure it's correct as there are timezones in the world.

src/plugins/lua/elastic.lua
dragoangel: Not see anything bad in sharing it to Clickhouse

src/plugins/lua/elastic.lua
dragoangel: As as saw it counts timezones

src/plugins/lua/elastic.lua
```diff
@@ -18,219 +18,615 @@ limitations under the License.
...
+          rspamd_logger.debugm(N, log_object, 'successfully sent payload with %s logs', nlogs_to_send)
+        else
+          push_errors = true
+          for _, value in pairs(obj['items']) do
+            if value['index']['status'] >= 400 then
+              if value['index']['error'] then
+                if value['index']['error']['type'] and value['index']['error']['reason'] then
+                  rspamd_logger.errx(log_object,
+                    'cannot send logs to elastic (%s) due to error: %s status, %s type, due to: %s; failed attempts: %s/%s',
+                    push_url, value['index']['status'], value['index']['error']['type'], value['index']['error']['reason'],
```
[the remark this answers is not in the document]
dragoangel: sometimes I need access variable like value[variable], so it was more convince to me use this syntax same

src/plugins/lua/elastic.lua
dragoangel: Double checked - zones is part of timestamp.

vstakhov (approved):

Issue #5173, opened 2024-10-04
ARC seal failed since 1 october


srsr:
Hello,

Since october 1st, forwarded mails are no longer sealed. I haven't noticed anything wrong with the dns, the spf/dkim is operational.
On the other hand, rspamd is updated automatically and the last update also dates from October 1st.
Is my problem linked to this update?

Here are the rspamd logs containing the sealing error

`
2024-10-04 11:04:08 #16511(rspamd_proxy) <5361B7>; proxy; rspamd_task_write_log: id: <CAFToh6_sq6EggAuFT7F_jEEeRGiEW2YRjdav0GsC+hvwCoqkBg@mail.gmail.com>, qid: <5361B7F846>, ip: 192.168.128.79, user: serveur.mail@AxxxxE.IxxxxA, from: <xxxxxx@gmail.com>, (default: F (no action): [2.40/15.00] [ARC_REJECT(2.00){seal check failed: fail, {[1] = sig:aXXXe.fr:reject};},R_PARTS_DIFFER(0.50){100.0%;},MIME_GOOD(-0.10){multipart/alternative;text/plain;},ARC_SIGNED(0.00){aXXXe.fr:s=arc:i=2;},DKIM_SIGNED(0.00){aXXXe.fr:s=dkim;},FORWARDED(0.00){samuel.test@aXXXe.fr;},FREEMAIL_ENVFROM(0.00){gmail.com;},FREEMAIL_FROM(0.00){gmail.com;},FROM_EQ_ENVFROM(0.00){},FROM_HAS_DN(0.00){},MID_RHS_MATCH_FROMTLD(0.00){},MIME_TRACE(0.00){0:+;1:+;2:~;},MISSING_XM_UA(0.00){},PREVIOUSLY_DELIVERED(0.00){samuel.test@aXXXe.fr;},RCPT_COUNT_ONE(0.00){1;},RCVD_COUNT_FIVE(0.00){5;},RCVD_TLS_LAST(0.00){},RCVD_VIA_SMTP_AUTH(0.00){},TO_DN_NONE(0.00){}]), len: 6322, time: 353.661ms, dns req: 6, digest: <e750b52d6047d1cb54468b902dca62c3>, rcpts: <test-4k4zgbd35@srv1.mail-tester.com>, mime_rcpts: <samuel.test@aXXXe.fr>
[...]

PR #5193, opened 2024-10-19
[CritFix] Fix ARC-Seal signing

jscissr:
Signing of ARC-Seal headers was recently broken; the created signatures failed to validate. Most likely, this was caused by commit 1e661a2fc6e3, which changed the way signatures are created in lua_rsa_sign_memory without adding the calls to EVP_PKEY_CTX_set_rsa_padding and EVP_PKEY_CTX_set_signature_md needed with the new interface.

After fixing this, some existing tests failed, because the test values passed to the hash parameter did not have the correct size for a sha256 hash. I fixed these by adjusting the length of the test values. Additionally, I extended the "RSA sign" unit test to compare the created signature against the expected one. This is possible because RSA signing is deterministic, and should prevent the same bug from occuring again.

Fixes: https://github.com/rspamd/rspamd/issues/5173


vstakhov:
This is very serious issue, thank you for finding it and fixing! I will release a new version asap.

PR #5195, opened 2024-10-24
[Project] Keep Changelog updated

bocan:
This seems to have fallen behind and I _use_ it when building a .deb package for Debian Sid.

I guessed at the categories.


vstakhov:
I usually do it before cutting a new release from the master branch. However, I see no issues with keeping it up to date indeed.

Issue #5148, opened 2024-09-22
Add `rspamadm` tool to manage last ratelimit hits


vstakhov:
We want to track last limits, modify buckets, unblock buckets etc.

PR #5216, opened 2024-11-07
Add `rspamadm` tool to manage last ratelimit hits

left-try:
This PR refers to issue #5148. Created `rspamadm` util to manage last filled `ratelimit` buckets.


lualib/rspamadm/ratelimit.lua
```diff
@@ -0,0 +1,202 @@
...
+local track_limits = parser:command 'track_limits'
+                           :description 'Track last limits of ratelimit module'
+
+track_limits:option "-q --quantity"
+            :description("Number of limits to track")
+            :argname("<quantity>")
+            :default(1)
+
+
+local upgrade_bucket = parser:command 'upgrade_bucket'
```
vstakhov, line 21:
```suggestion
local upgrade_bucket = parser:command 'upgrade'
```

lualib/rspamadm/ratelimit.lua
```diff
@@ -0,0 +1,202 @@
...
+              :args("?")
+
+local unblock_bucket = parser:command 'unblock_bucket'
+                             :description 'Unblock certain bucket'
+
+unblock_bucket:argument "prefix"
+              :description("Prefix of bucket to operate with")
+              :args(1)
+
+local unblock_buckets = parser:command 'unblock_buckets'
```
vstakhov, line 55: Why can't it be an option for `unblock`?

lualib/rspamadm/ratelimit.lua
```diff
@@ -0,0 +1,202 @@
...
+track_limits:option "-q --quantity"
+            :description("Number of limits to track")
+            :argname("<quantity>")
+            :default(1)
+
+
+local upgrade_bucket = parser:command 'upgrade_bucket'
+                             :description 'Upgrade certain bucket'
+
+upgrade_bucket:argument "prefix"
```
vstakhov, line 24: How a user should get this prefix?

.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ on:
       - master
   push:
     branches:
-      - master
+      - '**'
```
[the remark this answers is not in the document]
fatalbanana: The idea might be to make it possible to test feature branches ahead of PR'ing them; I've felt that pain.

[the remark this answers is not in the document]
left-try: Yes. I used it to test my branches. I think it could be great to use it in main project.

lualib/rspamadm/ratelimit.lua
left-try: He can get it from the first command. So i created another command to unblock plenty of buckets in a time. And, of course, it could be merged with this command, but I thought it would not be very user friendly to use the same command for multiple update and single update(that uses specific prefix) in one command

lualib/rspamadm/ratelimit.lua
left-try: He can get it from the first command. So i created another command to unblock plenty of buckets in a time. And, of course, it could be merged with this command, but I thought it would not be very user friendly to use the same command for multiple update and single update(that uses specific prefix) in one command

vstakhov (approved):

PR #5288, opened 2025-01-10
[Enhancement] Expand Detection of Fake Reply Subjects Across Multiple Languages

D0LLYNH0:
Expand subject detection for fake replies by adding abbreviations from multiple languages: Dutch/German (Antw), Latvian/Welsh (ATB), German (AW), Indonesian (BLS), Polish (ODP), Portuguese (RES), Italian (RIF), Danish/Icelandic/Norwegian/Swedish (SV), and Turkish (YNT).


dragoangel:
@vstakhov maybe then we can add also:

- Відп. - UA
- Отв. - RU
- Ré - FR

Issue #4883, opened 2024-03-21
[BUG]rbl configurations ipv4 and ipv6 are being ignored for urls checks + resolve_ip


japc:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

Configurations ipv4 and ipv6 for rbls are being ignored for urls checks + resolve_ip:

 It always queries the url for the A and AAAA records (url -> ips), and always issues the RBL queries for the reversed octets for both IPv4 and IPv6 (ips -> rbl), regardless of the configuration.

### Steps to Reproduce

1. Configure a RBL with, for example, ipv6 disabled:

```
   test_rbl {
      symbol = "TEST_RBL";
      enabled = true;
      rbl = "zen.spamhaus.org";
      resolve_ip = true;
      url_full_hostname = true;
      checks = ['urls'];
      ipv4 = true;
      ipv6 = false;
    }
```

**Expected behavior**

Expecting to see no query for AAAA and/or no rbl query for the obtained IPv6 but see both:

```
2024-03-21 09:00:14 #1149252(normal) <3ce83d>; rbl; rbl.lua:973: rbl TEST_RBL; resolve www.lovepop.com -> www.lovepop.com
2024-03-21 09:00:14 #1149252(normal) <3ce83d>; rbl; rbl.lua:953: rbl TEST_RBL; resolve 74.38.227.23 -> 74.38.227.23.zen.spamhaus.org
[...]

PR #5292, opened 2025-01-13
[Minor] Respect ipv4 and ipv6 configurations for rbl resolve_ip

japc:
Referring to issue #4883

### DIFF
--- a/src/plugins/lua/rbl.lua
+++ b/src/plugins/lua/rbl.lua
@@ -983,7 +983,7 @@ local function gen_rbl_callback(rule)
         if req.resolve_ip then
           -- Deal with both ipv4 and ipv6
           -- Resolve names first
-          if rule.ipv4 and r:resolve_a({
+          if (rule.ipv4 == nil or rule.ipv4) and r:resolve_a({
             task = task,
             name = req.n,
             callback = gen_rbl_ip_dns_callback(req),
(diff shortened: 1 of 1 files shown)


vstakhov:
I see some failed tests with that PR.

japc:
Without access to a test environment right now, but it makes sense if neither rule.ipv4 and/or rule.ipv6 are set that "resolve_ip = true" makes SPAMHAUS_ZEN_URIBL fail the tests. Maybe those should be set to true.

japc:
Or is it that both the resolver and the domain 8.8.8.9:example.ru are not working?

japc:
No, ok, it's the "ignore_defaults = true" on the config test block SPAMHAUS_ZEN_URIBL that make ipv4 and ipv6, if unset, as they are in the config block, end up as nil and never any resolve().

This fixes it.

```
@@ -985,3 +985,3 @@
           -- Resolve names first
-          if rule.ipv4 and r:resolve_a({
+          if (rule.ipv4 == nil or rule.ipv4) and r:resolve_a({
             task = task,
@@ -993,3 +993,3 @@
           end
-          if rule.ipv6 and r:resolve('aaaa', {
+          if (rule.ipv6 == nil or rule.ipv6) and r:resolve('aaaa', {
             task = task,
```

Don't know if it's proper fix though.

PR #5313, opened 2025-01-21
fix: use sub_utf8 to strip headers value to not break utf8 strings

dragoangel:
`h.decoded:sub(1, headers_text_ignore_above)` was leading to potential broken utf8 string as it had chance to strip line in the middle of utf8 character leading to errors like:
```
error while pushing logs to elastic, status: 400, type: mapper_parsing_exception, reason: failed to parse field [rspamd_meta.header_to] of type [text] in document with id 'X'. Preview of field's value: ''
```

also I added task digest

### DIFF
--- a/src/plugins/lua/elastic.lua
+++ b/src/plugins/lua/elastic.lua
@@ -449,6 +449,7 @@ local function get_general_metadata(task)
   local empty = settings['index_template']['empty_value']
   local user = task:get_user()
   r.rspamd_server = rspamd_hostname or empty
+  r.digest = task:get_digest() or empty
 
   r.action = task:get_metric_action() or empty
   r.score = task:get_metric_score()[1] or 0
@@ -1278,6 +1279,7 @@ local function configure_index_template(cfg, ev_base)
             type = 'object',
             properties = {
               rspamd_server = t_keyword,
+              digest = t_keyword,
               action = t_keyword,
               score = t_double,
               symbols = symbols_obj,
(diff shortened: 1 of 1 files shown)


src/plugins/lua/elastic.lua
```diff
@@ -608,25 +564,34 @@ local function get_general_metadata(task)
...
+          break
+        end
+        if not header_len or header_len == 0 then
+          table.insert(l, empty)
+          break
+        end
+        if settings['index_template']['headers_text_ignore_above'] > 0 and
+          header_len >= settings['index_template']['headers_text_ignore_above']
+        then
+          header = header:sub_utf8(1, settings['index_template']['headers_text_ignore_above'])
```
fatalbanana, line 592: We don't have this method in LuaJIT; maybe we should be shipping something like that.

dragoangel: Yep, it was added due to such need :)

vstakhov: Beware that it's a method of `rspamd_text` not Lua strings.

dragoangel: @vstakhov yes I understand that, but as far I tested on my local development - all data accepted just fine. First time I was do `table.insert(l, header:str)`, but just of curiosity tested without convertion to `str` and it just worked, so I left it as is to economy resources if it works. Or it could have some hidden rocks that I'm not aware off?

dragoangel: btw - digest is very useful thing on a scale, allows to find unknown yet spam by just doing TOP 100 and some conditions like score < 10 😉

dragoangel:
@vstakhov is all okay with this pr? Can it be merged please?

PR #5322, opened 2025-01-28
[Minor] rspamd_logger extended interface example was using wrong function

japc:
`rspamd_logger.info('a=%1, b=%2, c=%3, d=%4, e=%s', a, b, c, d, e)
`
vs

`rspamd_logger.infox('a=%1, b=%2, c=%3, d=%4, e=%s', a, b, c, d, e)
`

### DIFF
--- a/src/lua/lua_logger.c
+++ b/src/lua/lua_logger.c
@@ -42,7 +42,7 @@ local e = {
 -- %<number> means numeric arguments and %s means the next argument
 -- for example %1, %2, %s: %s would mean the third argument
 
-rspamd_logger.info('a=%1, b=%2, c=%3, d=%4, e=%s', a, b, c, d, e)
+rspamd_logger.infox('a=%1, b=%2, c=%3, d=%4, e=%s', a, b, c, d, e)
 -- Output: a=string, b=1.50000, c=1, d={[1] = aa, [2] = 1, [3] = bb} e={[key]=value, [key2]=1.0}
 
 -- Create string using logger API
--- a/src/lua/lua_logger.c
+++ b/src/lua/lua_logger.c
@@ -42,7 +42,7 @@ local e = {
 -- %<number> means numeric arguments and %s means the next argument
 -- for example %1, %2, %s: %s would mean the third argument
 
-rspamd_logger.infox('a=%1, b=%2, c=%3, d=%4, e=%s', a, b, c, d, e)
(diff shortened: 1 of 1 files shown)


vstakhov:
There is no difference for now between `info` and `infox` functions. They are literally mapped to the same C function.

japc:
It was failing as is in the example so I thought it was the wrong function.

Actually the problem is the %s format after the positional %4. The value of num_arg is never returned to false and so it ends up doing a strtoul to the s, returning 0 and failing with “wrong argument number”.

I am also not understanding very well how the %n and %s mix. For instance for

`rspamd_logger.infox('x=%s a=%1, b=%s, c=%3, d=%4, e=%s, f=%s, g=%1', 'a', 'b', 'c', 'd', 'e', 'f')
`

I got: `x=a a=a, b=a, c=c, d=d, e=d, f=e, g=a`

But was expecting: `x=a a=a, b=c, c=c, d=d, e=f, f=no value, g=a`, where the %s positions translate to their % index (%s=1, %1, %s=3, %3, %4, %s=6,....).

Will update the branch with some changes that may help clarify what I mean.

japc:
Some tests with

```
l = require 'rspamd_logger'
print(l.slog("test: %1 %2 %3 %s", '1', '2', '3', '4'))
print(l.slog("test: %1 %2 %s %4", '1', '2', '3', '4'))
print(l.slog("test: %s %2 %3 %4", '1', '2', '3', '4'))
print(l.slog("test: %1 %4 %s %s", '1', '2', '3', '4'))
```
# 
<ins>Expected output</Ins>:

```
test: 1 2 3 4
test: 1 2 3 4
test: 1 2 3 4
test: 1 4 3 4
```
# 
<ins>Output</ins>:


- present:

```
wrong argument number: 0
wrong argument number: 0
test: 1 2 3 4
wrong argument number: 0
```

Gives the `wrong argument number` error when `%s` follows a `%<number>`.


- with num_arg = false:

```
test: 1 2 3 3
test: 1 2 2 4
test: 1 2 3 4
test: 1 4 4 no value
```

Fixes the previous error on `%s` but `%s` doesn't seem to be positional.


- with num_arg = false and pos_arg:

```
test: 1 2 3 4
test: 1 2 3 4
test: 1 2 3 4
test: 1 4 3 4
```

amulet1:
This pull request has been superseded by PR #5378.

PR #5360, opened 2025-02-27
[Feature] Plugin to integrate with Contextal platform

fatalbanana:
Integrate with [Contextal Platform](https://platform.contextal.com/).

Yield symbols for Contextal's 'actions' tagged with the 'scenario' names, allow those to be used with composites or `force_actions`.

Optionally use Redis for caching.

TODO:

- [x] Support forcing `soft reject` in case results are unavailable
- [ ] Support collecting and processing work graphs (maybe rather not)
- [x] Support request TTL / waiting on results
- [x] Add configuration schema
- [x] Add stock config
- [ ] Other stuff?

### DIFF
--- a/src/plugins/lua/contextal.lua
+++ b/src/plugins/lua/contextal.lua
@@ -97,8 +97,8 @@ local function process_actions(task, obj, is_cached)
       key, -- hash key
       true, -- is write
       redis_set_cb, --callback
-      'SET', -- command
-      { key, ucl.to_format(cache_obj, 'json-compact') } -- arguments
+      'SETEX', -- command
+      { key, settings.cache_ttl, ucl.to_format(cache_obj, 'json-compact') } -- arguments
   )
 
   if not ret then
--- a/src/plugins/lua/contextal.lua
+++ b/src/plugins/lua/contextal.lua
@@ -153,17 +153,22 @@ local function submit(task)
   })
 end
 
+local function cache_hit(task, err, data)
+  if err then
+    rspamd_logger.err(task, 'error getting cache: %s', err)
+  else
+    process_cached(task, data)
+  end
+end
+
 local function submit_cb(task)
-  if redis_params then
-    redis_cache.cache_get(task, task:get_digest(), cache_context, settings.cache_timeout,
+  if cache_context then
+    redis_cache.cache_get(task,
+        task:get_digest(),
+        cache_context,
+        settings.cache_timeout,
         submit,
-        function(task, err, data)
-          if err then
-            rspamd_logger.err(task, 'error getting cache: %s', err)
-          else
-            process_cached(task, data)
-          end
-        end
+        cache_hit
     )
   else
     submit(task)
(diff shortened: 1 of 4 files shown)


src/plugins/lua/contextal.lua
```diff
@@ -0,0 +1,319 @@
...
+end
+
+local function set_url_path(base, path)
+  local slash = base:sub(#base) == '/' and '' or '/'
+  return base .. slash .. path
+end
+
+settings = lua_util.override_defaults(settings, opts)
+
+local res, err = config_schema:transform(settings)
```
vstakhov, line 257: Shouldn't we enrich it with Redis parameters + caching schema as we use caching toolkit?

src/plugins/lua/contextal.lua
```diff
@@ -0,0 +1,319 @@
...
+if not res then
+  rspamd_logger.warnx(rspamd_config, 'plugin %s is misconfigured: %s', N, err)
+  local err_msg = string.format("schema error: %s", res)
+  lua_util.config_utils.push_config_error(N, err_msg)
+  lua_util.disable_module(N, "failed", err_msg)
+  return
+end
+
+contextal_actions = lua_util.list_to_hash(contextal_actions)
```
vstakhov, line 266: I don't quite understand why is it needed? Is it like merging of 2 tables with `key`: `true` elements?

src/plugins/lua/contextal.lua
```diff
@@ -0,0 +1,319 @@
...
+      cache_ttl = settings.cache_ttl,
+      cache_format = 'json',
+      cache_use_hashing = false
+  })
+end
+
+rspamd_config:register_symbol({
+  name = settings.submission_symbol,
+  priority = lua_util.symbols_priorities.top,
+  type = 'prefilter',
```
vstakhov, line 299: Interesting, why do we need that on prefilter stage? Maybe we should do other filtering first to allow, for example, whitelisting or early greylist/ratelimit to avoid expensive checks?

vstakhov:
I'm sorry for such a long delay with the review :(

src/plugins/lua/contextal.lua
fatalbanana: The idea is merely to try execute it early which is a likely misuse.

If `request_ttl` is set (it is by default so far), and that is less than `task_timeout`, then probably we should use that to set a timer instead of using a postfilter.

src/plugins/lua/contextal.lua
fatalbanana: It's just laziness to write it out in the proper format I suppose, I'll fix it

src/plugins/lua/contextal.lua
fatalbanana: Yes, probably so; I'll check that

fatalbanana: Well, the schema seems to be in there somehow; also I don't see that `lua_cache` exports a schema

vstakhov: I think it should, but let it be there for now. In general, the configuration dependent tool (like lua_cache) should export some `enrich_schema` method like `lua_redis` does.

PR #5364, opened 2025-03-02
[WebUI] Remove Glyphicons and fix its last use

jeromelebleu:
Glyphicons has been replaced by FontAwesome - see https://github.com/rspamd/rspamd/pull/3423 - but the font is still loaded and used for one icon. This finishes the replacement and removes Glyphicons fonts and CSS files.


moisseev (changes requested):
- This PR breaks icons in FooTable as it uses Glyphicons.
- The "Reload" button icon is not appearing after the change.

moisseev:
@jeromelebleu Thanks for the update! I have concerns regarding the changes:

- Our project would benefit from having crisp, clear icon rendering. Is it possible to retain the `SVG + JS` approach for icons instead of switching entirely to `WebFonts`? Alternatively, we could consider removing the icon from the "reload" button if that's a viable option.

- I also noticed that the loading indicator icon for FooTable doesn't seem to display correctly during tables initialization. Could you please take a look into this issue?

Thanks again for your work on this!

jeromelebleu:
You're welcome! I thought it would be an easy task to remove Glyphicons as it seemed almost done… before discovering FooTable! I am now missing a broader vision of the concerns, goals and plans of the Rspamd's Web interface to achieve that - if it is really a goal - with your concerns included.

The bigger question I have is regarding the future of the old and sometimes unmaintained libraries - such as FooTable. To avoid a double work, we could start by replace FooTable instead if having a crisp icon rendering is more important - as I think having both Font Awesome methods is non-sense. It should then be really easy to remove Glyphicons. But we could also switch to the Web fonts method as suggested - which [has other advantages](https://docs.fontawesome.com/web/dig-deeper/webfont-vs-svg) - to get rid of Glyphicons and revert to the JS+SVG method later on.

Tell me if you see another approach and which one is the most relevant! I can continue to work on in it but I just need to know how.

moisseev:
I agree that the `Web fonts` method has its advantages, but IMO the `SVG + JS` approach outweighs them. I followed your link, and the "Side-by-side comparison" table made it even stronger. From a technical point of view, vector images and icons offer more possibilities and are very convenient in my opinion. From a user experience perspective, having vector images and icons is great as they look sharp at any scale, on any device with any screen resolution or rendering method. I don't think we should sacrifice user experience just to remove a couple of files.

Personally, I lean toward preserving the `SVG + JS` approach, although my opinion might be biased since fonts designed for antialiasing are hurting my eyes and working with such fonts is a nightmare for me.

I agree that replacing FooTable with a more modern and maintained solution would be the best long-term path. Do you have any suitable replacement in mind? In that case, we could remove Glyphicons entirely _after_ migrating from FooTable without sacrificing UI clarity. In my opinion, keeping Glyphicons for now just for FooTable is not a problem. This way, we ensure visual consistency and quality while moving forward. What are your thoughts on this approach?

PR #5378, opened 2025-03-08
[Minor] Fix and improve format string processing in lua_logger_log_format()

amulet1:
Fix and improve format string processing in lua_logger_log_format()
* Fix handling of %s after % with numeric argument (`num_arg` in the original code was not reset to `false`)
(this restores the functionality described in the [docs](https://rspamd.com/doc/lua/rspamd_logger.html))
* Fix argument number check
* Preserve unhandled % in the output
* Redesign to simplify the code

| Example (e.g. using `rspamd_logger.infox`) | Original code output | New code output |
|- | - | - |
|infox("Test: arg1=%1 arg2=%2 arg3=%s", "A", 123, "3rd") |`error` wrong argument number: 0 | Test: arg1=A arg2=123 arg3=3rd |
|infox("Test is %s% done", 50) |Test is 50 done | Test is 50% done |
|infox("Test: %s extra %s", "with") |Test: with extra no value | `error` wrong argument number: 2 |

See #5322 also.

### DIFF
--- a/src/lua/lua_logger.c
+++ b/src/lua/lua_logger.c
@@ -733,123 +733,69 @@ lua_logger_log_format(lua_State *L, int fmt_pos, gboolean is_string,
 {
 	char *d;
 	const char *s, *c;
-	gsize r, cpylen = 0;
-	unsigned int arg_num = 0, cur_arg;
-	bool num_arg = false;
+	gsize r;
+	unsigned int arg_num, arg_max, cur_arg;
 	struct lua_logger_trace tr;
-	enum {
-		copy_char = 0,
-		got_percent,
-		parse_arg_num
-	} state = copy_char;
+	int digit;
 
-	d = logbuf;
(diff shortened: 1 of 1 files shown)


src/lua/lua_logger.c
```diff
@@ -733,124 +733,70 @@ lua_logger_log_format(lua_State *L, int fmt_pos, gboolean is_string,
...
-			/* We have non numeric argument, e.g. %s */
-			arg_num = cur_arg;
-		}
-
-		if (arg_num < 1 || arg_num > (unsigned int) lua_gettop(L) + 1) {
-			msg_err("wrong argument number: %ud", arg_num);
-
-			return FALSE;
+			/* Copy % */
+			--s;
```
vstakhov, line 791: 👍

vstakhov (approved):
It looks much clearer now, thank you, much appreciated!

PR #5383, opened 2025-03-11
[Minor] Lua logging improvements

amulet1:
The code changes aim to provide consistent (and corrected) buffer overflow checks and a slightly cleaner code.
* Simplify `lua_logger_out_str()`, fix length check when doing hex output
* Add `lua_logger_out()` with embedded `lua_logger_trace struct`
* Replace high level `lua_logger_out_type()` calls with `lua_logger_out()`
* Remove unused `lua_logger_trace` struct references from parameters
* Include string terminator into _buffer length checks_, adjust affected places accordingly

For example, the last 2 lines of the below code now produce the identical output (both truncated after 8063 characters, i.e. 8192-128+1) .
```
local rspamd_logger = require("rspamd_logger")
local str = "Test: 789-"
local s
for i=10, 8200, 10 do
  s = i .. ""
  str = str .. s .. string.rep(".", 10 - s:len())
end

rspamd_logger.infox(str)
rspamd_logger.infox("%s", str)

```


vstakhov:
It seems our functional tests are crashing with this PR...

amulet1:
Yes, and so far I do not see why.

vstakhov:
I see some core dump there. I will check later

amulet1:
So far I found that if I add `g_assert` in `print.c` below, it crashes `rspamd` on Learn test defined in in `090.peruser.robot`  (i.e. when `rspamc` connects).

`max<=0` likely causes data corruption. Still looking into why/where it could happen. I can't seem to reproduce it (I have FreeBSD).

```lua
glong rspamd_snprintf(char *buf, glong max, const char *fmt, ...)
{
	char *r;
	va_list args;

g_assert(max > 0);
 
	va_start(args, fmt);
 
	r = rspamd_vsnprintf(buf, max, fmt, args);
 	va_end(args);
  
	return (r - buf);
}
```

amulet1:
I found the specific code change that caused the issue, will provide more details later.

amulet1:
Calling `rspamd_snprintf()` with buffer size less than 1 (maybe even 2) is unsafe.

If the buffer size is 1, then 'rspamd_printf' calls `rspamd_vsnprintf` with remaining buffer size 0, which is at best useless. Supposedly this should not be an issue, but I suspect it somehow causes errors.

If the buffer size is 0, then `rspamd_vsnprintf` is called with remaining buffer size set to a large positive number `(glong) -1`, which could cause buffer overrun and memory corruption (plus it saves terminating 0 after the buffer end in this case).

I could not pinpoint exact place(s) [probably in `rspamd_vprintf_common`?] that do(es) this. In fact I could even reproduce this under FreeBSD. In any case, with the adjusted `MOVE_BUF` macro `lua_logger_out_table()` would not trigger this issue.

One possible remaining concern is that `lua_logger_out_userdata()` does not always return a zero-terminated string (this is _not_ introduced by my changes). This means that `lua_logger_out[_type]` is also affected by this.

amulet1:
With the latest changes the output of this:
```lua
local t = { aa = 1, bb = 2 }
t = { [t] = { key = "value" } }
rspamd_logger.infox("Test: %s", t)
```
is this (note, table index is itself a table):
`Test: {[{[aa] = 1, [bb] = 2}] = {[key] = value}}`
it is now possible to do this:

amulet1:
Still need to fix something minor.

amulet1:
@vstakhov , please review or merge.

vstakhov:
> Calling `rspamd_snprintf()` with buffer size less than 1 (maybe even =1) is unsafe.

Yes, that's true, it was intended to be called with static buffer sizes only.

vstakhov (approved):

Issue #5382, opened 2025-03-10
[BUG] GPT Module:  reason_header is not working as aspected


alex081512:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
The config setting 'reason_header' is not implemented correctly. It's only added for GPT_HAM, It should also added to GPT_SPAM, like printed in the debug log.

> `gpt.lua:541: model: gpt-4o-mini; spam: 0.85; reason: 'The use of a suspicious email address and promotional language raises significant concerns about legitimacy.'`

Also It's only adding a header with the string 'value'.

> `X-GPT-Reason: value`

### Steps to Reproduce

1. Enable the gpt module
2. Set `reason_header = "X-GPT-Reason";`

**Expected behavior**
Always add the header with the reported reason.

### Versions

Rspamd daemon version 3.11.1
CPU architecture x86_64

alex081512:
Fixed with #5387

PR #5387, opened 2025-03-13
[Minor] gpt.lua improvements and one fix for reason_header

alex081512:
One bugfix for reason_header (see #5382) and some improvements that I think will be useful for everyone.

- Gpt prompt - Only ask for red flag when it's spam
- Insert Reason to result to show gpt answer in history

### DIFF
--- a/src/plugins/lua/gpt.lua
+++ b/src/plugins/lua/gpt.lua
@@ -496,7 +496,7 @@ local function insert_results(task, result, sel_part)
   end
 
   if result.probability > 0.5 then
-    task:insert_result('GPT_SPAM', (result.probability - 0.5) * 2, tostring(result.probability * 100))
+    task:insert_result('GPT_SPAM', (result.probability - 0.5) * 2, tostring(result.probability))
     if settings.autolearn then
       task:set_flag("learn_spam")
     end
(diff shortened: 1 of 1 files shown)


src/plugins/lua/gpt.lua
```diff
@@ -494,8 +494,16 @@ local function insert_results(task, result, sel_part)
...
+  local resultText
+  if result.reason then
+    resultText = tostring(result.probability * 100) .. '% - ' .. result.reason
+  else
+    resultText = tostring(result.probability * 100)
+  end
+
   if result.probability > 0.5 then
-    task:insert_result('GPT_SPAM', (result.probability - 0.5) * 2, tostring(result.probability))
+    task:insert_result('GPT_SPAM', (result.probability - 0.5) * 2, resultText)
```
vstakhov, line 506: It will massively pollute output, please don't do it. Symbols options must be short and compact as they are stored everywhere.

vstakhov (changes requested):
It looks good but symbol insertion should be reverted back to probability only.

vstakhov:
> Insert Reason to result to show gpt answer in history

No, it shouldn't be done in this way:
* Redis history is a limited module and it should not be used to store any analytics
* Clickhouse/ES module are more appropriate but they should use some custom selector to store LLM reply

Symbols' options are not intended to store any large portions of data. Same is applicable for history_redis.

src/plugins/lua/gpt.lua
alex081512: Okay, thank you for your feedback. That makes sense to me. I have reverted the changes.

alex081512 (commented):
Changes reviewed and implemented.

vstakhov (approved):
LGTM, thank you!

PR #5396, opened 2025-03-17
Add rspamadm ready command to check controller readiness

DarshanCode2005:
This PR adds a new command, `rspamadm ready`, which polls the Rspamd controller's `/ready` endpoint until it returns a successful response or times out. This addresses the need for a reliable way to determine when Rspamd is fully operational, which is particularly useful in automated environments, container orchestration, and init scripts.

## Features

- Polls the `/ready` endpoint with configurable intervals and timeout
- Returns exit code 0 when Rspamd is ready, 1 otherwise (for easy scripting)
- Configurable options for URL, timeout, polling interval, and verbosity
- Follows the established pattern of other rspamadm commands

## Usage Examples

Basic usage:
```
rspamadm ready
```

With custom URL and timeout:
```
rspamadm ready --url https://rspamd.example.com:11334 --timeout 120
```

For scripts/automation (quiet mode):
```
rspamadm ready --quiet && echo "Rspamd is ready, proceeding with next steps"
```

## Implementation Details

The command is implemented in `lualib/rspamadm/ready.lua` following the pattern of existing commands like `clickhouse.lua`. It uses the existing HTTP functionality in Rspamd to make requests to the controller's `/ready` endpoint.

This implementation directly addresses the issue of needing to wait for Rspamd to be fully operational before proceeding with dependent tasks, which is particularly important in containerized environments and automated deployment scenarios.

### DIFF
--- a/lualib/rspamadm/ready.lua
+++ b/lualib/rspamadm/ready.lua
@@ -1,0 +1,164 @@
+--[[
+Copyright (c) 2022, Vsevolod Stakhov <vsevolod@rspamd.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+]]--
+
+local argparse = require "argparse"
+local rspamd_http = require "rspamd_http"
+local rspamd_logger = require "rspamd_logger"
+local rspamd_upstream_list = require "rspamd_upstream_list"
+local lua_util = require "lua_util"
+
+local E = {}
+
+-- Define command line options
+local parser = argparse()
+    :name 'rspamadm ready'
+    :description 'Check if Rspamd controller is ready'
+    :help_description_margin(30)
+    :command_target('command')
+
+parser:option '-c --config'
+      :description 'Path to config file'
+      :argname('config_file')
+      :default(rspamd_paths['CONFDIR'] .. '/rspamd.conf')
+parser:option '-u --url'
+      :description 'URL of the Rspamd controller'
+      :argname('url')
+      :default('http://localhost:11334')
+parser:option '-t --timeout'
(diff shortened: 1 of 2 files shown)


lualib/rspamadm/ready.lua
```diff
@@ -0,0 +1,164 @@
...
+end
+
+local function poll_ready(args)
+  local total_timeout = tonumber(args.timeout)
+  local interval = tonumber(args.interval)
+  local url = args.url
+  
+  -- Add /ready endpoint if not specified
+  if not url:match("/ready$") then
+    url = lua_util.remove_trailing_slash(url) .. "/ready"
```
fatalbanana, line 94: The method doesn't exist

lualib/rspamadm/ready.lua
```diff
@@ -0,0 +1,164 @@
...
+        io.stderr:write("Timeout reached. Rspamd is not ready.\n")
+      end
+      break
+    end
+    
+    if not args.quiet and args.verbose then
+      io.stdout:write(string.format("Waiting %d seconds before next attempt...\n", interval))
+    end
+    
+    lua_util.sleep(interval)
```
fatalbanana, line 146: Method doesn't exist; we might need to reach for FFI here

lualib/rspamadm/ready.lua
```diff
@@ -0,0 +1,164 @@
...
+      :argname('verbose')
+
+local http_params = {
+  config = rspamd_config,
+  ev_base = rspamadm_ev_base,
+  session = rspamadm_session,
+  resolver = rspamadm_dns_resolver,
+}
+
+local function load_config(config_file)
```
fatalbanana, line 65: We don't need to load config I suppose. It could be used for finding controller address but I think having it reconfigurable & giving it a sensible default is fine.

src/rspamadm/configdump.c
```diff
@@ -38,6 +38,10 @@ extern worker_t *workers[];
 
 static void rspamadm_configdump(int argc, char **argv, const struct rspamadm_command *);
 static const char *rspamadm_configdump_help(gboolean full_help, const struct rspamadm_command *);
+/*Add local_only option to GOptionEntry entries array */
```
fatalbanana, line 41: Better push these changes to a different branch please

lualib/rspamadm/ready.lua
vstakhov: We can use `rspamadm_ev_base` to add async timers.

PR #5398, opened 2025-03-17
Fix Async Polling and URL Formatting in rspamadm ready Command

DarshanCode2005:
#### **Fix: Use `rspamadm_ev_base:add_timer` Instead of `lua_util.sleep` & Remove Invalid `remove_trailing_slash`**  

This PR addresses two issues in the `rspamadm ready` command:  

1. **Fix `remove_trailing_slash` issue:**  
   - The function `lua_util.remove_trailing_slash(url)` does not exist, leading to an error when appending `/ready` to the URL.  
   - **Fix:** Instead of relying on this non-existent function, the fix ensures proper URL formatting using `url:gsub("/$", "")`. This removes any trailing slashes safely and appends `/ready` only when needed.  

2. **Fix `lua_util.sleep(interval)` issue:**  
   - `lua_util.sleep(interval)` does not exist, causing an error when waiting between retries.  
   - **Fix:** Instead of using blocking sleep, the fix introduces `rspamadm_ev_base:add_timer(interval, retry_function)`, ensuring non-blocking, event-driven polling while waiting for Rspamd to be ready.  

#### **Impact of the Fixes:**  
1. Eliminates errors caused by missing functions.  
2. Improves robustness by correctly formatting the URL.  
3. Enhances efficiency by using an async timer instead of blocking execution. 

Thankyou! Let me know what further refinements can be made, Sir 🫡.

### DIFF
--- a/lualib/rspamadm/ready.lua
+++ b/lualib/rspamadm/ready.lua
@@ -17,10 +17,6 @@ limitations under the License.
 local argparse = require "argparse"
 local rspamd_http = require "rspamd_http"
 local rspamd_logger = require "rspamd_logger"
-local rspamd_upstream_list = require "rspamd_upstream_list"
-local lua_util = require "lua_util"
-
-local E = {}
 
 -- Define command line options
 local parser = argparse()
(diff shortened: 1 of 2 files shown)


fatalbanana:
Retries are not executed because the event loop is not running - please test your work.

fatalbanana (changes requested):
Please see that you test it. :)

PR #5404, opened 2025-03-23
Trying to add command for polling /ready

NarutoUchiha39:
I have added `ready.lua` in `lualib/rspamadm` folder. The endpoint has the following features:

1. It accepts an `interval` argument from the command line to specify the intervals in which the `/ready` should be polled 
2. It accepts a `timeout` argument from the command line to specify the time interval in which the polling should stop
3. It accepts a `-v` flag to print output to terminal
4. It accepts a `url` argument to specify a custom url to be polled
5. It accepts a `--ssl-verify` flag to enable ssl in the request made

I have made use of `rspamd_http` to make the http request and I have used the `rspamd_config:add_periodic()` function to add register the timer for the asynchronous request. `rspamadm_ev_base:loop()` has been used to keep the script running

Any suggestions or changes would be greatly appreciated

### DIFF
--- a/lualib/rspamadm/ready.lua
+++ b/lualib/rspamadm/ready.lua
@@ -36,7 +36,7 @@ parser:flag '-v --verbose'
 
 local function log_and_print(level, verbose,fmt,...)
     rspamd_logger[level](fmt, ...)
-    if verbose and level ~="errx" then
+    if verbose and (level ~="errx" or level~="warnx")  then
         print(string.format(fmt, ...))
     end
 end
@@ -75,6 +75,8 @@ local function try_again(url, ssl_verify, verbose)
 end
 
 local function check_ready_interval(interval, timeout, url, ssl_verify,verbose)
+
+    try_again(url, ssl_verify,verbose)
     local start_time = os.time()
 
     rspamd_config:add_periodic(rspamadm_ev_base, interval, function(cfg, ev_base)
--- a/lualib/rspamadm/ready.lua
+++ b/lualib/rspamadm/ready.lua
@@ -1,0 +1,111 @@
+local argparse = require "argparse"
+local rspamd_http = require "rspamd_http"
+local rspamd_logger = require "rspamd_logger"
+
+local parser = argparse()
+    :name 'rspamadm ready'
+    :description 'Check whether Rspamd is ready'
+    :help_description_margin(30)
+    :command_target('command')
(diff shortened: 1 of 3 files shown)


fatalbanana:
It looks like you've pushed the wrong commit.

NarutoUchiha39:
I fixed the issue. I am really sorry for the mistake

lualib/rspamadm/ready.lua
```diff
@@ -0,0 +1,111 @@
...
+      :description 'Disable SSL verification'
+      :argname('ssl_verify')
+      :default(false)
+
+parser:flag '-v --verbose'
+      :description 'Give verbose output'
+      :argname('verbose')
+      :default(false)
+
+local function log_and_print(level, verbose,fmt,...)
```
fatalbanana, line 37: The logger is printing to stderr for `warn` level or better. Better call it directly, use `verbose` to control if you would call it at all where helpful.

lualib/rspamadm/ready.lua
```diff
@@ -0,0 +1,111 @@
...
+                end
+            end
+        end
+    })
+end
+
+local function check_ready_interval(interval, timeout, url, ssl_verify,verbose)
+    local start_time = os.time()
+
+    rspamd_config:add_periodic(rspamadm_ev_base, interval, function(cfg, ev_base)
```
fatalbanana, line 80: We wait `interval` before sending the first probe, better it should be sent right away

fatalbanana (changes requested):

lualib/rspamadm/ready.lua
NarutoUchiha39: I have adjusted the logger such that output to terminal is not printed when the log level is errx or warnx

lualib/rspamadm/ready.lua
NarutoUchiha39: I have added a function call before adding the timer so that request is made straight away

lualib/rspamadm/ready.lua
NarutoUchiha39: I wanted to clarify I should call logging levels of warn or higher only when verbose is set? I was asking because I thought if `errx` or 'warnx` is called only when verbose is set, then some important logs can be missed if `verbose` is not set to true.

PR #5405, opened 2025-03-24
Add HEADER_FROM_CONTAINS_URL

dragoangel:
This symbol will detect if header from contains any URLs and if yes - will add negative score to email. Usually such cases has place when bots spam "contact us" forms on websites and website use "From" as visible part of from in email. Rspamd do not verify URLs in From header, so we can restrict them via this rule as is.

### DIFF
--- a/rules/headers_checks.lua
+++ b/rules/headers_checks.lua
@@ -552,6 +552,23 @@ rspamd_config.HEADER_FORGED_MDN = {
   description = 'Read confirmation address is different to return path'
 }
 
+rspamd_config.HEADER_FROM_CONTAINS_URL = {
+  callback = function(task)
+    local from = task:get_from('mime')
+    if from and from[1] and from[1].name then
+      local display_name = from[1].name
+      if display_name:match(' https?://') then
+        return true
+      end
+    end
+    return false
+  end,
+
+  score = 5.0,
+  group = 'headers',
+  description = 'Name in From header contains URL'
+}
+
 local headers_unique = {
   ['Content-Type'] = 1.0,
   ['Content-Transfer-Encoding'] = 1.0,


fatalbanana:
> @vstakhov @moisseev Hi, can you please review, is this change is okay for you?

It could be a job for regex I suppose; if we feel a need to make very sure it's in `name` portion of the `From` header we might reach for selectors.

dragoangel:
Hi @fatalbanana yes, selectors can work and regex too, question is - where then use this selector?

fatalbanana:
> Hi @fatalbanana yes, selectors can work and regex too, question is - where then use this selector?

They can be used in regex rules, like [here](https://github.com/rspamd/rspamd/blob/master/test/functional/lua/selector_test.lua).

dragoangel:
Tryining to understand what is the benefits?

> > Hi @fatalbanana yes, selectors can work and regex too, question is - where then use this selector?
> 
> They can be used in regex rules, like [here](https://github.com/rspamd/rspamd/blob/master/test/functional/lua/selector_test.lua).

dragoangel:
> > @vstakhov @moisseev Hi, can you please review, is this change is okay for you?
> 
> It could be a job for regex I suppose; if we feel a need to make very sure it's in `name` portion of the `From` header we might reach for selectors.

Technically it's only one possible place where you can put `http://` in `From` header, so we not need to exactly match only `name`, but we need receive header in decoded, not raw format.

fatalbanana:
`/X` is raw header, `/H` is decoded, refer [here](https://rspamd.com/doc/modules/regexp.html).

You can use `[[square brackets]]` to avoid double-escaping.

The benefit of using regexp rules over Lua is in performance.

dragoangel:
Created

PR #5406, opened 2025-03-24
Add R_HTTP_URL_IN_FROM

dragoangel:
This symbol will detect if header from contains any URLs and if yes - will add negative score to email. Usually such cases has place when bots spam "contact us" forms on websites and website use "From" as visible part of from in email. Rspamd do not verify URLs in From header, so we can restrict them via this rule as is.

Instead of https://github.com/rspamd/rspamd/pull/5405

### DIFF
--- a/rules/regexp/headers.lua
+++ b/rules/regexp/headers.lua
@@ -69,6 +69,14 @@ if rspamd_config:is_mime_utf8() then
   end
 end
 
+reconf['R_HTTP_URL_IN_FROM'] = {
+  re = [[From=/\shttps?:\/\/\S/iH]],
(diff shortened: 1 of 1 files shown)


dragoangel:
@fatalbanana please review 😊

dragoangel:
Added `(^|\s)` in case `http://` is a first thing in header.

dragoangel:
Changed regex to `/(^|"|'|\s)[hH][tT][tT][pP][sS]?(:|=3A)\/\/\S/H` so it will catch stuff like ` https=3A//example.com` too as in some broken cases spammers even encode `:` with `=3A` in header From.

PR #5414, opened 2025-03-30
[Fix] Prevent config failure with defined action as a number

### LINKED ISSUE #5384: [BUG] 'rspamadm configtest' throws Lua error
### Prerequisites

* [ x] Put an X between the brackets on this line if you have done all of the following:

**Describe the bug**
'rspamadm configtest' throws the following error:

> call to rspamadm lua script failed (2): /usr/share/rspamd/lualib/lua_cfg_transform.lua:378: attempt to compare number with userdata; trace: [1]:{/usr/share/rspamd/lualib/lua_cfg_transform.lua:378 - <unknown> [Lua]};

### Steps to Reproduce

1. Use the following 'local.d/actions.conf':
```
rewrite_subject = 6.0;
subject = "*****SPAM***** %s"
add_header = null;
reject = null;
greylist = null;
```
2. Call 'rspamadm configtest'

**Expected behavior**
Output:

> syntax OK

### Versions
Rspamd daemon version 3.11.1
CPU architecture x86_64
OS Debian 12

Optimus-NP:
[Fix #5384] Prevent config failure with defined action as a number  

### Problem  
The configuration fails when a **defined action** is set as a **numeric value**  
instead of a **string**, causing execution errors.  

#### Root Cause  
- If a defined action is assigned a **string** (`"6.0"`), it works correctly.  
- However, if it is assigned a **numeric value** (`6.0`), the code attempts to  
  `unwrap()` it, leading to failure.  

#### Solution  
- Check the type of the defined action before processing.  
- Handle both **string** and **numeric** values correctly without using `unwrap()`.  

#### Config File  
The following configuration was used to verify the fix:  

  ```lua
  rewrite_subject = "6.0";
  subject = "*****SPAM***** %s"
  add_header = false;
  reject = 7.0;
  greylist = null;
```

#### Command

```
rspamadm configtest
```

### DIFF
--- a/lualib/lua_cfg_transform.lua
+++ b/lualib/lua_cfg_transform.lua
@@ -367,8 +367,18 @@ return function(cfg)
     for i = 1, (#actions_order - 1) do
       local act = actions_order[i]
 
-      if actions:at(act) and actions:at(act):type() ~= 'object' then
-        local score = actions:at(act):unwrap()
+      local act_value = actions:at(act)
+
+      if act_value then
+        local val_type = act_value:type()
+        local score = 0
+        if val_type == 'string' then
+          if act_value:unwrap() ~= 'null' then
+            score = tonumber(act_value:unwrap())
+          end
+        elseif val_type == 'number' then
+          score = act_value:unwrap()
+        end
 
         for j = i + 1, #actions_order do
           local next_act = actions_order[j]

### CODE REVIEW

**Comment by Optimus-NP** [comment]
@vstakhov

Thanks for reviewing this! I've added a video demonstrating the issue reproduction along with the fix in action. The issue arises because the configuration fails when a defined action is set as a numeric value instead of a string, leading to execution errors. This change resolves the issue by handling numeric values and strings separately. Let me know if you need any further details!


https://github.com/user-attachments/assets/3be2431b-0c37-4380-9471-34490dbc33a3

**Comment by dragoangel** [comment]
> @vstakhov
> 
> Thanks for reviewing this! I've added a video demonstrating the issue reproduction along with the fix in action. The issue arises because the configuration fails when a defined action is set as a numeric value instead of a string, leading to execution errors. This change resolves the issue by handling numeric values and strings separately. Let me know if you need any further details!
> 
>  rspamd.mp4

also can't reproduce your tests. everything works without any "fixes":
```
root@81a9f639d91a:/# cat /etc/rspamd/local.d/actions.conf
add_header = {
  score = 8.0,
};

reject = {
  score = 15.0,
};

discard = {
  flags = ["no_threshold"],
};

subject = "***SPAM*** %s";
rewrite_subject = 7.0;

# to have quarantine action working smtp_message MUST be nil
# task:set_pre_result('quarantine', nil, module_name)
quarantine = {
  flags = ["no_threshold"],
  milter = "quarantine",
};

greylist = {
  flags = ["no_threshold"],
};

root@81a9f639d91a:/# rspamadm configtest
syntax OK
```

Issue #5370, opened 2025-03-03
[BUG] rspamd 3.11 segfaults on RHEL 9


johnmosli:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

rspamd segfaults, sadly I can't reproduce with a specific mail message.

### Steps to Reproduce

_service rspamd start_

**Expected behavior**

Service works as expected

### Versions
* Red Hat Enterprise Linux release 9.5 (Plow)
* rspamd-3.11.0-2.el9.x86_64 (rspamd-debuginfo-3.11.0-2.el9.x86_64)

### Additional Information
_# coredumpctl gdb 549967_

```
           PID: 549967 (rspamd)
           UID: 600 (_rspamd)
           GID: 599 (_rspamd)
        Signal: 11 (SEGV)
     Timestamp: Mon 2025-03-03 11:34:43 CET (17min ago)
  Command Line: $'rspamd: rspamd_proxy process (*:11332)'
    Executable: /usr/bin/rspamd
 Control Group: /system.slice/rspamd.service
          Unit: rspamd.service
         Slice: system.slice
       Boot ID: bb778379b67a4e3c9f12c66e7d30fc33
    Machine ID: b323518691d04191aa98d7a93311c96c
       Storage: /var/lib/systemd/coredump/core.rspamd.600.bb778379b67a4e3c9f12c66e7d30fc33.549967.1740998083000000.zst (present)
  Size on Disk: 102.9M
       Message: Process 549967 (rspamd) of user 600 dumped core.

                Stack trace of thread 549967:
                #0  0x00007f939f80ed0a free (libjemalloc.so.2 + 0xed0a)
[...]

PR #5415, opened 2025-03-31
[Fix]Solve #5370

Subhiiiiii:
# Bug Report

## Description
Segmentation faults occur when processing large messages (>50-60MB) due to null pointer dereferences and out-of-bounds array access in the dynamic items cache handling.

## Steps to Reproduce
1. Configure Rspamd 3.11 with default settings
2. Process a large email message (>50MB)
3. Observe segmentation fault in worker process

Alternatively:
1. Stress test Rspamd with many concurrent large messages
2. Segmentation faults will occur intermittently

## Expected Behavior
Rspamd should process large messages without crashing, gracefully handling memory constraints.

## Actual Behavior
Rspamd worker processes crash with segmentation faults when processing large messages.

## Environment
- Rspamd version: 3.11
- OS: [e.g. Linux distribution and version]
- Hardware: [if relevant]

## Additional Information
The issue occurs in the dynamic items cache handling:
1. `process_pre_postfilters` attempts to access `dyn_item->status` without null check
2. `get_dynamic_item` accesses `dynamic_items` array without bounds checking

## Proposed Fix
The issue is fixed by:
1. Adding null pointer checks before accessing dynamic items
2. Adding bounds checking for array accesses

```cpp
// In process_pre_postfilters:
if (dyn_item && dyn_item->status == cache_item_status::not_started) {

// In get_dynamic_item:
if (our_id_maybe && our_id_maybe.value() < dynamic_items.size()) {
    return &dynamic_items[our_id_maybe.value()];
}

### DIFF
--- a/src/libserver/symcache/symcache_runtime.cxx
+++ b/src/libserver/symcache/symcache_runtime.cxx
@@ -279,11 +279,10 @@ auto symcache_runtime::is_symbol_enabled(struct rspamd_task *task, const symcach
 
 auto symcache_runtime::get_dynamic_item(int id) const -> cache_dynamic_item *
 {
-
 	/* Not found in the cache, do a hash lookup */
 	auto our_id_maybe = rspamd::find_map(order->by_cache_id, id);
 
-	if (our_id_maybe) {
+	if (our_id_maybe && our_id_maybe.value() < dynamic_items.size()) {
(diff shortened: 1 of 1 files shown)


vstakhov:
No, that is not a proper sollution, as it has nothing related to the underlying issue.

Issue #5410, opened 2025-03-27
[BUG] Spammers can easily bypass greylist by sending the same content to multiple recipients.


mbrother2:
### Prerequisites

* [ X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
As title spammers can easily bypass greylist by sending the same content to multiple recipients.
If the `timeout` is set to 5 minutes by default (I think should be the case so that valid emails can be accepted). After 5 minutes, all spam emails that was previously greylisted will become valid by it's body.

Rspamd log: `GREYLIST(0.00){pass;body;}`

**Expected behavior**
Greylist module should combine from:to:ip:data into a single hash for checking, so that spammers will always be greylisted if they send to multiple recipients.

### Versions

Rspamd latest

vstakhov:
Definitely not IP, that will break the whole idea. But `rcpt` sounds like a good idea indeed.

mbrother2:
Hi @vstakhov,

As rspamd's document:
```
When the Greylisting module is enabled, two hashes are saved for each message in Redis:

meta hash is based on triplet from:to:ip
data hash is taken from the message’s body if it has enough length for it
```

So I think should greylist module combine from:to:ip:data into a single hash ?

vstakhov:
No, it's not true. Body hash was designed to prevent endless greylisting if a sender uses multiple ip ranges.

mbrother2:
Hello,

I don't think a legitimate user sends 2 identical emails multiple times. So i think hashing the body is something that only helps spammers. Only spammers send mass emails with the same content to many people. Even i think we should just hash triplet from:to:ip to bypass greylist or have new option to disable bypass greylist for body bash!

moisseev:
[...]

PR #5416, opened 2025-03-31
[Fix]Solve #5410

Subhiiiiii:
# Bug Report: Greylisting Bypass via Multiple Recipients

## Description
Spammers could bypass the greylisting module by sending identical content to multiple recipients. The issue occurred because the greylist hash didn't incorporate recipient information, allowing spammers to reuse the same content across different recipients.

## Steps to Reproduce
1. Configure Rspamd with greylisting enabled (default settings)
2. Send identical email content to multiple recipients (recipient1@domain, recipient2@domain)
3. Observe that subsequent identical messages to different recipients bypass greylisting

## Expected Behavior
Each unique combination of message content and recipient(s) should be treated as a distinct greylist entry, requiring separate greylisting periods.

## Actual Behavior
Identical message content would bypass greylisting when sent to different recipients, as only the message body was considered in the greylist hash.

## Environment
- Rspamd version: Affects versions before the fix
- Redis: Required for greylisting persistence

## Root Cause
1. The `data_key()` function didn't include recipient information in the hash calculation
2. The greylist check logic treated body and metadata hashes separately, creating potential bypass vectors

## Solution Implemented
1. Modified `data_key()` to include sorted recipient addresses in the hash:
```lua
local rcpt = task:get_recipients('smtp')
if rcpt then
  table.sort(rcpt, function(r1, r2) return r1['addr'] < r2['addr'] end)
  fun.each(function(r) h:update(r['addr']) end, rcpt)
end

2. Enhanced `greylist_check()` to:
- Check both body and metadata hashes together
- Apply consistent greylisting logic
- Improve error handling:
```lua
if data[1] and type(data[1]) ~= 'userdata' and data[2] and type(data[2]) ~= 'userdata' then
  -- Combined check logic
else
  -- Treat as greylisted if either hash is missing
  task:get_mempool():set_variable("grey_greylisted", 'true')
end
```

## Impact
- Improved security against spam campaigns targeting multiple recipients
- Maintained backward compatibility with existing configurations
- No impact on legitimate email flow
- Slightly increased Redis storage requirements due to more unique hashes

### DIFF
--- a/src/libserver/symcache/symcache_runtime.cxx
+++ b/src/libserver/symcache/symcache_runtime.cxx
@@ -279,11 +279,10 @@ auto symcache_runtime::is_symbol_enabled(struct rspamd_task *task, const symcach
 
 auto symcache_runtime::get_dynamic_item(int id) const -> cache_dynamic_item *
 {
-
 	/* Not found in the cache, do a hash lookup */
 	auto our_id_maybe = rspamd::find_map(order->by_cache_id, id);
 
-	if (our_id_maybe) {
+	if (our_id_maybe && our_id_maybe.value() < dynamic_items.size()) {
 		return &dynamic_items[our_id_maybe.value()];
 	}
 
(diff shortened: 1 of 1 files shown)


fatalbanana:
Same changes as in #5415 - that's not likely intended.

Issue #5418, opened 2025-04-01
[Feature] [Apache Kvrocks compatibility] EVALSHA_RO command adoption


quantranhong1999:
## Summary

Recently, we have been migrating Redis to Apache Kvrocks as the Rspamd bayes training storage, as we need to train a great amount of user mails and Redis has memory limitations.

However, the migration does not work straightforwardly. It seems that Rspamd relies on `EVALSHA` a read-only Lua script to get its BAYES statistics:

```lua
local cursor = tonumber(KEYS[1])
local ret = redis.call('SSCAN', KEYS[2] .. '_keys', cursor, 'COUNT', tonumber(KEYS[4]))
local new_cursor = tonumber(ret[1])
local nkeys = #ret[2]
local learns = 0

for _, key in ipairs(ret[2]) do
    learns = learns + (tonumber(redis.call('HGET', key, KEYS[3])) or 0)
end

return { new_cursor, nkeys, learns }
```

While Redis replicas accept `EVALSHA` for the readonly Lua script, Kvrocks replicas seem to be more strict and reject Rspamd: `lua_bayes_redis.lua:145: cannot get bayes statistics for BAYES_HAM: READONLY You can't write against a read only slave`.

## Motivation

While we have reported the EVALSHA issue for the read-only Lua script to the Apache Kvrocks community and received supportive feedback, we think it would be good too if Rspamd could adopt the [`EVALSHA_RO`](https://redis.io/docs/latest/commands/evalsha_ro/) usage so the Rspamd <-> Apache Kvrocks integration can be smoother.

## Describe alternatives you've considered

For foreseen read-only Lua scripts, Rspamd uses `EVALSHA_RO` to execute them. 

## Additional context

Rspamd 3.11.1.
Apache Kvrocks 2.11.1 sentinel mode.

Thank you in advance for considering my request.

dragoangel:
Can it be closed as it resolved afaik?

PR #5419, opened 2025-04-02
[Feature] Allow to specify Redis version

vstakhov:
Some Redis versions introduce important changes to the commands, for
example `EVAL` vs `EVAL_RO`. Rspamd should be able to use these commands
where possible.

Issue: #5418


quantranhong1999 (commented):
Great job. Thank you <3.

PR #5452, opened 2025-05-01
Added --non-default Option to rspamadm configdump for Local Configuration

### LINKED ISSUE #5394: [Feature] Support dumping of the local configuration
## Summary

`rspamadm configdump` should be able to return only local configuration. Basically it means traversing `ucl_object_t` finding paths with children that have priority > 0.

kevinzb56:
Fixes #5394 

- Added -n/--non-default option to dump local configs (priority > 0).
- Implemented filter_non_default function recursively to traverse ucl_object_t, retaining objects with priority > 0 or their parent nodes.
- Updated configuration output logic in rspamadm_configdump to apply filter_non_default when --non-default is specified, with proper memory management.
- Modified help text in rspamadm_configdump_help to document the new -n option.

### DIFF
--- a/src/rspamadm/configdump.c
+++ b/src/rspamadm/configdump.c
@@ -31,6 +31,7 @@ static gboolean symbol_groups_only = FALSE;
 static gboolean symbol_full_details = FALSE;
 static gboolean skip_template = FALSE;
 static char *config = NULL;
+static gboolean non_default = FALSE;
 extern struct rspamd_main *rspamd_main;
 /* Defined in modules.c */
 extern module_t *modules[];
@@ -66,6 +67,7 @@ static GOptionEntry entries[] = {
 	 "Show full symbol details only", NULL},
 	{"skip-template", 'T', 0, G_OPTION_ARG_NONE, &skip_template,
 	 "Do not apply Jinja templates", NULL},
+	{"non-default", 'n', 0, G_OPTION_ARG_NONE, &non_default, "Show only non-default configuration", NULL},
 	{NULL, 0, 0, G_OPTION_ARG_NONE, NULL, NULL, NULL}};
 
 static const char *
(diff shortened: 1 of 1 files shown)


kevinzb56:
@vstakhov made the required changes

vstakhov:
Btw, your code is clearly wrong: it emits an important compile warning and after run it dumps a core file. Have you even tested your changes?

src/rspamadm/configdump.c
```diff
@@ -96,6 +99,54 @@ config_logger(rspamd_mempool_t *pool, gpointer ud)
...
+
+	if (ucl_object_type(obj) == UCL_OBJECT || ucl_object_type(obj) == UCL_ARRAY) {
+		result = ucl_object_typed_new(ucl_object_type(obj));
+		while ((cur = ucl_object_iterate(obj, &it, true))) {
+			ucl_object_t *filtered = filter_non_default(cur);
+			if (filtered) {
+				ucl_object_insert_key(result, filtered, ucl_object_key(cur), cur->keylen, true);
+			}
+		}
+		if (ucl_object_iterate_reset(result, &it) == NULL) {
```
vstakhov, line 140: What is that?

vstakhov (dismissed):

vstakhov:
I think I have fixed the issues above. I also think we need to rename option, as I personally don't like `--non-default` - it should be `--local` and `--override` instead. Let me finish that...

dragoangel:
From my local tests I got strange results, it not works as expected, f.e. it basically fail to render configuration in actions or other stuff, just add some, you will quickly find what I mean: integers get 0, lists become empty and so on...

vstakhov:
> From my local tests I got strange results, it not works as expected

Yes, the logic of filtering was completely broken. I have fixed it I think.

src/rspamadm/configdump.c
```diff
@@ -537,17 +634,29 @@ rspamadm_configdump(int argc, char **argv, const struct rspamadm_command *cmd)
 				else {
 					LL_FOREACH(obj, cur)
 					{
+						const ucl_object_t *output_obj = cur;
+						if (local_conf_only || override_conf_only) {
+							output_obj = filter_non_default(cur, override_conf_only);
+							if (!output_obj) {
+								rspamd_printf("No non-default configuration found for section %s\n", argv[i]);
```
vstakhov, line 641: So how one will filter that from other output, oops?

vstakhov (changes requested):

dragoangel:
@vstakhov do you want me to test it again with the latest changes?

vstakhov:
Please do it if you can

PR #5458, opened 2025-05-07
[Fix] libmime: doctest 2.4.12 compatibilty

xandris:
Petr Vaněk pointed out a change in doctest 2.4.12 that broke the libmime test code:
https://github.com/doctest/doctest/commit/d7a5eeb820b61cc3af8a61b3117f76a403b7742a

CHECK is now treats LHS as const and the compiler won't find non-const operator== as a result. It's reasonable for comparators to be const, so make them const.

### DIFF
--- a/src/libmime/mime_string.hxx
+++ b/src/libmime/mime_string.hxx
@@ -497,19 +497,19 @@ public:
 	}
 
 	/* Comparison */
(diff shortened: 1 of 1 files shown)


vstakhov:
I agree, it seems I have forgotten to add `const` to those methods, thank you!

PR #5464, opened 2025-05-12
fix(lua_redis): Improve Redis script loading

jendis:
This change modifies how Rspamd handles Redis script loading across multiple servers to improve resilience during server failures and restarts.

Previously, the system required ALL Redis servers to successfully load a script before considering it ready for use. This caused modules to get stuck in a waiting state when any Redis server was down, reporting errors like: "redis script is not ready, waiting it to be loaded" even when most servers were operational.

Key changes:
- Replace is_all_servers_ready() with is_any_server_ready() to allow operation when at least one server has successfully loaded the script
- Reset all servers to "unsent" status when NOSCRIPT errors are encountered to properly handle server restarts
- Ensure script loading is retried appropriately on reconnection

This fix allows Rspamd to continue operating when some Redis servers are temporarily unavailable and to recover gracefully when servers rejoin the pool.

### DIFF
--- a/lualib/lua_redis.lua
+++ b/lualib/lua_redis.lua
@@ -1207,15 +1207,13 @@ local function prepare_redis_call(script)
   return options
 end
 
-local function is_all_servers_ready(script)
+local function is_any_server_ready(script)
   for _, s in ipairs(script.servers_ready) do
-    if s == "unsent" or s == "tempfail" then
-      return false
+    if s == "done" then
+      return true
     end
   end
-
-  -- We assume that permanent errors are not recoverable, so we will just skip those servers
-  return true
+  return false
 end
 
 local function is_all_servers_failed(script)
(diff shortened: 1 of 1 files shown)


vstakhov:
This change might be dangerous for some use cases I suppose. I'm also thinking about transferring this data to the health report system and maybe to the upstreams logic. Maybe we should postpone merging that for the next stable version to allow more time for a decision.

jendis:
Could you please mention the dangerous cases, so I can eventually propose better solution?
In our environment, we're running instances of rspamd with this patch and it definitely fixes the most critical&usual problem when any redis backend restarts (due to server updates etc.) which makes Bayes, Neural and other defunct. And that's very bad.

vstakhov:
I'm not saying there is anything wrong with this approach :) I just want to think more about it, e.g. for a sharded cluster scenario, or with some other scenarios.

PR #5494, opened 2025-06-05
[Enhancement] hfilter: Add OVH VPS generic hostname

g00g1:
This pull requests adds `*.vps.ovh.us` and `*.vps.ovh.net` hostnames that are set by default for OVH VPS. Frequently seen from spam sources.

### DIFF
--- a/src/plugins/lua/hfilter.lua
+++ b/src/plugins/lua/hfilter.lua
@@ -131,6 +131,7 @@ local checks_hellohost = [[
 /modem[.-][0-9]/i 5
 /[0-9][.-]?dhcp/i 5
 /wifi[.-][0-9]/i 5
+/\.vps\.ovh\.(us|net)$/i 5
 ]]
 local checks_hellohost_map
 


dragoangel:
I think `5` is creepy a lot. More over it's not a place for shared hosting like OVH in this module. If go that way lets list there 100500 vps providers then (sarcasm 😅)? For such purpose dedicated multimap at https://github.com/rspamd/maps/tree/master/rspamd would fit much better and ability to change scores for symbol and relatively not big score by default (`1` or `0.5`)

g00g1:
You might be right. There is `/[.-]clients?[.-]/i 1` expression that matches Hetzners' default PTR like `static.0.0.21.65.clients.your-server.de.`, hence score of `1` should fit much better.

As for multimap, I am aware of it. My intention was contribution to enhance defaults and make hfilter being more consistent in its' judgements - I think it would be unfair to add a score for one default PTRs used by one big provider but not other.

dragoangel:
`/[.-]clients?[.-]/i 1` is a generic - common wording for default ptrs glob-wide, same as `dsl`, `pool` and so on..., while `*.vps.ovh.net` is specific PTR, that's what I try to say ;)

g00g1:
In your opinion, would it make sense to add a generic regexp like `/[.-]vps[.-]/i 1` instead?

dragoangel:
Yep, it would fit better that way then

PR #5496, opened 2025-06-05
[Fix] Correct logic error in milter_headers.lua: skip_wanted()

drclau:
A mechanism that allows overriding the active [`milter_headers`](https://docs.rspamd.com/modules/milter_headers) routines via [`user settings`](https://docs.rspamd.com/configuration/settings) has been added in commit https://github.com/rspamd/rspamd/commit/c116e4e35c7caf9b9cd454d85ea088a2c960f62f.
However, due to a bug in the implementation no routines are running when this mechanism is used.

This PR fixes the logic error and adds a functional test that covers this scenario.

**Changes:**
- fix logic error in `skip_wanted()`
- add new functional test for this scenario
- add override settings in `milter_headers.conf` to support the testing
- add supporting functionality in `rspamd.robot`

### DIFF
--- a/src/plugins/lua/milter_headers.lua
+++ b/src/plugins/lua/milter_headers.lua
@@ -138,7 +138,7 @@ local function milter_headers(task)
 
   local function skip_wanted(hdr)
     if settings_override then
-      return true
+      return false
     end
     -- Normal checks
     local function match_extended_headers_rcpt()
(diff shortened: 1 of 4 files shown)


vstakhov:
It looks very good, thank you!

drclau:
Thanks! Happy to help.

PR #5500, opened 2025-06-09
Remove aweber.com from spf_dkim_whitelist.inc

TaaviE:
`aweber.com` is an ESP who provides bulk mail services like MailChimp and other similar services and they sign their customers' letters using that specific domain. Unfortunately they sometimes let spam through and should not be whitelisted like this.


vstakhov:
In fact, this PR should actually be made to the https://github.com/rspamd/maps/blob/master/rspamd/spf_dkim_whitelist.inc as this file is merely used as a backup.

Issue #5505, opened 2025-06-11
[BUG] rspamd-proxy wrong Connection and Log-Tag headers


dragoangel:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
With adding logic of `keepalive` and `log_tag_type` rspamd-proxy worker started to create duplicated `Connection` and `Log-Tag` headers.

If `keepalive` disabled we got 2 `Connection: close`, if enabled - 1 `Connection: close` and second `keep-alive`.

If `log_tag_type` set to `none` or `skip` we still have `Log-Tag: queue-id`. If we set it to `queue_id` in settings we would get 2 `Log-Tag: queue-id` headers, and if leave default or set `session` explicitly we would get 1st `Log-Tag: queue-id` and second `Log-Tag: session-id` header

### Steps to Reproduce

1. Install rspamd v3.12.1
2. Configure rspamd-proxy and rspamd-normal accordingly
3. Setup tcpdump/http-mitm-proxy etc to monitor traffic
4. Sent test payload to rspamd-proxy via milter protocol

**Expected behavior**
Headers doesn't duplicate and has values accordingly to settings.

### Versions

v3.12.1

### Additional Information

N/A

dragoangel:
Log-Tag header is fixed fully, but Connection header with latest PR still is always `closed`, but not duplicated.

PR #5507, opened 2025-06-11
[Fix] Fix proxy headers duplication

vstakhov:
Issue: #5505


dragoangel:
Unfortunately still same issue even with this PR 😔

PR #5509, opened 2025-06-11
[Minor] Treat *.library-ms and *.search-ms attachments as harmful

twesterhever:
See: https://www.bleepingcomputer.com/news/security/microsoft-outlook-to-block-more-risky-attachments-used-in-attacks/


vstakhov:
This PR has syntax errors, so I had to revert it. You cannot use `-` in lua tables without proper escaping.

twesterhever:
> This PR has syntax errors, so I had to revert it. You cannot use `-` in lua tables without proper escaping.

Oh, apologies. I'll submit a fixed PR in due course.

vstakhov:
Don't worry, I have already made this change in a proper way.

twesterhever:
Oh, I see. Thank you very much!

PR #5512, opened 2025-06-12
[Fix] Another fix for maps concurrent load

vstakhov:


dragoangel:
Unfortunately this revert doesn't helps.
Need revert 3 commits:
```
2d2a4cf45f59a4c0f60755e8a4d736c87615ee7a
4c98aab6f670c659dff4c7e0cf392576f7850732
a63e77fa0c46197b2a04b8e750de32beaf173493
``` 
Specifically `a63e77fa0c46197b2a04b8e750de32beaf173493` can be fixed instead of reverting. Problem with it that it expect expire header and compare it's value to our max interval, problem here that we have other 2 other mechanism in Rspamd: Etag and Modified-Date, both should be honored too, but with this PR, it ignores them, but tries download each time map over 16x of max interval even there was no cache directive at all. I think Rspamd should in first place look at Etag, then at Modified-Date and only then at Caching expiration, we don't need to download file if it's Etag or Modified-Dates are same as they was, but we should check them with out max interval if there no cache header exist.

dragoangel:
Last changes fixes maps download and updates, but there is still small side effects exists:

1. `cached` status in this version works inconsistent, it's not displays correct data potentially for maps that defined multiple times or map that composed from multiple files
2. `http_map_finish: data is not modified for server maps.local (https://maps.local/redirectors.inc), next check at Mon, 16 Jun 2025 08:28:35 GMT (http cache based: (NULL))` - assume `NULL` is not a desired text to display
3. rspamd create new shm files even when was no changes to the map with log `cached data is now expired (gen mismatch 1 != 2)` and so on, I don't see such logs on 3.11 or older versions, but this not guarantee that this was not a case because it can just not had a log for that?
4. `2025-06-16 08:27:39.85175 #208(controller) <4i5gi98rz7qnarkf>; map; http_map_finish: need to reread map from https://maps.rspamd.com/rspamd/redirectors.inc.zst (reply code 200); date timestamp: 1750062460, last modified: 1749481622` if modified time is lower that we get previously we still re-read map, not sure if it's desired, but see reason to leave it as is - now maps.rspamd.com provides non consistent modify dates over 2 different IPs which forces each rspamd to download them each time they get different IP from the host.

vstakhov:
3) You might be wrong about this assumption: we have desycnhronized modify times on maps and cache expiration might be the case.
4) It is nginx `if_modified_since before`, default is `exact` and that's how Nginx works, not Rspamd.

dragoangel:
> 3\. You might be wrong about this assumption: we have desycnhronized modify times on maps and cache expiration might be the case.

I see same behavior on my maps that have only `Etag` and `Last-Modified` headers, so I don't think so.

About 4th, okay :)

dragoangel:
> 3\. You might be wrong about this assumption: we have desycnhronized modify times on maps and cache expiration might be the case.

Okay, now I get it, we have one map, which actually an array of maps, f.e. mentioned redirectors is `[
'https://maps.local/redirectors.inc','https://maps.rspamd.com/rspamd/redirectors.inc.zst','file://etc/rspamd/maps.d/redirectors.inc']` and so on, and if 1 map is changes, then Rspamd regenerate shm, as it composed from multiple resources, **but** log line will always mention only first resource in the list, which is in my case was `https://maps.local/redirectors.inc` and this was creating assumption that something is wrong, but it was issue with `maps.rspamd.com` serving different timestamps for same file on different hosts which trigger map redownload.

Great, then 3th is false statement 😊

vstakhov:
Nice, so I think this PR is finally ready for merge, thanks @dragoangel !

PR #5515, opened 2025-06-14
Add FROM_NO_ADDR symbol

dragoangel:
Should cover cases when header From looks like `"Display Name" <>`

### DIFF
--- a/rules/headers_checks.lua
+++ b/rules/headers_checks.lua
@@ -663,6 +663,8 @@ local check_from_id = rspamd_config:register_symbol {
     if (from and from[1]) then
       if not (from[1]["flags"]["valid"]) then
         task:insert_result('FROM_INVALID', 1.0)
+      else (from[1].addr == nil or from[1].addr == '') then
+        task:insert_result('FROM_NO_ADDR', 1.0)
       end
       if (from[1].name == nil or from[1].name == '') then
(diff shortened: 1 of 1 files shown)


dragoangel:
@vstakhov I think that all default scoring regarding broken headers in Rspamd is too permissive, we can punish stronger missconfigured senders, what do you think?

dragoangel:
Technically I see issue in `task:get_from([type])` function because it returns `flag.valid` for emails like: `"DN" <""@example.com>`, and not return any from object for emails like: `"DN"` or `""`

We need cover cases like and mark it as from invalid (in addition to existing invalid cases):
```
<>
DN <>
DN <""@example.com>
DN
"DN
"DN"
```

https://github.com/rspamd/rspamd/blob/88928efcfee39b99f9c8caaa9d05b2b0028ee962/rules/headers_checks.lua#L615-L626
also not check for `'""'`

I would propose fix `task:get_from()` and change `rspamd_config.MISSING_FROM` to simpler function `has_header` instead, all other logic cover in `FROM_NO_ADDR` and `FROM_INVALID` symbols

dragoangel:
@vstakhov btw as far I can see there no symbol for FROM_DOMAIN_IP or something similar even function `task:get_from()` has `flag.ip`, right? We can add such symbol for both `smtp` and `mime` from 😊, what you think?

PR #5518, opened 2025-06-17
[Minor] Treat *.svg attachments as slightly bad

twesterhever:
Prompted by a recent malspam campaign flagged by CERT.at, which highlights the abuse of Scalable Vector Graphics attachments: https://www.cert.at/de/warnungen/2025/6/phishing-angriffe-mit-manipulierten-svg-dateien-vorsicht-geboten

Since these might appear in legitimate e-mail traffic as well, only treat their occurrence as slightly bad for the time being.


dragoangel:
SVG can include http redirects from image to http page and other evil stuff even

PR #5520, opened 2025-06-18
[WebUI] Add Bayes classifier selector to Learn card

moisseev:
This enables users to choose a specific Bayes classifier for training from WebUI.


dragoangel:
Works great, thank you @moisseev  👍🎉

PR #5524, opened 2025-06-21
[WebUI] Add fuzzy flag selectors to Scan/Learn tab

moisseev:
Allows retrieving the list of configured fuzzy storages via HTTP API, including associated flags, read_only status, and server addresses.


dragoangel:
Tested, works awesome, thanks @moisseev 😊

PR #5527, opened 2025-06-24
Update protocol for ftp.arin.net in asn.pl

radub2012:
According to https://www.arin.net/blog/2025/02/10/ftp-retirement/ ftp.arin.net is no longer accessible via the FTP protocol since 31 March 2025. This changes the default protocol from ftp://ftp.arin.net to http://ftp.arin.net (which works)

### DIFF
--- a/utils/asn.pl
+++ b/utils/asn.pl
@@ -17,7 +17,7 @@ use URI;
 
 my %config = (
   asn_sources => [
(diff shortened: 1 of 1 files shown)


fatalbanana:
If you use this script consider adopting [rspamd/goasn](https://github.com/rspamd/goasn) instead - it's somewhat faster, otherwise largely identical & should hopefully see improvements in future.

PR #5528, opened 2025-06-25
icap.lua: ipv6 format in GET request format

rheoli:
For ipv6 [..| is needed in GET request who is encapsulated in ICAP.

Didn't use Lua often maybe the code can be optimized.


rheoli:
There is another problem with IPv6 URLs in SkyHigh Web Gateway

PR #5532, opened 2025-06-26
icap.lua: ipv6 pseudo hostname in GET request

rheoli:
ipv6 addresses need to be changed to a pseudo hostname in GET request who is encapsulated in ICAP. A change to [::] did not work in SkyHigh Web Gateway (aka McAfee)

Didn't use Lua often maybe the code can be optimized.


vstakhov:
> Didn't use Lua often maybe the code can be optimized.

I think it looks quite sane, as we still need to call tostring for all IP versions after all.

PR #5533, opened 2025-06-27
[Minor] Fix FreeBSD build by including sys/user.h

moisseev:


vstakhov:
BTW, don't we need to do the same on other BSD, like Open/Net etc? I guess we need.

moisseev:
I don't know for sure, but my guess is that on other BSDs `struct kinfo_proc` is already defined in `<sys/sysctl.h>`, so including `<sys/user.h>` isn't needed there.

PR #5542, opened 2025-07-14
[Fix] : CVE-2021-32765 in hiredis

KIMDONGYEON00:
### Summary
Fixes CVE-2021-32765 integer overflow vulnerability in bundled hiredis library's createArrayObject function. Adds overflow check to prevent buffer overflow when parsing maliciously crafted Redis RESP multi-bulk data.
[https://nvd.nist.gov/vuln/detail/CVE-2021-32765](https://nvd.nist.gov/vuln/detail/CVE-2021-32765)
[https://github.com/redis/hiredis/commit/76a7b10005c70babee357a7d0f2becf28ec7ed1e](https://github.com/redis/hiredis/commit/76a7b10005c70babee357a7d0f2becf28ec7ed1e)

### Location: hiredis library bundled in KBEngine (hiredis.c createArrayObject)

- Vulnerability: Integer overflow when parsing Redis RESP multi-bulk data
- Impact: Buffer overflow from count * sizeof(redisReply*) calculation exceeding SIZE_MAX → potential memory corruption
- Risk: Server crash or RCE when KBEngine's Redis integration receives malicious data (session management, caching, etc.)

###How it was fixed

- Patch: Added overflow check in createArrayObject function
- Specific fix: `if (SIZE_MAX / sizeof(redisReply*) < elements) return NULL;`
- Source: Applied patch from hiredis upstream commit [76a7b10](https://github.com/redis/hiredis/commit/76a7b10005c70babee357a7d0f2becf28ec7ed1e)

### DIFF
--- a/contrib/hiredis/hiredis.c
+++ b/contrib/hiredis/hiredis.c
@@ -179,6 +179,7 @@ static void *createArrayObject(const redisReadTask *task, size_t elements) {
         return NULL;
 
     if (elements > 0) {
+        if (SIZE_MAX / sizeof(redisReply*) < elements) return NULL;  /* Don't overflow */
         r->element = hi_calloc(elements,sizeof(redisReply*));
         if (r->element == NULL) {
             freeReplyObject(r);
--- a/contrib/hiredis/test.c
+++ b/contrib/hiredis/test.c
@@ -552,6 +552,20 @@ static void test_reply_reader(void) {
     freeReplyObject(reply);
     redisReaderFree(reader);
 
+    test("Multi-bulk never overflows regardless of maxelements: ");
+    size_t bad_mbulk_len = (SIZE_MAX / sizeof(void *)) + 3;
+    char bad_mbulk_reply[100];
+    snprintf(bad_mbulk_reply, sizeof(bad_mbulk_reply), "*%llu\r\n+asdf\r\n",
+        (unsigned long long) bad_mbulk_len);
+
+    reader = redisReaderCreate();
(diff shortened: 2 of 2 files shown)


vstakhov:
Well, while this is good to fix but I don't like the security vulnerability mood here. In our threat model Redis server is a trusted server managed within our infrastructure. If this assumption is not correct we would have much more OMG SECURITY VULNERABILITIES for sure. I'm pretty sceptical that any hiredis consumers use anyhow different model of threats.

vstakhov:
Ok, I have manually checked the `hi_calloc` function and it looks like this:

```c
static inline void *hi_calloc(size_t nmemb, size_t size) {
    /* Overflow check as the user can specify any arbitrary allocator */
    if (SIZE_MAX / size < nmemb)
        return NULL;

    return hiredisAllocFns.callocFn(nmemb, size);
}
```

So I'm interested now what problem are you trying to solve by your suggested change.

vstakhov:
Also, your patch introduces a memory leak.

KIMDONGYEON00:
You're absolutely right, and I apologize.

I initially looked only at the CVE-2021-32765 patch and mistakenly thought our codebase might be vulnerable because I didn't see the overflow check in hiredis.c. However, I overlooked the fact that the hi_calloc function already includes equivalent overflow protection.

I'm sorry for proposing a patch without properly reviewing the existing codebase, and as you pointed out, for introducing a memory leak with my careless implementation.

Going forward, I'll make sure to thoroughly review the codebase and existing security mechanisms before applying CVE patches.

Thank you for your patience and for pointing this out.

PR #5543, opened 2025-07-14
[Rework] Return back N-ary optimizations for arithmetic-alike expressions

vstakhov:
### Summary

This change restores arithmetic expression optimizations for N-ary operations (like `+`, `*`, `&`, `|`) that were previously removed. The optimization allows expressions to short-circuit evaluation when the result can be determined early, significantly improving performance for rules with multiple operands.

### What Changed

**Before:** Expressions like `A + B + C + D > 2` would evaluate all operands (A, B, C, D) regardless of intermediate results.

**After:** The same expression will stop evaluating once the sum exceeds the threshold. For example, if `A=1`, `B=1`, `C=1`, the expression stops at C since `1+1+1 > 2` is already true, and D is never evaluated.

### Impact on Existing Rules

Some existing rules were written with the assumption that **all** operands would be evaluated, even when the result could be determined earlier. These rules may now behave differently and could potentially:

1. **Miss expected matches** if they relied on side effects from evaluating all operands
2. **Change scoring behavior** if they expected certain atoms to always be processed
3. **Break test cases** that were written expecting the non-optimized behavior

### Migration Path

For rules that require the old behavior (evaluating all operands), you can now disable optimizations using the new `expression_flags` parameter:

```lua
regexp = {
  -- Rule that needs all operands evaluated
  MY_RULE = {
    regexp = "A + B + C + D > 2",
    expression_flags = {"noopt"}, -- Disable optimizations
    description = "Rule requiring evaluation of all operands",
  },
  
  -- Normal rule with optimizations (default)
  NORMAL_RULE = {
    regexp = "X + Y + Z > 1",
    description = "Optimized rule (default behavior)",
  },
}
```

### Performance Benefits

Rules that can use the optimizations will see significant performance improvements:
- **Faster evaluation** for expressions with many operands
- **Reduced CPU usage** when early termination is possible
- **Better scalability** for complex rule sets

### Identifying Affected Rules

Rules most likely to be affected are those that:
- Use arithmetic operations (`+`, `*`) with comparison operators (`>`, `>=`, `<`, `<=`, `==`, `!=`)
- Have multiple operands that perform expensive operations
- Were written or tested expecting all operands to be evaluated

### Testing Recommendations

1. **Review rule behavior** after upgrading, especially for rules with arithmetic expressions
2. **Check test cases** that may have been written expecting non-optimized behavior
3. **Monitor rule effectiveness** to ensure expected matches are still occurring
4. **Add `expression_flags = {"noopt"}` to rules** that require the old behavior

This change represents a significant performance improvement while maintaining backward compatibility through the `expression_flags` mechanism.

### DIFF
--- a/rules/bitcoin.lua
+++ b/rules/bitcoin.lua
@@ -183,8 +183,9 @@ local btc_bleach_re = [[/\b(?:(?:[a-zA-Z]\w+:)|(?:bc1))?[qpzry9x8gf2tvdw0s3jn54k
 
 config.regexp['BITCOIN_ADDR'] = {
   description = 'Message has a valid bitcoin wallet address',
-  -- Use + operator to ensure that each expression is always evaluated
   re = string.format('(%s) + (%s) > 0', normal_wallet_re, btc_bleach_re),
+  -- Disable optimizations for this expression to get all addresses
+  expression_flags = { 'noopt' },
   re_conditions = {
     [normal_wallet_re] = function(task, txt, s, e)
       local len = e - s
--- a/src/libutil/expression.c
+++ b/src/libutil/expression.c
@@ -80,6 +80,10 @@ struct rspamd_expr_process_data {
 	/* != NULL if trace is collected */
 	GPtrArray *trace;
 	rspamd_expression_process_cb process_closure;
+	/* Optimization thresholds for arithmetic operations */
+	double threshold;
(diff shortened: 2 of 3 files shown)


vstakhov:
The failing test is interesting, more like the test is broken as it relies on no optimisation logic, hummm

Issue #5558, opened 2025-08-01
[Feature] Allow to apply "pass" ARC signature/seal although there were changes on headers making ARC checks fail


uschindler:
## Summary

It would be good if one could configure that ARC signatures from a specifc domain (our own signing domain) are always treated as PASS, although they aren't. Afterwards when applying a new signature and seal, the seal and authentication results should state "cv=pass".

## Motivation

This is a usual mailing list problem. You have a mailman instance in a docker container:
- All mail coming in from internet are passed through rspamd and a new ARC seal and authentication is done. At this point all is fine. The new signature and results are added all with "pass".
- Now the message is passed to Mailman instance which happily inserts new headers or modifies subject
- The message is injected from mailman into queue again
- rspamd should now again ARC sign/seal, but it figures out that the ARC signature is broken (due to changes in mailman). It therefore signs/seals the message but sets `cv=fail`

As I trust mailman that it does not do bad things, there should be an option that you can "whitelist" the ARC signature of the previous run (you can nail it down using the domain name), but don't actually check it.

So in short: it should apply a new ARC signature/seal with `cv=pass` and not even check the broken signature. Basically something saying: "if domain name is example.org, don't verify arc and just assume `pass`".

## Describe alternatives you've considered

The first idea was to just disable ARC signature checking using the `settings.conf`: All mail coming in from mailman (based on docker IP address) have `symbols_disabled = [ "ARC_CHECK" ]`. This works somehow, but as this completely disables ARC checks the signing part does not know any previous ARC signatures and starts again with `i=1` which breaks downstream.

I also tried `whitelist [ "example.org" ]` because the documentation suggest that this should work, but actually this is not even implemented - so documentation is wrong: https://docs.rspamd.com/modules/arc/#configuration (there is no such "whitelisting" implemented!

Maybe easiest is to implement exactly that "whitelist" property and make it set the ARC result "pass" for specific signers, allowing downstream signing.

vstakhov:
It looks like a documentation issue: whitelisted_signers_map - that's the proper option name

vstakhov:
Or wait, we need to disable checks and not improve scores for some senders, then we need to implement this in the module indeed.

uschindler:
Hi,

> It looks like a documentation issue: whitelisted_signers_map - that's the proper option name

That's fine, but this flag does not do what is expected here - please read the whole description. It only fixes dkim or adds the "whitelisted" symbol if the ARC validation works.

[...]

PR #5559, opened 2025-08-01
[Fix] Fix whitelist options in the arc module

vstakhov:
Issue: #5558


uschindler:
This fix works fine!

vstakhov:
Since it actually matches the documentation, I think it's good to merge it as is :)

uschindler:
Many thanks! So the next version should contain this. So my patched local version can stay alive until next update!

P.S.: I also did more extensive testing with a test mailing list where I sent messages with a large chain of hops finally landing in a redirected gmail account. As expected the new message "ARC seal validation failed for whitelisted domain %s, treating as valid: %s" appears in the longs for the mails coming FROM the mailing list (as such has modified the subject) but a valid "pass" ARC signature/seal is applied. Later server like Google then accepted the signed ARC message as "pass".

Issue #5565, opened 2025-08-08
[Feature] Add support for GPT-5-mini etc. in gpt-plugin


laitco:
## Summary

[Feature] Add support for GPT-5-mini etc. in gpt-plugin

## Motivation

Usage of new response apis

## Describe alternatives you've considered

None

## Additional context

None

hunter-nl:
@vstakhov I've created a pull request #5572 to solve this feature/issue.

laitco:
Wow, that was fast. Thank you!

hunter-nl:
@vstakhov I've updated the PR. Maybe also add some extra documentation. The example can be extended with additional extra_symbols. The following is recommended by GPT itself to be used for more control on different categories GPT can classify.
```
extra_symbols = {
    # --- PHISHING family ---
    GPT_PHISHING = {
      description = "Credential theft or impersonation",
      category = "phishing",
      score = 8.0
    },
    GPT_SPEAR_PHISHING = {
      description = "Highly targeted phishing (BEC, CEO fraud, supplier fraud)",
      category = "spear_phishing",
      score = 8.5
    },
    GPT_IMPERSONATION = {
      description = "Brand/domain abuse to trick recipient",
      category = "brand_impersonation",
      score = 7.5
    },

    # --- SCAM family ---
    GPT_SCAM = {
      description = "Generic scam or confidence trick",
      category = "scam",
      score = 6.5
    },
    GPT_FRAUD_FINANCIAL = {
      description = "Financial or investment scam",
      category = "fraud_financial",
      score = 7.0
    },
    GPT_ADVANCE_FEE_SCAM = {
      description = "Advance fee / 419 scam",
      category = "advance_fee_scam",
      score = 6.5
    },

    # --- MALWARE family ---
    GPT_MALWARE = {
      description = "Attachment or link delivering malicious code",
      category = "malware",
      score = 9.0
    },
    GPT_EXPLOIT_DELIVERY = {
      description = "Exploit delivery via file or link",
      category = "exploit_delivery",
      score = 9.0
    },
    GPT_TROJAN_ATTACHMENT = {
      description = "Attachment contains trojan or malicious script",
      category = "trojan_attachment",
      score = 9.5
    },

    # --- MARKETING family ---
    GPT_MARKETING = {
      description = "Legitimate or unsolicited marketing",
      category = "marketing",
      score = 3.0
    },
    GPT_BULK_UNSOLICITED = {
[...]

PR #5572, opened 2025-08-14
Update GPT plugin to support OpenAI GPT-5 and other newer models

hunter-nl:
OpenAI GPT-5 family and some other models has changed attribute max_tokens into max_completion_tokens. Also attribute temperature isn't supported anymore. For each parallel model request, if configured multiple models, the attributes are now set correctly.

### DIFF
--- a/src/plugins/lua/gpt.lua
+++ b/src/plugins/lua/gpt.lua
@@ -716,7 +716,7 @@ local function openai_check(task, content, sel_part)
     return true
   end
     
-  local body = {
+  local body_base = {
     model = settings.model,
     messages = {
       {
@@ -741,21 +741,7 @@ local function openai_check(task, content, sel_part)
       }
     }
   }
-
-  -- Set the correct token limit field
-  local token_field = get_max_tokens_field(settings.model)
-  body[token_field] = settings.max_tokens
-
-  -- Set the temperature field if model supports it
-  if supports_temperature(settings.model) then
-    body.temperature = settings.temperature
-  end
-
-  -- Conditionally add response_format
-  if settings.include_response_format then
-    body.response_format = { type = "json_object" }
-  end
-
+  
   if type(settings.model) == 'string' then
     settings.model = { settings.model }
   end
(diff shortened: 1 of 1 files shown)


hunter-nl:
This pull request solves the feature/issue request #5565

src/plugins/lua/gpt.lua
```diff
@@ -683,10 +683,54 @@ local function openai_check(task, content, sel_part)
 
   local from_content, url_content = get_meta_llm_content(task)
 
-  local body = {
-    model = settings.model,
-    max_tokens = settings.max_tokens,
-    temperature = settings.temperature,
+  -- Deep copy to avoid table reuse between iterations
+  local function table_deep_copy(orig)
```
vstakhov: We have this function in lua_util.

src/plugins/lua/gpt.lua
```diff
@@ -683,10 +683,54 @@ local function openai_check(task, content, sel_part)
...
+
+  -- Only send temperature if model supports it
+  local function supports_temperature(model)
+    if not model then return true end
+    -- Disallow for reasoning models and GPT-5 family
+    if model:match('^gpt%-5') or
+       model:match('^o%d') or
+       model:match('^o%d%-mini') or
+       model:match('^gpt%-4%.1') or
+       model:match('reasoning') then
```
vstakhov: I don't like this approach. We need to create either config option or a table with something like `model` -> `parameters` where we can set it. Matching 5 regular expressions is very far from optimal approach.

vstakhov (changes requested):

src/plugins/lua/gpt.lua
hunter-nl: I've changed it to use the lua_lib function.

src/plugins/lua/gpt.lua
hunter-nl: I've made a change to get the model parameters from the settings now. Due this change the documentation needs to be updated.

vstakhov (approved):

PR #5577, opened 2025-08-16
Updated gpt.lua to fix issue with the reason_header is not ASCII

hunter-nl:
When GPT responding reason, it contains often UTF8 characters. This results into issues when sending these messages to mail servers not supporting SMTPUTF8. reason_header should be in ASCII only to support any mailserver.

Fixed by requesting GPT to answer back only in ASCII.

### DIFF
--- a/src/plugins/lua/gpt.lua
+++ b/src/plugins/lua/gpt.lua
@@ -1007,14 +1007,14 @@ if opts then
     if settings.extra_symbols then
       settings.prompt = "Analyze this email strictly as a spam detector given the email message, subject, " ..
           "FROM and url domains. Evaluate spam probability (0-1). " ..
-          "Output ONLY 3 lines:\n" ..
+          "Output ONLY 3 lines containing only ASCII characters:\n" ..
           "1. Numeric score (0.00-1.00)\n" ..
           "2. One-sentence reason citing whether it is spam, the strongest red flag, or why it is ham\n" ..
           "3. Primary concern category if found from the list: " .. table.concat(lua_util.keys(categories_map), ', ')
     else
       settings.prompt = "Analyze this email strictly as a spam detector given the email message, subject, " ..
           "FROM and url domains. Evaluate spam probability (0-1). " ..
-          "Output ONLY 2 lines:\n" ..
+          "Output ONLY 2 lines containing only ASCII characters:\n" ..
           "1. Numeric score (0.00-1.00)\n" ..
           "2. One-sentence reason citing whether it is spam, the strongest red flag, or why it is ham\n"
     end


vstakhov:
We should probably just conditionally encode headers if we're not in smtputf8 mode.

hunter-nl:
Do you have already a lua function for this encoding?

vstakhov:
Here's an idiomatic way:

```lua
if rspamd_config:is_mime_utf8() then
  if not rspamd_util.is_valid_utf8(folded) then
    folded = rspamd_util.mime_header_encode(folded)
  end
else
  folded = rspamd_util.mime_header_encode(folded)
end
```

vstakhov:
Maybe we should have a separate helper in `lua_util` - we can also cache value returned by `rspamd_config:is_mime_utf8()` in some lua variable to avoid extra C call.

hunter-nl:
This function ```rspamd_config:is_mime_utf8()``` results into if rspamd/postfix is in SMTPUTF8 mode? The issue is not if postfix/rspam can handle utf8 in the headers or not. The problem is that when the message is sent to other mail servers (backend or outside), that's not SMTPUTF8 compliant. And there are a lot of them. So to be at save side it is better to set the reason_header always in ASCII only. With the new GPT prompt requesting to answer in ASCII only, it works. But there is no 100% guarantee GPT is answer in ASCII only.

hunter-nl:
So maybe change this part in the GPT module to add ```rspamd_util.mime_header_encode```?
```
if result.reason and settings.reason_header then
    lua_mime.modify_headers(task,
      { add = { [settings.reason_header] = { value = rspamd_util.mime_header_encode(tostring(result.reason)), order = 1 } } })
  end
```

vstakhov:
> The problem is that when the message is sent to other mail servers (backend or outside), that's not SMTPUTF8 compliant.

That's out of scope problem related to MTA and MTA only. If local MTA works in SMTPUTF8 mode, Rspamd should not use encoding. Otherwise, it should. But we should not encode headers without 8bit characters, so please check the snippet I've provided, as it conforms standards and does it's job. Also, please do not forget folding, maybe the routine helper should also do folding before (optional) mime encoding.

hunter-nl:
> That's out of scope problem related to MTA and MTA only. If local MTA works in SMTPUTF8 mode, Rspamd should not use encoding.

If message is delivered without any UTF8 to local MTA and then GPT adds UTF8 header, this gives a lot of issues passing the message to remote/backend servers not being UTF8 compliant. The message gets bounced, the sender mail administrator clearly asked why, due they do not send it with UTF8.
And even if local MTA is SMTPUTF8 disabled. GPT still adds UTF8 reason_header, if reason contains non ASCII characters, which it should not do this.
There is also no need to add UTF8 to reason_header. Reason is just some plain explanation why GPT has classified the message. There's no need to have it with non ASCII characters. So, I changed the GPT prompt. But this gives not 100% the guarantee it always return in ASCII only. So better is to encode it if it contains non ASCII. Then it is always save, regardless if local/remote MTA have SMTPUTF8 enabled. 

What do you mean with folding?

vstakhov:
> If message is delivered without any UTF8 to local MTA and then GPT adds UTF8 header, this gives a lot of issues passing the message to remote/backend servers not being UTF8 compliant. The message gets bounced, the sender mail administrator clearly asked why, due they do not send it with UTF8.

Again, it's all about MTA configuration and it should live within MTA. If Rspamd works in SMTPUTF8 mode, you should not bother about it - it's MTA problem now. Please don't try to expose this to Rspamd, as it has clearly nothing to do but to conform standards (however, I personally think that they are utterly stupid and breaking, but I'm not a standards author anyway).



> What do you mean with folding?

https://github.com/rspamd/rspamd/blob/master/src/plugins/lua/milter_headers.lua#L216

It is also SMTP standard...

hunter-nl:
Sadly, I know SMTP is not the greatest with standards.
But I do not agree with you.
When a message is received by MTA with headers with only ASCI or mime_encoded without any non ASCII UTF8 characters, then GPT module should NOT add non ASCII UTF8 characters into a header. 
That's just breaking the standards ;-)
MTA is doing just correctly, but rspam (GPT module) breaks it.

I will see in the next few days if I can improve the module and do some tests with it.

vstakhov:
> But I do not agree with you.
> When a message is received by MTA with headers with only ASCI or mime_encoded without any non ASCII UTF8 characters, then GPT module should NOT add non ASCII UTF8 characters into a header.
> That's just breaking the standards ;-)
> MTA is doing just correctly, but rspam (GPT module) breaks it.

So, you do not agree despite of not knowing SMTP protocol, right? Then I can only advice you to read the documentation: https://www.postfix.org/SMTPUTF8_README.html

Again, it's not Rspamd issue, so please stop trying to solve it. If MTA operator decides to work in SMTPUTF8 mode, then they will just enable it **both** in Postfix and Rspamd and all points from https://www.postfix.org/SMTPUTF8_README.html will be valid (including NDA for non-SMTPUTF8 peers).

If MTA administrator decides to operate in ASCII mode, then they will not turn this option on in Rspamd, so Rspamd will encode all non-ASCII headers as required by standards.

Solving SMTP interoperability issues in Rspamd are counter-productive and I will not accept such changes.

hunter-nl:
Then the current PR is ready to go.

vstakhov:
Obsoleted by #5604

PR #5592, opened 2025-08-29
[Fix] Use C++20 standard consistently to resolve ODR violations

arkamar:
This commit resolves an ODR violations when compiling with -flto and -Werror=odr [1]. The main project used a newer C++20 standard, while the backward-cpp and simdutf libraries used an older C++11 standard. This difference caused the linker to fail.

Setting C++20 standard in both libraries resolves the ODR issue.

Link: https://bugs.gentoo.org/962041 [1]


vstakhov:
That's very good catch, thank you! I guess we have seen it with some modern gcc recently with @fatalbanana .

PR #5602, opened 2025-09-04
DCC plugin cleanup

flesueur:
This PR proposes to clean some parts of the DCC plugin:
* DCC can be used through `external_services` or the dedicated plugin. Removed dcc from the confighelp in `external_services` since the dedicated plugin should be the right place.
* Removed the `DCC_REJECT` symbol which was never returned by the plugin
* Added `grey-off no-reject` to the DCC query to prevent G (greylist) and R (reject) answers since we are only interested by the score. It allows to remove those `G` and `R` cases
* Use `rule.default_score` to calculate the score
* Check that some `info` is returned before using it (sometimes there are no infos and the log displayed a LUA error in this case)
* Warn unexpected results

These modifications have been tested on a mail flow. I'll be happy to improve this PR if needed.


vstakhov:
It looks good to me, but I cannot test it with real dcc. I really appreciate unifying between external services and a standalone plugin.

flesueur:
I'll correct linter results and rework a bit to make it closer to the currently released implementation, I'll update you when it's ready.

In the meantime, this plugin (in the currently released implementation) returns results through both `common.yield_result()` and `task:insert_result()`. Is there a preferred way among those ?

flesueur:
The current PR is working well on our mail flow and is now closer to the currently released version. I hope someone will be able to check/test.

PR #5605, opened 2025-09-05
Improve GPT prompt and GPT default conf

hunter-nl:
The default GPT conf file needs to be adjusted to new model_parameters.
GPT prompt improvement is required to correctly proces the returned categories line.


hunter-nl:
@vstakhov I see there is a new llm_common lib and changes in gpt.lua module.
Is it correct the model_parameters config are not needed anymore due the recent llm_common lib changes?
I think as for now GPT-5-mini model (and some other models) will not work as they require different parameters. The body_base structure have now always including max_tokens and temperature parameters. This is not allowed with for example openai gpt-5-mini model. I had this already fixed in PR #5572. But maybe you made some changes with the llm_common lua lib and gpt.lua to handle this better. But I couldn't figure it out yet how this is done.
Second question with the categories handling. The current gpt.lua code seems to me to not getting it extracted from the GPT reply. I added in the same PR to set the categories variable that is returned by the reply_conversion(task, body) call and further process it correctly. Or the categories will be removed completely soon? I hope not as it provides additional value symbols.

I can re-apply the changes to handle newer openai models (which required different parameters) and fix processing the returned category if you like to this PR.

Issue #5611, opened 2025-09-10
[Feature] improve http_client tunning options


dragoangel:
## Summary

the current implementation of the `http_client` in rspmd has several limitations that affect reliability, performance, and flexibility:

1. **single timeout for the entire request lifecycle**  
   - currently, there is only one global `timeout` parameter applied to the whole request.  
   - for example, if `timeout = 12s` and the ssl handshake itself takes 12 seconds, the client never even attempts to send or receive data.  
   - as a result, connections can hang too long without an option to cut them earlier and rspamd own clients will get very high response latency.

2. **keep_alive is only a boolean flag**  
   - the client provides only an on/off toggle for keep-alive behavior.  
   - there is no ability to configure `ttl`, `idle_timeout`, or `max_reuse`.  
   - in practice, this leads to inefficient behavior: too many new connections are created instead of reusing existing ones, this is especially sensitive for ssl (https) upstreams.

3. **connection pooling is limited and suboptimal**  
   - there is a connection pool for streams, but it appears inefficient and lacks configurability.  
   - clients cannot tune pooling policies such as `pool_size`, `connection_ttl`, `idle_timeout`, or `eviction_policy`.  
   - **current behavior:**  
     - if `http_request` does not define a `connection_pool`, the request uses the shared pool from streams.  
     - if `http_request` specifies a `connection_pool`, a **dedicated connection pool** is created and used for that request.  
   - as a result, connection reuse is poor, and system load is higher than necessary.

## Motivation

1. **introduce separate timeouts for different request stages**  
   - `connect_timeout` (tcp connection)  
   - `ssl_timeout` (ssl handshake)  
   - `write_timeout` (payload write)  
   - `read_timeout` (response read)  

   this would allow the client to abort failing connections much earlier and avoid wasting the entire global timeout on just one stage.

2. **add configurable keep_alive options**  
   - `connection_ttl` (time-to-live)  
   - `idle_timeout` (max idle time before closing a connection)  
   - `max_reuse` (maximum number of reuses for one connection)  

   this provides better control over the balance between performance and stability.

3. **enhance the connection_pool**  
   - add support for configurable parameters:  
     - `pool_size`  
     - `connection_ttl`  
     - `idle_timeout`  
     - `eviction_policy`  
   - optimize session reuse to minimize new connection creation.  
   - allow clients to override defaults when necessary.  
   - clearly define the behavior between shared pool (default) and dedicated pool (when defined in `http_request`).  

[...]

PR #5614, opened 2025-09-11
[Project] Implement more flexible http timeouts

vstakhov:
Issue: #5611

### DIFF
--- a/conf/options.inc
+++ b/conf/options.inc
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2025 Vsevolod Stakhov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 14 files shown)


dragoangel:
Nice 🎉 Lets also add default maps timeouts configuration globally via `options` or to not touch what user may change - for each default http map in `multimap.conf`, so Rspamd public infra will be on safer side in case of some network degradation, overload, etc

vstakhov:
> Lets also add default maps timeouts configuration globally via `options` or to not touch what user may change - for each default http map in `multimap.conf`, so Rspamd public infra will be on safer side in case of some network degradation, overload, etc

Yeah, added

dragoangel:
From my tests everything looks good 🔥🎉

PR #5625, opened 2025-09-19
[fix] build issues with lua version higher than 5.1

netoholic:
refer to https://bugs.gentoo.org/963055

I´ve been asked to pass this upstream


vstakhov:
We have `rspamd_lua_table_size` that does that. On the other hand, I see no harm in having that define as well.

Issue #4201, opened 2022-06-24
[Feature] Managed Redis SSL/TLS Support


aleshaczech:
## Summary

SSL/TLS Support for Redis.
(related to but not duplicate of https://github.com/rspamd/rspamd/issues/3435)

## Motivation

When using Redis server managed by a Cloud provider (e.g. Digital Ocean), there's no way how to install stunnel on Redis side. Cloud providers do not support stunnel as it's a non-standard solution to the problem. The only feasible way how to achieve encryption with managed Redis is to have rspamd conform to the standard native SSL support in the Redis protocol.

## Describe alternatives you've considered

With managed Redis instances, there are really no alternatives. It's a tradeoff between security (not using encryption at all which is a major security concern) and installing and managing own Redis instances.

## Additional context

I commented in https://github.com/rspamd/rspamd/issues/3435, but I am not sure if anyone reads comments on closed issues, I thus decided to create this new issue.

As I said, every major protocol supports SSL/TLS and encrypted traffic should be nowadays considered a must, especially in the cloud environment. Not using the standard way and instead promoting workarounds (stunnel) disqualifies Rspamd from many usecases.

vstakhov:
I could only wish a good luck to any brave soul who would like to dig into this issue :) 

For me personally, this is not an issue as there is a clear workaround that has proven to work better than Redis TLS (at the moment I have checked it, it even failed to use sesions resumption and hence sucked for any sane use case), so I will not invest my time in it.

aleshaczech:
@vstakhov I got it, but can you please elaborate more on the clear workaround?
I am confused if I missed something, but do I get it right that this clear workaround actually does not work for the managed Redis? If so, can you please point me further? I don't have an issue with tunneling everything as long as it's possible.

vstakhov:
https://www.digitalocean.com/community/tutorials/how-to-connect-to-managed-redis-over-tls-with-stunnel-and-redis-cli

aleshaczech:
@vstakhov Would https://github.com/rspamd/rspamd.com/pull/567 work?

vstakhov:
Yes, thank you for submitting the PR!

PR #5626, opened 2025-09-19
[Feature] Redis - add configurable TLS for connections

justindossey:
- Add TLS options to Redis config schema (ssl, no_ssl_verify, ssl_ca, ssl_ca_dir, ssl_cert, ssl_key, sni)
- Thread TLS options through lua_redis and request helpers
- Implement TLS handshake in redis pool using hiredis SSL (redisCreateSSLContextWithOptions + redisInitiateSSLWithContext)
- Keep plain connections default and support Unix sockets
- Add unit test to validate TLS options propagation (test/lua/unit/redis_tls.lua)

Tested against local TLS Redis on localhost:6379 (PING returns PONG).

Invited to try this in closed issue #4201


vstakhov:
That was on my long terms TODO list, thank you! I will examine it more thoroughly later, as I'm not quite sure that our bundled hiredis even support SSL.

src/libserver/redis_pool.cxx
```diff
@@ -442,6 +498,21 @@ redis_pool_connection::redis_pool_connection(redis_pool *_pool,
 	}
 }
 
+static bool ensure_ssl_inited()
+{
+    static bool inited = false;
+    if (!inited) {
+        if (redisInitOpenSSL() == REDIS_OK) {
```
vstakhov: We actually don't need it - SSL is guaranteed to be initialised before that, so the whole function can be safely removed.

src/libserver/redis_pool.cxx
```diff
@@ -487,9 +558,36 @@ auto redis_pool_elt::new_connection() -> redisAsyncContext *
...
+							redisFreeSSLContext(ssl_ctx);
+						}
+					}
+				}
+
+				if (nctx) {
+					active.emplace_front(std::make_unique<redis_pool_connection>(pool, this,
+																		 db.c_str(), username.c_str(), password.c_str(), nctx));
+					active.front()->elt_pos = active.begin();
+				}
```
vstakhov, line 580: We repeat this code part 2 times, it's good to have it as a separate functor/lambda to avoid cut&paste.

src/libserver/redis_pool.h
```diff
@@ -24,6 +24,15 @@ extern "C" {
 struct rspamd_config;
 struct redisAsyncContext;
 struct ev_loop;
+struct rspamd_redis_tls_opts {
+    int use_tls;            /* 0/1 */
```
vstakhov:
```suggestion
    bool use_tls;
```

src/libserver/redis_pool.h
```diff
@@ -24,6 +24,15 @@ extern "C" {
...
+    int no_ssl_verify;      /* 0/1 */
```
vstakhov:
```suggestion
    bool no_ssl_verify;
```

src/lua/lua_redis.c
```diff
@@ -1009,6 +1013,56 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
 		}
 		lua_pop(L, 1);
 
+		/* TLS options (optional) */
+		lua_pushstring(L, "ssl");
+		lua_gettable(L, -2);
+		if (!!lua_toboolean(L, -1)) {
+			tls_opts.use_tls = 1;
```
vstakhov:
```suggestion
			tls_opts.use_tls = true;
```

src/lua/lua_redis.c
```diff
@@ -1009,6 +1013,56 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
...
+		}
+		lua_pop(L, 1);
+
+		lua_pushstring(L, "no_ssl_verify");
+		lua_gettable(L, -2);
+		if (!!lua_toboolean(L, -1)) {
+			tls_opts.no_ssl_verify = 1;
```
vstakhov:
```suggestion
			tls_opts.no_ssl_verify = true;
```

src/lua/lua_redis.c
```diff
@@ -1009,6 +1013,56 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
...
+		}
+		lua_pop(L, 1);
+
+		lua_pushstring(L, "ssl_ca");
+		lua_gettable(L, -2);
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			tls_opts.ca_file = lua_tostring(L, -1);
```
vstakhov: that won't work: lua_tostring returns ephemeral string, so we must do some proper memory management, as it will cause hard to debug issues.

src/lua/lua_redis.c
```diff
@@ -1009,6 +1013,56 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
...
+		}
+		lua_pop(L, 1);
```
vstakhov: E.g. here tls_opts.ca_file can become invalid due to garbadge collection (in fact it won't, but we can't rely on that shit).

src/lua/lua_redis.c
```diff
@@ -1009,6 +1013,56 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
...
+
+		lua_pushstring(L, "ssl_ca_dir");
+		lua_gettable(L, -2);
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			tls_opts.ca_dir = lua_tostring(L, -1);
```
vstakhov: Same.

src/lua/lua_redis.c
```diff
@@ -1009,6 +1013,56 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
...
+		}
+		lua_pop(L, 1);
+
+		lua_pushstring(L, "ssl_cert");
+		lua_gettable(L, -2);
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			tls_opts.cert_file = lua_tostring(L, -1);
```
vstakhov: Same.

src/lua/lua_redis.c
```diff
@@ -1009,6 +1013,56 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
...
+		lua_gettable(L, -2);
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			tls_opts.key_file = lua_tostring(L, -1);
+		}
+		lua_pop(L, 1);
+
+		lua_pushstring(L, "sni");
+		lua_gettable(L, -2);
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			tls_opts.sni = lua_tostring(L, -1);
```
vstakhov: Same.

test/lua/unit/redis_tls.lua
```diff
@@ -0,0 +1,36 @@
+-- Simple unit test to validate TLS options propagation in lua_redis
```
vstakhov: I'm sorry but we simply can't test Redis in unit tests, please remove it. We need some functional test but I can do it by myself.

src/libserver/redis_pool.cxx
justindossey: Got it. Removed.

src/libserver/redis_pool.cxx
justindossey: Pulled it into a helper

src/libserver/redis_pool.h
justindossey: thanks, I should have caught that

src/lua/lua_redis.c
justindossey: Hopefully duping the values into variables before using them, then g_free-ing them after connecting is sufficient.

test/lua/unit/redis_tls.lua
justindossey: Pulled it out. I didn't want to require redis on unit tests, but I thought I might be able to verify that the options propagated.

src/lua/lua_redis.c
justindossey: I'm not marking this "Resolved" until you confirm that my change does the trick.

src/lua/lua_redis.c
```diff
@@ -1009,6 +1016,62 @@ rspamd_lua_redis_prepare_connection(lua_State *L, int *pcbref, gboolean is_async
...
+		if (!!lua_toboolean(L, -1)) {
+			tls_opts.no_ssl_verify = true;
+		}
+		lua_pop(L, 1);
+
+		/* Duplicate string options to avoid ephemeral Lua string pointers */
+		lua_pushstring(L, "ssl_ca");
+		lua_gettable(L, -2);
+		if (lua_type(L, -1) == LUA_TSTRING) {
+			dup_ca_file = g_strdup(lua_tostring(L, -1));
```
vstakhov: I think we can also live without temp copies here, look at the following pattern:

```c
// Stack: table at -1
const char *field1 = NULL, *field2 = NULL, *field3 = NULL;

// Extract "field1"
lua_getfield(L, -1, "field1");
if (lua_isstring(L, -1)) {
    field1 = lua_tostring(L, -1);
} else {
    field1 = NULL; // Not a string or nil
}

// Extract "field2"
lua_getfield(L, -2, "field2");  // Note: parent table moved down!
if (lua_isstring(L, -1)) {
    field2 = lua_tostring(L, -1);
} else {
    field2 = NULL;
}

// Extract "field3"
lua_getfield(L, -3, "field3"); // Same: -3 here (maybe use another for stack pos)
if (lua_isstring(L, -1)) {
    field3 = lua_tostring(L, -1);
} else {
    field3 = NULL;
}

// All fields are now on the stack in reverse field order: field1, field2, field3
// The string pointers are valid until you lua_pop(3)
func(field1, field2, field3);

// Cleanup
lua_pop(L, 3); // Remove the field values, table remains
```

justindossey: That's cool, but I view it as a micro-optimization that has a negative complexity tradeoff. The strings we're duping are short strings, parsed rarely, and the g_strdup logic is simple and self-contained. While the above pattern is safe today because we don't yield back to Lua between reading the strings and calling into c++, if future refactors keep tls_opts beyond the call, or start doing any Lua activity before/inside the connect path, those pointers could become invalid. It trades explicit safety for a very small reduction in allocations.

So if you insist on this change, I'll do it, but I would prefer to leave it as is. It’s clear, robust, and the overhead is immaterial at startup/connection time.

vstakhov: I disagree that it happens rarely - it happens on each Redis connection attempt, and we can have many of them.

vstakhov: ca_file, ca_dir, ssl_cert - all those can be quite a large strings, as they are file paths.

justindossey: Fair enough. What if we were to keep the lua strings alive during the connect call and pass raw pointers? Since the strings are "safe" from GC as long as they're referenced, we could keep them on the Lua stack until after the `rspamd_redis_pool_connect_ext(...)` returns, then pop them. Or, we could store temporary registry references (luaL_ref), retrieve pointers, call connect,  then luaL_unref.  Then we wouldn't need the g_strdup/g_free stuff, and the C++ stays developer-friendly.

vstakhov: luaL_ref is not free, so it's good for long term objects. We just need to call lua_pop once but for all objects at once after we call for rspamd_redis_pool_connect_ext. Calls for `lua_gettable` should just be adjusted to consider that our stack will grow on each call as we postpone `lua_pop`

justindossey: ok, I pushed a change that should do what you want. Please take a look when you have time.

PR #5628, opened 2025-09-22
[Project] Add encrypted zip support to archives module

vstakhov:
### ZIP writing via libarchive with AES and Traditional encryption

- Add in-memory ZIP creation through libarchive with interoperable encryption:
  - Supported ZIP encryption types: traditional (“zipcrypt”), AES-128, AES-256.
  - Plain ZIP when no password is provided.

- Lua API
  - archive.pack(format, files[, options])
    - options.password: passphrase (ZIP only)
    - options.filters: compression filter(s) (e.g. "gzip", "xz", "zstd")
    - options.format_options or options.zip:
      - ZIP: { encryption = "traditional" | "aes128" | "aes256", compression = "store" | "deflate", compression-level = 0..9, zip64 = true|false, hdrcharset = "..." }
  - archive.zip_encrypt(files[, password])
    - With password → ZIP (traditional encryption)
    - Without password → plain ZIP
  - archive.unpack(data[, format][, password]) supports reading encrypted ZIPs (ZipCrypto/AES)

- Interoperability
  - Compatible with unzip/Info‑ZIP, 7‑Zip, libarchive.

- Tests
  - Round-trips for AES‑128/AES‑256 and Traditional; wrong password cases fail as expected.

- Security
  - Traditional ZipCrypto is legacy and weak; prefer AES for strong protection.


vstakhov:
In fact, we should just use libarchive as it seems to support all necessary formats.

Issue #13, opened 2013-12-24
R_DKIM_REJECT with valid DKIM signature and empty message body


AlexeySa:
Incorrect check of DKIM

Issue #10, opened 2013-12-05
Variable one_shot is initialized two times in options.conf


AlexeySa:
option.conf:
options {
    pidfile = "$RUNDIR/rspamd.pid";
    filters = "chartable,dkim,spf,surbl,regexp,fuzzy_check";
    raw_mode = false;
    one_shot = false;
    dns_timeout = 1s;
    dns_retransmits = 5;
    cache_file = "$DBDIR/symbols.cache";
    map_watch_interval = 1min;
    dynamic_conf = "$DBDIR/rspamd_dynamic";
    history_file = "$DBDIR/rspamd.history";
    one_shot = true;
}

FIX:
-one_shot = false;

PR #5631, opened 2025-09-25
Base64 encode subject in meta_headers to prevent http error 400/Bad Request

bilias:
When using the metadata_exporter module with meta_headers = true and http backend, Apache **returns 400 Bad Request** for emails with MIME-encoded subjects that contain non-ASCII characters, such as emojis.

This patch base64 encodes the subject in meta_headers in order to prevent this.

### DIFF
--- a/test-err.eml
+++ b/test-err.eml
@@ -1,0 +1,18 @@
+Return-Path: <SRS0=ggxX=4D=bounce.flybondi.com=emblue3prd_bm@example.com>
+Received: from emark188.embluebtl.com (emark188.embluebtl.com [185.98.145.188])
+	(using TLSv1.3 with cipher TLS_AES_256_GCM_SHA384 (256/256 bits)
+	 key-exchange X25519 server-signature RSA-PSS (4096 bits) server-digest SHA256)
+	(No client certificate requested)
+	by mail.example.com (Postfix) with ESMTPS id 85EBF202598
+	for <user@example.com>; Wed, 24 Sep 2025 15:26:18 +0300 (EEST)
+MIME-Version: 1.0
+From: Flybondi <noreply@flybondi.com>
+To: jhep@example.com
+Subject: =?utf-8?q?=C2=A1Con_estos_precios=2C_el_norte_es_tuyo!_=F0=9F=8F=9C?=
+ =?utf-8?q?=EF=B8=8F?=
+reply-to: noreply@flybondi.com
+Feedback-ID: ENVIOSIMPLE:11668:5416913:emb
+X-emmkt: E;;8d6fi54;;5b7fq;;9e2fr57;;7c1,q88
+Message-ID: <En7-1aa797eefa-O@embluemail.com>
+Date: Wed, 24 Sep 2025 09:22:57 -0300 (-03)
+


src/plugins/lua/metadata_exporter.lua
```diff
@@ -328,7 +328,13 @@ local pushers = {
         if type(v) == 'table' then
           hdrs[pfx .. k] = ucl.to_format(v, 'json-compact')
         else
-          hdrs[pfx .. k] = rspamd_util.mime_header_encode(tostring(v) or '')
+			 if k == 'subject' then
+			   local original_subject = tostring(v)
+				local b64_subject = tostring(rspamd_util.encode_base64(original_subject, 0))
+				hdrs[pfx .. k] = '=?UTF-8?B?' .. b64_subject .. '?='
```
vstakhov, line 334: What problem are you trying to solve here? From what I see it's a typical GIGO approach. Who is going to decode this and how do you plan end-to-end compatibility? Why the caller can't simply use json in this case? Why it is limited to subject?

src/plugins/lua/metadata_exporter.lua
```diff
@@ -328,7 +328,13 @@ local pushers = {
...
+			 else
+			   hdrs[pfx .. k] = rspamd_util.mime_header_encode(tostring(v) or '')
```
vstakhov, line 336: I'm not quite sure why this code is not enough - it will do efficiently the same as yours:

```
LuaJIT 2.1.1744318430> rspamd_util = require "rspamd_util"
LuaJIT 2.1.1744318430> rspamd_util.mime_header_encode("лол")
=?UTF-8?Q?=D0=BB=D0=BE=D0=BB?=
```

src/plugins/lua/metadata_exporter.lua
bilias: json does not send full mail, only metadata.
I need full mail as well as rspamd headers (specifically rcpt_to, mail_to, ip from envelope/milter) that's why I'm using the default formatter.

What I do there, is base64 the subject

So far only subject triggered error 400.
Rest of web header fields seem ok. 

Don't know about symbols. Subject might leak there as well.

src/plugins/lua/metadata_exporter.lua
bilias: Your test string `лол` `Subject: =?UTF-8?Q?=D0=BB=D0=BE=D0=BB?=` passes.

My test emoji `🏜️` `Subject: =?utf-8?q?=F0=9F=8F=9C=EF=B8=8F?=` passes.

The full subject `¡Con estos precios, el norte es tuyo! 🏜️` 
```
Subject: =?utf-8?q?=C2=A1Con_estos_precios=2C_el_norte_es_tuyo!_=F0=9F=8F=9C?=
 =?utf-8?q?=EF=B8=8F?=
```
fails without base64 encoding.

And the encoding between mime_header_encode() and encode_base64() are ofcourse different:
```
rspamd_util.mime_header_encode("¡Con estos precios, el norte es tuyo! 🏜️")
=?UTF-8?Q?=C2=A1Con_estos_precios=2C_el_norte_es_tuyo=21_=F0=9F=8F=9C=EF=B8'U�h\7f?=
LuaJIT 2.1.1753364724> print('=?UTF-8?B?' .. rspamd_util.encode_base64("¡Con estos precios, el norte es tuyo! 🏜️") .. '?=')
=?UTF-8?B?wqFDb24gZXN0b3MgcHJlY2lvcywgZWwgbm9ydGUgZXMgdHV5byEg8J+PnO+4jw==?=
```

Notice the `'U�h\7f?=` in `mime_header_encode()` which I believe is probably treated like garbage from the http remote and closes the connection with 400

This also fails
```
Subject: =?UTF-8?B?TyDPg8+Nzr3OtM61z4POvM6/z4Igz4TOv8+FIM6x?=
 =?UTF-8?B?z4HPh861zq/Ov8+FIM+DzrHPgiAizqDOsc+Bzr8=?=
 =?UTF-8?B?z4XPg865zrHPg863IM6zzrnOsSDPhM63zr0gzrI=?=
 =?UTF-8?B?z4HOsc60zrnOsSDOtc+BzrUuLi4gLnBwdHgiIM67?=
 =?UTF-8?B?zq7Os861zrkgz4POtSAyIM63zrzOrc+BzrXPgg==?=
 ```
 which is `O σύνδεσμος του αρχείου σας "Παρουσιαση για την βραδια ερε... .pptx" λήγει σε 2 ημέρες`
 and if you `mime_header_encode` it it also gives some funny characters.
 It passes with base64
 
 Don't know if the problem is the line folding or a combination, or just some characters, but base64 seems to solve the problems found in subject and send in web headers.

#5632 has also details on this.

Comment (vstakhov) [reply src/plugins/lua/metadata_exporter.lua:336]: There is no need to use b64 here for sure - that's all I can say. We might have some issue with mime_header_encode, I need to check.

Comment (vstakhov) [reply src/plugins/lua/metadata_exporter.lua:336]: Yes, there is definitely something wrong:

```
LuaJIT 2.1.1744318430> rspamd_util.mime_header_encode("¡Con estos precios, el norte es tuyo! 🏜️")
=================================================================
==92808==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x00016d5cb1a3 at pc 0x000104cdaf0c bp 0x00016d5cb130 sp 0x00016d5ca900
READ of size 68 at 0x00016d5cb1a3 thread T0
    #0 0x000104cdaf08 in strlen+0x24c (libclang_rt.asan_osx_dynamic.dylib:arm64+0x16f08)
    #1 0x0001078000a4 in rspamd_mime_header_encode mime_headers.c:889
    #2 0x000106f71578 in lua_util_mime_header_encode lua_util.c:2577
    #3 0x000102fa2188  (libluajit-5.1.2.1.1744318430.dylib:arm64+0x6188)
    #4 0x000106fcba70 in lua_do_resume_full(lua_State*, int, char const*) lua_thread_pool.cxx:271
    #5 0x000106fc6f18 in lua_resume_thread_internal_full(thread_entry*, int, char const*) lua_thread_pool.cxx:290
    #6 0x000106fc75d8 in lua_thread_call_full lua_thread_pool.cxx:358
    #7 0x00010284f214 in lua_repl_thread_call lua_repl.c:305
    #8 0x00010284fed8 in rspamadm_exec_input lua_repl.c:260
    #9 0x00010285103c in rspamadm_lua_run_repl lua_repl.c:657
    #10 0x00010284eed8 in rspamadm_lua lua_repl.c:998
    #11 0x00010283db28 in main rspamadm.c:597
    #12 0x000182b07150  (<unknown module>)
    #13 0x7329fffffffffffc  (<unknown module>)
```

bilias: Thanks for looking into it

Issue #5649, opened 2025-10-02
[BUG] maps.rspamd.com maps downloads results into IO read error: unexpected EOF


hunter-nl:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
In logging I see "connection with http server terminated incorrectly: IO read error: unexpected EOF"
```
2025-10-02 10:01:58 #2417177(controller) <orfb3c>; map; http_map_finish: data is not modified for server maps.rspamd.com (https://maps.rspamd.com/rspamd/disabled_rbls.inc.zst), next check at Thu, 02 Oct 2025 08:06:58 GMT (http cache based: Thu, 02 Oct 2025 12:01:58 GMT)
2025-10-02 10:07:05 #2417177(controller) <orfb3c>; map; http_map_finish: data is not modified for server maps.rspamd.com (https://maps.rspamd.com/rspamd/disabled_rbls.inc.zst), next check at Thu, 02 Oct 2025 08:12:05 GMT (http cache based: Thu, 02 Oct 2025 12:07:05 GMT)
2025-10-02 10:12:05 #2417177(controller) <orfb3c>; map; http_map_error: error reading https://maps.rspamd.com/rspamd/disabled_rbls.inc.zst(157.180.118.77:443): connection with http server terminated incorrectly: IO read error: unexpected EOF
```
This happens with all the maps from maps.rspamd.com and only when an actual download of the file is needed."

### Steps to Reproduce
I see these errors since about a week. Are there some issues with the maps.rspamd.com server?

**Expected behavior**
No ```connection with http server terminated incorrectly: IO read error: unexpected EOF``` errors.

### Versions
Rspamd daemon version 3.13.1

[...]

PR #5660, opened 2025-10-03
[Fix] Enforce server-controlled HTTP map refresh intervals

vstakhov:
Prevent aggressive HTTP map polling by implementing proper interval bounds:
- Cap absurdly high Expires headers (>8h) to min(map_interval * 10, 8h)
- Enforce configured map_interval as minimum when server requests faster refresh
- Apply 10 minute minimum interval when no Expires header and low map_interval
- Simplify logic by consolidating interval calculation in single function

This change ensures servers can control refresh rates and prevents clients
from causing issues with overly aggressive polling behavior.


vstakhov:
Issue #5649 

Rspamd should be really more gentle to maps servers.

vstakhov:
@cursoragent do review

PR #5666, opened 2025-10-06
[Fix] WebUI: repopulate classifier dropdown

moisseev:
after reconnect from read-only

Ensure classifiers are fetched when the dropdown is empty even if cache suggests skipping,
preventing an empty selector on Scan tab after RO → Disconnect → Enable.


moisseev:
@cursor review

vstakhov:
@cursor review

vstakhov:
@cursor review

PR #5668, opened 2025-10-06
[Feature] Add type specifiers support to lua_logger

vstakhov:
Add support for format type specifiers in lua_logger:
- %d - signed integer (int64)
- %ud - unsigned integer (uint64)
- %f - floating point with smart formatting (no trailing zeros)
- %.Nf - floating point with N decimal places precision
- %% - escape literal percent sign

Type specifiers can be combined with positional (%1d) and
sequential (%d) argument references. String to number conversion
is supported. Added comprehensive unit tests.


vstakhov:
@cursor review

Issue #4417, opened 2023-03-01
[BUG] ed25519 keys can not be loaded by the ARC module


Avamander:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
Loading ed25519 ARC keys from a file fails.

### Steps to Reproduce

1. Create an ed25519 key using the rspamd utility
2. Configure the ARC module to use it
3. Key fails to load (error indicates that it's interpreted as an RSA key)

**Expected behavior**
Ed25519 keys are loaded by the ARC module

### Versions

3.4

### Additional Information

Previously discussed on Telegram and confirmed to be a bug.

SoLoR1:
I was trying to make this work today, it also breaks fallback to rsa selector/key file. As soon as you add ed25519 key file to arc it fails so sign emails.

phonon112358:
Same problem still in Rspamd 3.13.2: 
- 
- OS: Debian 13 "Trixie"
- Rspamd version: 3.13.2 (official rspamd.org repo, not Debian standard repo)
- OpenSSL version: 3.5.1
- LuaJIT enabled: TRUE

DKIM signing works perfectly with both rsa2048 and ed25519 selectors (with both keys present).
**ARC signing only works if only the RSA selector is configured.** As soon as an Ed25519 selector is also present (even with a correct PEM file generated via `rspamadm ...`), ARC signing fails completely with:

    lua_rsa_privkey_load_file: cannot open private key from file: /var/lib/rspamd/dkim/example.com/ed25519-2025.key, error:1E08010C:DECODER routines::unsupported
    arc.lua:544: cannot load private key for signing

The ARC module does **not fallback** to the RSA selector if the Ed25519 selector cannot be loaded.
[...]

PR #5684, opened 2025-10-15
[Fix] ARC module now supports ed25519 keys

vstakhov:
This PR fixes the issue where ed25519 keys could not be loaded by the ARC module (RSP-76).

## Changes Made

- **Remove hardcoded RSA-only restriction** in `do_sign()` function
- **Replace manual RSA-specific key loading and signing** in `arc_sign_seal()` function  
- **Use native C `dkim_sign()` function** with `sign_type='arc-seal'`
- **Leverage existing C infrastructure** that already supports both RSA and ed25519 keys
- **No FFI dependency** - works with plain Lua installations

## Technical Details

- **Algorithm detection**: Automatically uses `rsa-sha256` or `ed25519-sha256` based on key type
- **Code simplification**: Reduces `arc_sign_seal()` from ~100 lines to ~50 lines
- **Backward compatibility**: RSA keys continue to work exactly as before
- **Error fix**: Resolves 'DECODER routines::unsupported' error when loading ed25519 keys

## Testing

- ✅ All unit tests pass (C++ and Lua)
- ✅ Luacheck passes
- ✅ Build completes successfully
- ✅ Pre-commit hooks pass

## Issue Resolution

Fixes the core issue described in RSP-76 where users got:
```
lua_rsa_privkey_load_file: cannot open private key from file: /var/lib/rspamd/dkim/example.com/ed25519-2025.key, error:1E08010C:DECODER routines::unsupported
```

Now ed25519 keys are loaded and used automatically by the native C DKIM infrastructure.

Resolves RSP-76

### DIFF
--- a/test/functional/cases/320_arc_signing/003_roundtrip.robot
+++ b/test/functional/cases/320_arc_signing/003_roundtrip.robot
@@ -63,6 +63,70 @@ ARC RSA ALGORITHM CHECK
   Should Contain  ${result.stdout}  a=rsa-sha256
   Should Not Contain  ${result.stdout}  a=ed25519-sha256
 
+ARC CHAIN ROUNDTRIP RSA DOUBLE SIGN AND VERIFY
+  # First sign: Create i=1
+  ${result1} =  Scan Message With Rspamc  ${MESSAGE_RSA}  -u  bob@cacophony.za.org  --mime
+  Should Contain  ${result1.stdout}  ARC_SIGNED
+
+  # Write first signed message
+  ${signed_file1} =  Write Mime Message To File  ${result1}  rsa_chain_i1.eml
+
+  # Verify i=1 header is present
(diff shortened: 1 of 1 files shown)


vstakhov:
Issue: #4417

vstakhov:
I'm still quite scared of this change as we have actually no real test of arc sign + arc verify compatibility. Let me think about it...

vstakhov:
It was a huge refactoring after all.

vstakhov:
@cursor review

phonon112358:
I confirmed that this PR resolves issue https://github.com/rspamd/rspamd/issues/4417 .  I have just tested the patch on Debian 13 with a sieve redirection to gmx. The ed25519 ARC works correctly!
Thanks for the patch!

vstakhov:
@cursor review

PR #5696, opened 2025-10-22
Refactor boundary handling in lua_mime.lua

shaneodoo:
It has been reported that attachments are being dropped with sending emails. Mailcow uses domain wide footers and when added to emails with attachments, they become either lost or corrupt. Reasons for this is the boundary closes before attachment are added and become orphaned outside the main boundary of the email. 

The Key Difference

### Original Logic:
"Boundary changed? Close the old one immediately!" ❌ Doesn't check if more content is coming
❌ Closes too early
❌ Leaves attachments outside

### Fixed Logic:
"Boundary changed? Let me check first..."
✅ Looks ahead at remaining parts
✅ Keeps boundary open if attachments coming
✅ Closes only when everything is inside

### DIFF
--- a/lualib/lua_mime.lua
+++ b/lualib/lua_mime.lua
@@ -224,21 +224,41 @@ exports.add_text_footer = function(task, html_footer, text_footer)
 
       if boundary then
         if cur_boundary and boundary ~= cur_boundary then
-          -- Need to close boundary
-          out[#out + 1] = string.format('--%s--%s',
-              boundaries[#boundaries].boundary, newline_s)
-          -- Need to close previous boundary, if ct_subtype is related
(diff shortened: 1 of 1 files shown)


vstakhov:
It looks like the tests are failing but I'm not sure why.

vstakhov:
@cursor review

fatalbanana:
> It looks like the tests are failing but I'm not sure why.

`if ... end` stuff is unbalanced

vstakhov:
> `if ... end` stuff is unbalanced

In fact, we should just close PRs that do not pass basic checks after some time. I like the intent to make Rspamd better but we have some standards that all contributors **must** follow.

shaneodoo:
Rspamd is prematurely closing the boundary before all attachments are in the email when sending and they are getting lost. Does this not concern you? it is affecting a lot of people.

PR #5699, opened 2025-10-24
Minor: Update dmarc_report.lua with new report template

croessner:
Add some headers to the report_templaate:
- Auto-Submitted, Precedence and X-Auto-Response-Suppress


fatalbanana:
> Add some headers to the report_template:
>     * Auto-Submitted, Precedence and X-Auto-Response-Suppress

We have a nearly-identical set of headers at the bottom of the stack

croessner:
Sorry for the noise.

Issue #5708, opened 2025-10-27
[BUG] DNS TCP packets have trailing data, causing problems with Knot-Resolver


WRMSRwasTaken:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
Okay, this is a weird one. As told in https://github.com/rspamd/rspamd/issues/5706, I am currently troubleshooting why my rspamd fails to resolve SPF records for domains with large TXT records, requiring TCP shunting (for example `github.com` or `amazon.com`) because they all have this site-verification "stuff" in their zone apex. I am running Knot-Resolver behind dnsdist in another VM but I was able to reproduce it all locally in a fresh LXC container.

Disclaimer: I am not yet sure which side is at fault here, I am completely clueless right now. Even by comparing the pcaps, I am not able to spot anything wrong, besides rspamd tearing down the TCP connection after the ACK from Knot-Resolver. Rspamd might not be at fault here, but I am originally suspecting rspamd, as the dig utility is working with Knot-Resolver.

I've attached two pcaps, both are done with `rspamadm dnstool spf -f notifications@github.com -i 192.30.252.201`.
One is with rspamd running into a timeout with Knot-Resolver and tearing down the TCP connection without an answer, the other one is a fully working DNS transaction with Unbound, where everything is working as expected.

The broken transaction between rspamd and Knot-resolver:
<img width="1411" height="117" alt="Image" src="https://github.com/user-attachments/assets/2aa6b9d2-4b65-44ec-904c-d9b0c7f33395" />
[knot.pcap.txt](https://github.com/user-attachments/files/23173017/knot.pcap.txt)

A working transaction between rspamd and Unbound:
[...]

PR #5716, opened 2025-10-29
[Fix] Fix TCP DNS uninitialized memory leak

vstakhov:
## Summary

This PR fixes a bug in librdns where uninitialized memory was being sent in TCP DNS requests.

## Problem

When a DNS request is rescheduled from UDP to TCP (e.g., when the response is truncated), the `rdns_reschedule_req_over_tcp` function was using `req->packet_len` instead of `req->pos` to determine how much data to copy and send.

- `req->packet_len` is the **allocated buffer size** (includes extra space for EDNS0, etc.)
- `req->pos` is the **actual packet size** (only the data that was written)

This caused random garbage from uninitialized heap memory to be appended to TCP DNS queries.

## Impact

The bug was particularly noticeable with short queries like TXT records, where:
- Allocated buffer: `namelen + 96 + 2 + 4 + 11` bytes
- Actual packet: typically 2-3x smaller
- Difference: filled with random heap garbage

## Changes

Fixed three instances in `contrib/librdns/resolver.c` where `req->packet_len` was incorrectly used:
1. Line 476: Buffer allocation size
2. Line 485: `memcpy` length  
3. Line 486: TCP packet size header

All three now correctly use `req->pos` to only copy and send the actual packet data.

## Testing

This fix ensures TCP DNS requests contain only valid DNS packet data without trailing garbage bytes.

### DIFF
--- a/contrib/librdns/resolver.c
+++ b/contrib/librdns/resolver.c
@@ -473,7 +473,7 @@ rdns_reschedule_req_over_tcp(struct rdns_request *req, struct rdns_server *serv)
 
 		struct rdns_tcp_output_chain *oc;
 
-		oc = calloc(1, sizeof(*oc) + req->packet_len);
+		oc = calloc(1, sizeof(*oc) + req->pos);
 
 		if (oc == NULL) {
 			rdns_err("failed to allocate output buffer for TCP ioc: %s",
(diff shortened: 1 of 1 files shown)


vstakhov:
Issue #5708

vstakhov:
@cursor review

PR #5720, opened 2025-11-01
Add structured fuzzy checks configuration

vstakhov:
## Summary
- allow configuring HTML-only fuzzy rules and structured per-check options
- introduce new checks object in fuzzy plugin while keeping legacy flags
- document example usage in default fuzzy_check.conf

## Testing
- not run


vstakhov:
@cursor review

vstakhov:
@cursor review

PR #5733, opened 2025-11-05
[Feature] Update RBL configuration with new from selectors, and small refactor

dragoangel:
This PR adds `smtp` and `mime from` checks via selectors, as they not covered by anything else:
- `emails` covers only emails returned by [task:get_emails()
](https://docs.rspamd.com/lua/rspamd_task/#m798cb) from message body
- `replyto` is usually seen in just hacked emails
- `dkim` still not fully popular, especially not by spammers

So usage of `from` here is important.

Ideally it would be nice to have `from_smtp` and `from_mime` out of the box in `checks` option, and for clarity rename `from` to `from_ip` (leaving `from` as deprecating, but still valid name). Current `from` in RBL module very confusing.

I explicitly used `lower_utf8` for cases when we work with hash of email or domain (not just plain request to RBL), for hashing this is important to have lowercase variants of email, so if `from_smtp` and `from_mime` will be implemented they need to follow same logic.

### DIFF
--- a/.github/workflows/droid-code-review.yml
+++ b/.github/workflows/droid-code-review.yml
@@ -1,0 +1,206 @@
+name: Droid Code Review
+
+on:
+  issue_comment:
+    types: [created]
+
+concurrency:
+  group: droid-review-${{ github.event.issue.number }}
+  cancel-in-progress: true
+
+permissions:
+  pull-requests: write
+  contents: read
+  issues: write
+
+jobs:
+  code-review:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    # Only run on PR comments that contain "@droid review" from authorized users
+    if: |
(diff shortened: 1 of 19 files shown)


dragoangel:
Another thing I think off - maybe we would be enough with just `mime from` without `smtp`, it thing to think about, from my view it can have a place in some cases and in some cases not fit well...

dragoangel:
Test selectors get me to wrong assumption that Rspamd do not apply tag removal and googlemaiil->gmail changes, as well as dots removal, on practice with enabled `rbl` debug logs it clear that all this have a place with my configuration from my PR. But ideally it would be nice that controller-worker test selectors would follow same logic as real selector would do, as it can shoot in the foot 😓

vstakhov:
@droid review

PR #5735, opened 2025-11-07
[Fix] Improve handling of excluded DMARC RUA addresses

croessner:


src/plugins/lua/dmarc.lua
```diff
@@ -434,11 +435,116 @@ local function dmarc_validate_policy(task, policy, hdrfromdom, dmarc_esld)
...
+          elseif e.schema == 'unknown' then
+            table.insert(out, e.raw_tail)
+          else
+            table.insert(out, e.schema .. ':' .. (e.raw_tail or ''))
+          end
+        end
+        return table.concat(out, ',')
+      end
+
+      if rua_map.__external then
```
vstakhov, line 491: I don't like this approach: if you can handle external maps then you'll be able to handle ordinary maps, that's the main idea. Like, if you have a callback to be called on map match, you're safe with both of map types.

PHPGangsta: You mean we should use the existing `check_map()` function?

We were thinking about that, but it wasn't that easy (at least for us).

The `check_map()` function needs to be called for each single rua address. But `check_map` does not return a boolean, it calls a callback, but only if the address is NOT found. This makes it hard for us to know when all checks are finished. Because only when all checks are finished, we know which rua addresses to exclude, and run `generate_dmarc_report()` with the remaining ones at the end a single time.

`check_map` only calls the callback if the key is NOT found in the map. It's not called if it has been found. This makes it hard to use in this case, at least in our view. If `check_map` would return a boolean, or call a callback with the result ("found" of "not found") it would be easier in our view.

vstakhov: Uh, I was under impression that callback is always called - I need to check.

PR #5739, opened 2025-11-10
[Fix] Preserve req->pos during reply validation to prevent packet truncation

vstakhov:
## Problem

The rdns library was truncating DNS packets on retransmit, causing the OPT additional section to be missing. This made resolvers like Knot-Resolver and PowerDNS-Recursor unable to parse retransmitted packets.

## Root Cause

The `rdns_request_reply_cmp` function modifies `req->pos` as a side effect during reply validation. The `req->pos` field serves dual purposes:
- During packet construction: tracks the write position (full packet length)
- During reply validation: tracks the read position for comparison

When a request times out and needs retransmit:
1. Initial send uses correct `req->pos` (includes OPT section)
2. Reply validation overwrites `req->pos` with end of question section
3. Retransmit uses corrupted `req->pos`, truncating the packet

## Solution

Save and restore `req->pos` around the reply comparison logic in `rdns_parse_reply`. This ensures the original packet length is preserved for potential retransmits while still allowing the comparison function to work as designed.

The fix avoids adding new struct fields and has minimal performance impact (one variable save/restore).

### DIFF
--- a/contrib/librdns/resolver.c
+++ b/contrib/librdns/resolver.c
@@ -484,19 +484,6 @@ rdns_reschedule_req_over_tcp(struct rdns_request *req, struct rdns_server *serv)
 			return false;
 		}
 
-		oc->write_buf = ((unsigned char *) oc) + sizeof(*oc);
-		memcpy(oc->write_buf, req->packet, req->pos);
-		oc->next_write_size = htons(req->pos);
-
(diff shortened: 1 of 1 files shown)


vstakhov:
@droid review

vstakhov:
@droid review

PR #5742, opened 2025-11-11
[Minor] WebUI: Restore hover colors for symbols

moisseev:
Fix regression introduced in #5725 where symbol hover states were unified to a single color, losing the visual distinction between symbol types.


moisseev:
@droid review

PR #5746, opened 2025-11-14
[Minor] whitelist: use contemporary API for maps

fatalbanana:
Necessary for something like `glob;/foo` to work


vstakhov:
@droid review

PR #5755, opened 2025-11-17
[Feature] Lua_scanners: Add eXpurgate engine support

peick:
# Description

Add integration for the eXpurgate AntiSpam into lua scanners module `external_services`.

See:
- https://eleven.de/expurgate/
- https://docs.halon.io/expurgate/index.html

## Requirements

One or more `expurgate-spamd` instances.

Please check the documentation of [expurgate-spamd](https://docs.halon.io/expurgate/install.html) for installation.

## Configuration

The plugin is configured in the `local.d/external_services.conf` file.
A minimal configuration requires only the `servers` key, which is the DNS name or IP address of the **expurgate-spamd**
server.

Here is an example configuration block with all available options and their default values:

    expurgate {
      # Mandatory: DNS name or IP address of the expurgate-spamd server.
      # Can be a single server or a comma-separated list.
      # servers = "expurgate.example.com";
 
      # Optional settings:
      default_port = 783,
      url = '/checkv2',
      use_https = false,
      timeout = 5.0,
      retransmits = 1,
      log_spamcause = false,
      add_header_x_purgate_id = true,
      add_header_x_purgate_type = false,
 
      # Symbol definitions
      symbol_fail = 'EXPURGATE_FAIL',
      symbol = 'EXPURGATE_CHECK',
      symbols = {
        unknown = {
          symbol = 'EXPURGATE_UNKNOWN',
          score = 0.0,
        },
        ...
      },
    }

### Configuration Options

* `servers` (string, **mandatory**): The DNS name(s) or IP address(es) of the `expurgate-spamd` server(s). For multiple
  servers, provide a comma-separated list (e.g., `"server1.example.com,server2.example.com"`).
* `default_port` (integer, default: `783`): The default port for `expurgate-spamd` servers.
* `url` (string, default: `'/checkv2'`): The URL path for the check endpoint.
* `use_https` (boolean, default: `false`): If `true`, the plugin will use HTTPS to connect to the server.
* `timeout` (float, default: `5.0`): The timeout in seconds for requests to the server.
* `retransmits` (integer, default: `1`): The number of retransmits for failed requests.
* `log_spamcause` (boolean, default: `false`): If `true`, the expurgate type is logged as the spam cause.
* `add_header_x_purgate_id` (boolean, default: `true`): If `true`, adds the `X-Purgate-ID` header to the email.
* `add_header_x_purgate_type` (boolean, default: `false`): If `true`, adds the `X-Purgate-Type` header to the email.
* `symbol_fail` (string, default: `'EXPURGATE_FAIL'`): The Rspamd symbol inserted if the check fails (e.g., due to a
  timeout).
* `symbol` (string, default: `'EXPURGATE_CHECK'`): The Rspamd symbol inserted for every successful check.
* `symbols` (table): A table to map expurgate result types to custom Rspamd symbols and scores.

### Default Symbols and Scores

The `symbols` table allows you to customize the scoring for each `expurgate type`. Each entry in the table can have the
following keys:

* `symbol`: The Rspamd symbol name.
* `score`: The score to apply for this symbol.

The following table shows the default mappings:

| expurgate type             | rspamd symbol                        | score |
|----------------------------|--------------------------------------|------:|
| `unknown`                  | `EXPURGATE_UNKNOWN`                  |   0.0 |
| `clean`                    | `EXPURGATE_CLEAN`                    |  -1.0 |
| `suspect`                  | `EXPURGATE_SUSPECT`                  |   3.0 |
| `bulk`                     | `EXPURGATE_BULK`                     |  -1.0 |
| `spam`                     | `EXPURGATE_SPAM`                     |   8.0 |
| `clean.empty`              | `EXPURGATE_CLEAN_EMPTY`              |   5.0 |
| `clean.empty-body`         | `EXPURGATE_CLEAN_EMPTY_BODY`         |  -1.0 |
| `clean.bounce`             | `EXPURGATE_CLEAN_BOUNCE`             |  -1.0 |
| `bulk.advertising`         | `EXPURGATE_BULK_ADVERTISING`         |   7.0 |
| `bulk.porn`                | `EXPURGATE_BULK_PORN`                |   7.0 |
| `dangerous.virus`          | `EXPURGATE_DANGEROUS_VIRUS`          |   8.0 |
| `dangerous.attachment`     | `EXPURGATE_DANGEROUS_ATTACHMENT`     |   5.0 |
| `dangerous.code`           | `EXPURGATE_DANGEROUS_CODE`           |   8.0 |
| `dangerous.iframe`         | `EXPURGATE_DANGEROUS_IFRAME`         |   5.0 |
| `dangerous.virus-outbreak` | `EXPURGATE_DANGEROUS_VIRUS_OUTBREAK` |   8.0 |
| `suspect.url`              | `EXPURGATE_SUSPECT_URL`              |   3.0 |
| `suspect.url-count`        | `EXPURGATE_SUSPECT_URL_COUNT`        |   3.0 |
| `suspect.mail-count`       | `EXPURGATE_SUSPECT_MAIL_COUNT`       |   3.0 |
| `suspect.sender`           | `EXPURGATE_SUSPECT_SENDER`           |   3.0 |
| `spam.phishing`            | `EXPURGATE_SPAM_PHISHING`            |   8.0 |

### Important Categories

Special attention should be paid to the following categories, as their default scores might not be suitable for all
environments. You can adjust their scores in the `symbols` table.

* `clean.empty`: An email with no content. By default, this is considered unwanted and given a positive score. If
  you wish to allow these emails, you should lower the score.
* `dangerous.attachment`: The treatment of certain attachments depends on the `expurgate-spamd` server
  configuration. You can adjust the Rspamd score to treat these as spam or ham. This check can also be disabled entirely
  in `expurgate-spamd`.
* `dangerous.iframe`: The presence of an `<iframe>` tag can be an indicator of a malicious email. It is up to you to
  decide how strictly to treat this by adjusting its score.

### DIFF
--- a/src/libmime/archives.c
+++ b/src/libmime/archives.c
@@ -783,20 +783,27 @@ rspamd_archive_process_rar(struct rspamd_task *task,
 					p + fname_len + extra_sz < end) {
 					/* Try to find encryption record in extra field */
 					const unsigned char *ex = p + fname_len;
+					const unsigned char *extra_end = ex + extra_sz;
 
-					while (ex < p + extra_sz) {
+					while (ex < extra_end) {
 						const unsigned char *t;
 						int64_t cur_sz = 0, sec_type = 0;
+						int vint_len;
 
-						r = rspamd_archive_rar_read_vint(ex, extra_sz, &cur_sz);
-						if (r == -1) {
+						vint_len = rspamd_archive_rar_read_vint(ex, extra_end - ex, &cur_sz);
+						if (vint_len == -1) {
 							msg_debug_archive("rar archive is invalid (bad vint)");
 							return;
 						}
 
-						t = ex + r;
+						if (cur_sz <= 0 || ex + vint_len + cur_sz > extra_end) {
+							msg_debug_archive("rar archive is invalid (bad extra record size)");
+							break;
+						}
 
-						r = rspamd_archive_rar_read_vint(t, extra_sz - r, &sec_type);
+						t = ex + vint_len;
+
+						r = rspamd_archive_rar_read_vint(t, cur_sz, &sec_type);
 						if (r == -1) {
 							msg_debug_archive("rar archive is invalid (bad vint)");
 							return;
@@ -808,7 +815,7 @@ rspamd_archive_process_rar(struct rspamd_task *task,
 							break;
 						}
 
-						ex += cur_sz;
+						ex += vint_len + cur_sz;
 					}
 				}
 			}
@@ -1143,7 +1150,11 @@ rspamd_7zip_read_folder(struct rspamd_task *task,
 		t = *p;
 		SZ_SKIP_BYTES(1);
 		sz = t & 0xF;
-		/* Codec ID */
+		/* Codec ID - validate bounds before reading */
+		if (end - p < (goffset) sz) {
+			msg_debug_archive("7zip archive is invalid (codec id truncated)");
+			return NULL;
+		}
 		tmp = 0;
 		for (j = 0; j < sz; j++) {
 			tmp <<= 8;


peick:
What’s the merge status? Is there anything I can do to help or clarify?

peick:
Thank's for the feedback.

### Bug: `header_value()` — Operator Precedence Error
Fixed


### Bug: `upstream:fail()` Not Called on Final Retransmit Exhaustion
> In `expurgate_requery()`, when retransmits reaches 0, the upstream is not marked as failed

The current code is actually correct. Looking at the flow:

1. `expurgate_requery()` is called when there's an HTTP error
2. `upstream:fail()` (line 187) is exected **first** - marking the current upstream as failed
3. Only then does it check `if retransmits > 0`

So when retransmits reaches 0, the upstream has **already** been marked as failed by the call at line 187.

No change needed.

### Missing: No Redis Cache Integration
> The scanner never calls `common.save_cache()`. This means every message always hits the eXpurgate server, even for identical content seen recently. 

After reviewing the caching implementation patterns in rspamd's external_services scanners, we've decided
to intentionally disable Redis caching for eXpurgate.

**Rationale**:
eXpurgate verdicts can indeed change for the same content due to live updates on certifications for multiple entities considered when producing a classification. The verdicts are also context-dependent. Moreover, "hiding" similar messages from eXpurgate's by caching also has a negative impact on the reputation system and live traffic analysis.

**Changes made**:
- Added explicit `no_cache = true` to the configuration
- Added documentation comment explaining the intentional omission

This aligns with rspamd's `no_cache` configuration option for scanners where results are context-dependent rather than content-deterministic. The cache lookup path in condition_check_and_continue() will now correctly skip caching for eXpurgate while still respecting size limits and
other preconditions.

### Missing: Config Defaults Expected by `common.lua`

As already mentioned, we now explicitly set `rule.no_cache = true`, which does not require any of the following configurations:

* **`detection_category`**
* **`cache_expire`**
* **`default_score`**

Because `yield_result()` is not used at all, neither direct nor indirect, the following configurations are not needed:

* **`action`**
* **`message`**

Because `log_clean()` is not used at all, neither direct nor indirect, the following configuration are not needed:

* **`log_clean`**

But to harmonize the look and feel of this plugin with the rest of the ecosystem I added them with the suggested defaults.

### `EXPURGATE_CHECK` Symbol
> `symbol = 'EXPURGATE_CHECK'` is defined in the config but never inserted via `task:insert_result`.

You are right, the symbol `EXPURGATE_CHECK` is never used and can be removed. 

Done.

### Style: 4-Space Indentation
> Rspamd Lua code uses **2-space indentation**. This file uses 4 spaces throughout. Needs reformatting to match the project convention.

Done.

### Minor: `socket` → `servers` Fallback

The `expurgate-spamd` does not support socket mode. Only tcp is supported.

PR #5761, opened 2025-11-22
[Feature] Add obfuscated URL detection to url_suspect plugin

vstakhov:
## Summary

Detect URLs hidden in message text using various obfuscation techniques commonly used by spammers to evade detection:

- **Spaced protocols**: `h t t p s : / / example.com`
- **hxxp variants**: `hxxp://malicious.com`
- **Bracket dots**: `example[.]com`, `phishing(.)site`
- **Word dots**: `secure-login dot net`
- **HTML entities**: `https://bad&#46;co&#46;za`

## Implementation

- Hyperscan-based prefiltering for fast pattern matching
- Text normalization to remove obfuscation
- URL extraction and injection with `obscured` flag
- Configurable via built-in settings or external maps
- DoS protection with strict limits (max matches, max URLs, max text length)

## New symbol

`URL_OBFUSCATED_TEXT` (score: 5.0) - indicates obfuscated URL found in message text

## Configuration

```hcl
url_suspect {
  checks {
    obfuscated_text {
      enabled = true;
      patterns_enabled {
        spaced_protocol = true;
        hxxp = true;
        bracket_dots = true;
        word_dots = true;
        html_entities = true;
      }
      # Optional: use map for pattern control
      # pattern_map = "/etc/rspamd/maps.d/url_obfuscated_patterns.map";
    }
  }
}
```

## Test plan

- [x] Test with sample emails containing obfuscated URLs
- [x] Verify URL injection works correctly
- [x] Check performance on large messages
- [ ] Validate DoS protection limits

### DIFF
--- a/src/lua/lua_util.c
+++ b/src/lua/lua_util.c
@@ -20,6 +20,7 @@
 #include "libmime/content_type.h"
 #include "libmime/mime_headers.h"
 #include "libutil/hash.h"
+#include "libserver/html/html.h"
 
 #include "lua_parsers.h"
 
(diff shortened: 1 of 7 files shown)


vstakhov:
@droid review

vstakhov:
> **Memory leak in lua_util_decode_html_entities**: The function always allocates inlen + 1 bytes via g_malloc, but after rspamd_html_decode_entitles_inplace decodes the entities, the actual used length (outlen) may be significantly smaller than the allocated size. This wastes memory, especially when processing many HTML entity-heavy strings.

No, it's definitely not a memory leak and it's actually a very stupid thing to do that. False alarm

PR #5763, opened 2025-11-24
Cover more possible use cases of task:set_milter_reply

dragoangel:
This PR covers gap in function documentation. As based on:
https://github.com/rspamd/rspamd/blob/4e2572467284d4df84421c9a92b388355088aef7/src/libserver/milter.c#L2177-L2213
Rspamd can handle more then just adding and removing of mime headers from email 😊


vstakhov:
Yes, but in general we want to link it with other message modification functions :)

PR #5766, opened 2025-11-27
[Feature] Add combinator option for multimap selector rules

vstakhov:
## Summary

This PR adds support for structured data output from selectors in multimap rules. Previously, selectors always produced concatenated strings which made it impossible to send structured JSON data to external map services.

## New `combinator` option

For selector-type multimap rules, a new `combinator` option specifies how multiple selector results should be combined:

| Combinator | Description | Output Type |
|------------|-------------|-------------|
| `string` | Concatenate results with delimiter (default, existing behavior) | string |
| `array` | Flatten all results into a flat array | table (array) |
| `object` | Convert pairs of selectors into key-value object | table (object) |

## Example Configuration

Send structured JSON to external API:

```lua
multimap {
  MY_EXTERNAL_CHECK {
    type = "selector";
    selector = "id('from');from('smtp'):addr;id('ip');ip";
    combinator = "object";  -- produces {"from": "user@example.com", "ip": "1.2.3.4"}
    map = {
      external = true;
      backend = "http://api.example.com/check";
      method = "body";
      encode = "json";
    };
  }
}
```

Send array of values:

```lua
multimap {
  CHECK_RCPTS {
    type = "selector";
    selector = "rcpts:addr";
    combinator = "array";  -- produces ["rcpt1@example.com", "rcpt2@example.com"]
    map = {
      external = true;
      backend = "http://api.example.com/check-recipients";
      method = "body";
      encode = "json";
    };
  }
}
```

## Changes

- **lua_selectors/init.lua**: Added combinator registry and helper functions
  - `get_combinator(name)`: returns combinator function by name
  - `list_combinators()`: returns available combinator names
  - `create_selector_closure_with_combinator()`: creates closure with named combinator

- **multimap.lua**: Added `combinator` option support for:
  - Regular selector-type rules (`type = "selector"`)
  - Redis selector maps (`redis+selector://`)

## Backward Compatibility

Fully backward compatible - existing configurations without `combinator` option continue to work as before (defaulting to string concatenation).

### DIFF
--- a/lualib/lua_selectors/init.lua
+++ b/lualib/lua_selectors/init.lua
@@ -662,6 +662,61 @@ exports.add_map = function(name, map)
   end
 end
 
+-- Combinator functions registry for use with external maps and other consumers
+-- that need structured data instead of plain strings
+local combinators = {
+  -- Default: concatenate all selector results into a single string
+  string = exports.combine_selectors,
+  -- Flatten all results into a flat array
+  array = exports.flatten_selectors,
+  -- Convert pairs of selectors into key-value object (use `id` extractor for keys)
+  object = exports.kv_table_from_pairs,
+}
+
+--[[[
+-- @function lua_selectors.get_combinator(name)
+-- Returns a combinator function by name
+-- @param {string} name combinator name: 'string', 'array', or 'object'
+-- @return {function} combinator function or nil if not found
+--]]
+exports.get_combinator = function(name)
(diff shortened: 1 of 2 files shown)


vstakhov:
@droid review

vstakhov:
> But line 717 passes `cfg` as both the first and second parameter:

Yes, because it is also a log object. Passing nil there is stupid.

PR #5769, opened 2025-11-29
[Feature] Auto-mark whitelist symbols with SYMBOL_TYPE_FINE flag

vstakhov:
## Summary

- Automatically marks symbols with negative weight as `SYMBOL_TYPE_FINE` during config validation
- New `rspamd_composites_mark_whitelist_deps()` function marks all symbols used in whitelist composites (composites with negative score) as FINE
- Ensures whitelist symbols always execute regardless of reject threshold (early-stop optimization)

## Problem

When the reject score threshold is reached, rspamd's early-stop optimization skips remaining symbols. This could cause whitelist symbols to not execute, leading to false positives where emails should have been whitelisted but weren't.

## Solution

Symbols marked with `SYMBOL_TYPE_FINE` flag are never skipped, even when the reject threshold is reached. This PR automatically marks:

1. All symbols with negative weight (whitelist symbols)
2. All symbols used in composites that have negative score (whitelist composites)

The transitive expansion ensures that nested composites are properly handled.

## Test plan

- [x] Build passes
- [x] Unit tests pass (100 tests)
- [ ] Manual testing with whitelist symbols under reject threshold

### DIFF
--- a/src/libserver/cfg_utils.cxx
+++ b/src/libserver/cfg_utils.cxx
@@ -1016,6 +1016,8 @@ rspamd_config_post_load(struct rspamd_config *cfg,
 			rspamd_composites_set_inverted_index(cfg->composites_manager,
 												 cfg->composites_inverted_index);
 			rspamd_composites_process_deps(cfg->composites_manager, cfg);
+			/* Mark symbols used by whitelist composites (negative score) as FINE */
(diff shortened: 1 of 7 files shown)


vstakhov:
@droid review

PR #5754, opened 2025-11-17
[Feature] Add lua_shape validation library as tableshape replacement

vstakhov:
## Overview

This PR introduces `lua_shape`, a comprehensive schema validation library designed to replace `tableshape` across Rspamd. The library addresses key limitations of tableshape with improved error reporting, documentation generation, and extensibility.

### 1. Better one_of Error Reporting
- Intersection analysis shows common requirements across variants
- Identifies conflicting field requirements
- Clearer error messages for validation failures

### 2. Schema-Driven Documentation
- Generate documentation IR from schemas
- Mixin-aware field grouping with origin tracking
- Markdown rendering support
- Preserves field provenance for better docs

### 3. Rich Type Constraints
- **Lua pattern support** for string validation (`pattern = "^%d+$"`)
- Numeric ranges (min/max)
- String length constraints (min_len/max_len)
- Enum and literal types
- Optional lpeg support for complex parsing

### 4. Open/Closed Tables
- `open = false` (default): Reject unknown fields
- `open = true`: Allow additional fields not in schema
- `extra = schema`: Validate extra fields with schema

### 5. First-Class Mixins
- Mixin composition with origin tracking
- Field conflict resolution (host overrides mixin)
- Supports documentation and JSON Schema export

### 6. JSON Schema Export
- Full Draft 7 support
- Maps all lua_shape types to JSON Schema
- Registry-wide export capability
- UCL validation support

### 7. Transform Support
- Immutable transformations
- Validation before transform
- Chained transforms with `:transform_with`

### 8. Pure Lua Implementation
- No mandatory dependencies
- Optional lpeg for complex patterns
- Clean, focused API

## Library Structure

```
lualib/lua_shape/
├── core.lua (915 lines)        # All type constructors, validation, utilities
├── registry.lua (223 lines)    # Schema registration and references
├── jsonschema.lua (230 lines)  # JSON Schema Draft 7 export
├── docs.lua (282 lines)        # Documentation IR generation
├── README.md (401 lines)       # Complete API documentation
└── MIGRATION.md (394 lines)    # tableshape → lua_shape migration guide

test/lua/unit/
└── lua_shape.lua (552 lines)   # 44 unit tests, 119 assertions
```

## Test Results

```
44 tests 44 passed 119 assertions 0 failed 0 errors ✅
```

All tests cover:
- Scalar types (string, number, integer, boolean, enum, literal)
- Constraints (ranges, lengths, patterns)
- Structured types (arrays, tables, one_of)
- Optional and default values
[...]

PR #5774, opened 2025-12-02
[Fix] Use 0 instead of LUA_OK for lua_pcall result comparison

arkamar:
Lua 5.1 does not define LUA_OK, so the result of lua_pcall is compared to 0 instead, which is consistent with the rest of the codebase.

This solves the Gentoo Bug: https://bugs.gentoo.org/967009

### DIFF
--- a/src/rspamadm/confighelp.c
+++ b/src/rspamadm/confighelp.c
@@ -207,7 +207,7 @@ rspamadm_confighelp_load_plugins_doc(struct rspamd_config *cfg)
 	lua_getglobal(L, "require");
 	lua_pushstring(L, "rspamadm.confighelp_plugins");
 
-	if (lua_pcall(L, 1, 1, 0) != LUA_OK) {
+	if (lua_pcall(L, 1, 1, 0) != 0) {
 		rspamd_fprintf(stderr, "cannot load confighelp_plugins module: %s\n",
 					   lua_tostring(L, -1));
 		lua_pop(L, 1);
@@ -221,7 +221,7 @@ rspamadm_confighelp_load_plugins_doc(struct rspamd_config *cfg)
 		return NULL;
 	}
 
-	if (lua_pcall(L, 0, 1, 0) != LUA_OK) {
+	if (lua_pcall(L, 0, 1, 0) != 0) {
 		rspamd_fprintf(stderr, "cannot execute confighelp_plugins function: %s\n",
 					   lua_tostring(L, -1));
 		lua_pop(L, 1);


arkamar:
The issue was introduced in PR #5754.

vstakhov:
Where do we use Rspamd with lua 5.1? In fact, I plan to move the default Lua version from luajit to lua 5.4 at some point.

arkamar:
We allow Lua 5.1 in Gentoo because it still seems to be supported:
https://github.com/rspamd/rspamd/blob/68c42ce5c6c9b1b95b8f2c17e696ac5138665ee8/CMakeLists.txt#L207-L215

Delicates:
Version 3.14.1 has introduced a handful of issues when running on Lua-5.1:
```
# /etc/init.d/rspamd restart
init of /usr/share/rspamd/plugins/aliases.lua failed: error loading module 'lua_aliases' from file '/usr/share/rspamd/lualib/lua_aliases.lua':
        /usr/share/rspamd/lualib/lua_aliases.lua:344: '=' expected near 'continue'; trace: [1]:{[C]:-1 - <unknown> [C]}; [2]:{[C]:-1 - require [C]}; [3]:{/usr/share/rspamd/plugins/aliases.lua:33 - <unknown> [main]};
load of /usr/share/rspamd/plugins/multimap.lua failed: /usr/share/rspamd/plugins/multimap.lua:751: '=' expected near 'continue_meta'
cannot register delayed dependency FREEMAIL_REPLYTO_NEQ_FROM_DOM -> FREEMAIL_REPLYTO: destination FREEMAIL_REPLYTO is missing
cannot register delayed dependency FREEMAIL_REPLYTO_NEQ_FROM_DOM -> FREEMAIL_FROM: destination FREEMAIL_FROM is missing
 * Stopping rspamd ...                                                                                                                                                                                                                                  [ ok ]
 * Starting rspamd ...
2025-12-05 09:34:02 #13674(main) <c3da3c>; main; main: rspamd 3.14.1 is loading configuration, build id: release                                                                                                                                        [ ok ]
```

vstakhov:
> We allow Lua 5.1 in Gentoo because it still seems to be supported:

Ok, I plan to drop support for lua < 5.3 soon in Rspamd, I don't have resources to support archeological stuff. So in general, we should use Rspamd with Lua 5.4 or LuaJIT 2.1. I'm even not sure that we should still use LuaJIT by default, as I have gut feeling that plain Lua 5.4 will be better in our scenario.

vstakhov:
> Version 3.14.1 has introduced a handful of issues when running on Lua-5.1:

Yeah, that's expected, as nobody tests Rspamd with 5.1.

arkamar:
> > We allow Lua 5.1 in Gentoo because it still seems to be supported:
> 
> Ok, I plan to drop support for lua < 5.3 soon in Rspamd, I don't have resources to support archeological stuff. So in general, we should use Rspamd with Lua 5.4 or LuaJIT 2.1. I'm even not sure that we should still use LuaJIT by default, as I have gut feeling that plain Lua 5.4 will be better in our scenario.

Absolutely, I am fine with that change. Anyway, will tests still require luajit or will they work with lua 5.{3,4} as well? We currently have luajit as a requirement for tests, but it might be historical (in case I missed something).

vstakhov:
Luajit will be supported for sure, as we have many years of successful work with that implementation. I just don't think it's anyhow beneficial to support Lua 5.1.

arkamar:
Yes, I agree with that.

My question was about tests, as I thought that the luajit constraint comes from upstream, but no, it is our downstream thing... Thanks.

PR #5778, opened 2025-12-03
[Fix] null-terminate the log tag when copying the header

zdAlexM:
This avoids leaking uninitialized heap memory.


vstakhov:
@droid review

vstakhov:
Yeah, my own feeling was right :)

PR #5784, opened 2025-12-08
Add remove_ar_from option for selective AR header removal

### LINKED ISSUE #5652: [Feature] Remove existing "Authentication-Results" only if own server name
According to the documentation at
https://docs.rspamd.com/modules/milter_headers/#authentication-results
I can set "remove = 1" to remove existing "Authentication-Results" headers:

```
routines {
  authentication-results {
    # Name of header
    header = "Authentication-Results";
    # Remove existing headers
    remove = 1;
    ....
  }
}
```

I tested it, and it works by removing all existing "Authentication-Results" headers.

I would like to get rid of existing "Authentication-Results" headers, but only if there is an existing "Authentication-Results" header **with my current own server name**: If there is another milter before rspamd which adds that header, or if a spammer is faking that header with my current server name, I would like to remove it. Other "Authentication-Results" headers of other providers who added them while redirecting/forwarding should stay.

I didn't find a way to achieve this in rspamd. Should there be a new setting which can be used to control if **all** "Authentication-Results" headers should be removed, or **only the ones with my own server name**?

vstakhov:
## Summary
- Adds `remove_ar_from` option to the `authentication-results` routine in milter_headers
- Allows selective removal of Authentication-Results headers by hostname, similar to OpenDKIM's `RemoveARFrom`
- Supports single hostname, array of hostnames, domain patterns (`.example.com`), or map files

Fixes #5652

### DIFF
--- a/src/plugins/lua/milter_headers.lua
+++ b/src/plugins/lua/milter_headers.lua
@@ -557,9 +557,13 @@ local function milter_headers(task)
               if local_mod.remove_ar_from:get_key(ar_hostname) then
                 should_remove = true
               else
-                local domain_part = ar_hostname:match('%.(.+)$')
-                if domain_part and local_mod.remove_ar_from:get_key('.' .. domain_part) then
-                  should_remove = true
+                for i = 1, #ar_hostname do
+                  if ar_hostname:sub(i, i) == '.' then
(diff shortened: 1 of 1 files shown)


vstakhov:
@droid review

PR #5785, opened 2025-12-08
[Fix] Handle HTTP early server responses during request write

vstakhov:
## Summary
- Fix HTTP client to properly handle early server responses (e.g., 413 Too Large) that arrive before the client has finished sending the request body
- Use bitwise AND for event flag checks to handle combined `EV_READ|EV_WRITE` events from libev
- Watch for both READ and WRITE events during write phase to detect early responses
- Check for early response on write errors (EPIPE, ECONNRESET, ENOTCONN)

This addresses an edge case allowed by HTTP/1.1 (RFC 7230 Section 6.5) where servers can respond before reading the full request body.

## Test plan
- [x] Added comprehensive functional tests in `221_http_early_response.robot`
- [x] Tests cover: early 200 reply, early 413 error, keepalive with early response, coroutine-based requests, stress tests with large bodies
- [x] All C++ unit tests pass (100/100)
- [x] All Lua unit tests pass (755/759)
- [x] Pre-commit hooks pass (ClangFormat, LuaCheck)

### DIFF
--- a/src/libserver/http/http_connection.c
+++ b/src/libserver/http/http_connection.c
@@ -1003,8 +1003,9 @@ rspamd_http_event_handler(int fd, short what, gpointer ud)
 	rspamd_http_connection_ref(conn);
 
 	if (what & EV_READ) {
-		/* Check for early response while still sending request */
-		if (priv->wr_pos < priv->wr_total && priv->wr_total > 0) {
+		/* Check for early response while still sending request (client only) */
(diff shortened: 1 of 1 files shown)


vstakhov:
@droid review

vstakhov:
Yes, that's a long term concern - http parser itself calls for all callbacks and it can cause different funny side effects. That's why we have refcounting logic. The second notice might be valid but since we have timeout I don't really think it's dangerous. However, I'll leave it here.

PR #5786, opened 2025-12-08
[Fix] Remove mime encoding of meta header value in metadata_exporter

dragoangel:
This PR fixes issue due to which metadata_exporter generate binary invalid payload, but even when if doesn't - it creates incorrect payload by design.

1. We already have `symbols` table which we not encode, we just `json-compact` it, and it have quite high chance to contain UTF8 characters inside in some of symbol options. So we can safely state that if metadata_exporter receiver must be able to process such payload to not face errors - it should be able to accept UTF8 payload in header value even when by RFC it have to be pure ASCII.
2. If we have SMTPUTF8 envelope from or rcpt - they should not be encoded with mime encoding, this is just wrong. The only valid place it could be - it's subject here and it doesn't fix other non ASCII chars, so better just skip this encoding at a start.

Test:
```
cat > test.lua << 'EOF'
local rspamd_util = require "rspamd_util"
local encoded = rspamd_util.mime_header_encode("güstaff@test.de")
print("Re-encoded:", encoded)
local encoded = rspamd_util.mime_header_encode("🤩 SALE ALERT! 60% OFF Everything + Extra 15% OFF 🤩")
print("Re-encoded:", encoded)
EOF
```

Output:
```
rspamadm lua test.lua
Re-encoded:     g=?UTF-8?Q?=C3=BC?=staff@test.de
Re-encoded:     =?UTF-8?Q?=F0=9F=A4=A9_SALE_ALERT=21_60=25_OFF_Everything_=2B_Extra_15=25_OFFJ�u?==?UTF-8?Q?=9F=A4=A9?=
```

Expected re-encoded string for subject would be something like:
```
=?UTF-8?B?8J+kqQ==?= SALE ALERT! 60% OFF Everything + Extra 15% OFF =?UTF-8?B?8J+kqQ==?=
=?UTF-8?Q?=F0=9F=A4=A9 SALE ALERT! 60% OFF Everything + Extra 15% OFF =F0=9F=A4=A9?=
```

I can guess that `J�u?=` in my `test.lua` was result of `rspamd_util.mime_header_encode` function doing line-folding - aka adding tabulation or other sort of new lines to honor SMTP line limits and then converted to a string in such abnormal way as result:
https://github.com/rspamd/rspamd/blob/d954ba4bdf173c9382c51bd5111ef06318cd281b/src/libmime/mime_headers.c#L820-L1032

### DIFF
--- a/src/libmime/mime_headers.c
+++ b/src/libmime/mime_headers.c
@@ -883,8 +883,14 @@ rspamd_mime_header_encode(const char *in, gsize len, bool is_structured)
 			size_t pending_ws_budget = include_pending_ws ? (pending_spaces + pending_tabs * 3) : 0;
 
 			/* Determine how much of this piece needs encoding and fits the budget */
+			size_t utf8_char_start = 0;
+			size_t enc_span_at_utf8_start = 0;
+			size_t encoded_len_at_utf8_start = 0;
+
 			for (size_t i = 0; i < piece_len; i++) {
 				unsigned char c = p[i];
+				/* UTF-8 lead byte or ASCII */
+				gboolean is_utf8_start = (c < 0x80) || (c >= 0xC0);
 
 				if (!need_encoding) {
 					if (c >= 128 || (is_structured && !g_ascii_isalnum(c))) {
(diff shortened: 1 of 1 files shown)


dragoangel:
@vstakhov also for me unclear why " " (spaces) replaced by "_" (underscore), I think we need check `rspamd_util.mime_header_encode` in general, because I think it in general somehow wrong and it used as function in `rspamd_util.fold_header_with_encoding` which used in many other modules:
* dmarc
* arc
* bimi
* gpt
* milter-header

which can lead to bugs in mentioned modules

UDP: fixed in https://github.com/rspamd/rspamd/commit/53e6957a80609f87ebb847872289adfeb3ef3ea6

PR #5793, opened 2025-12-11
Use PUC-Rio Lua in packages

fatalbanana:


vstakhov:
I'm going to revert this unless we have some major Rspamd release.

PR #5810, opened 2025-12-31
[Feature] Add extra tables API for clickhouse plugin

vstakhov:
## Summary

This PR adds a dynamic API that allows other Rspamd plugins to register custom Clickhouse tables at runtime.

### New API

Exposed via `rspamd_plugins['clickhouse']`:

- `register_extra_table(opts)` - Register a new table with schema, insert query, and row callback
- `unregister_extra_table(name)` - Remove a registered table
- `get_extra_tables()` - List all registered tables  
- `is_enabled()` - Check if clickhouse plugin is active

### Features

- **Lazy schema upload**: Table schema is created on first data flush, not at registration time
- **Multi-row support**: `get_row` callback can return a single row `{...}` or array of rows `{{...}, {...}}`
- **Per-table retention**: Each registered table can have independent retention period and method (drop/detach)
- **Error resilience**: Callbacks are wrapped in pcall; errors are logged but don't affect other tables

### Example Usage

```lua
rspamd_config:add_on_load(function(cfg, ev_base, worker)
  if worker:is_scanner() and rspamd_plugins['clickhouse'] and 
     rspamd_plugins['clickhouse'].is_enabled() then
    rspamd_plugins['clickhouse'].register_extra_table({
      name = 'my_plugin_stats',
      table_name = 'rspamd_my_plugin',
      schema = [[CREATE TABLE IF NOT EXISTS rspamd_my_plugin (
        Date Date, TS DateTime, MessageId String, CustomData String
      ) ENGINE = MergeTree() PARTITION BY toMonday(Date) ORDER BY TS]],
      insert_query = function() 
        return 'INSERT INTO rspamd_my_plugin (Date, TS, MessageId, CustomData)' 
      end,
      get_row = function(task)
        local ts = task:get_date({format = 'connect', gmt = true})
        return { os.date('!%Y-%m-%d', ts), ts, task:get_message_id() or '', 'data' }
      end,
      retention = { enable = true, period_months = 3, method = 'detach' }
    })
  end
end)
```

## Test plan

- [ ] Verify luacheck passes
- [ ] Verify build succeeds
- [ ] Verify unit tests pass
- [ ] Manual testing with a test plugin that registers an extra table

### DIFF
--- a/src/plugins/lua/clickhouse.lua
+++ b/src/plugins/lua/clickhouse.lua
@@ -35,6 +35,9 @@ local last_collection = 0
 local final_call = false -- If the final collection has been started
 local schema_version = 9 -- Current schema version
 
+local extra_tables = {}
+local extra_table_rows = {}
+
 local settings = {
   limits = { -- Collection limits
     max_rows = 1000, -- How many rows are allowed (0 for disable this)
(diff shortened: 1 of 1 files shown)


vstakhov:
Ok, I have some manual testing ideas for that, so merging.

PR #5816, opened 2026-01-03
Issue 5808

Premkumar-2004:
added comments also please check it out


vstakhov:
I don't think it's right. If we have no symbol result, it means there is no match. The assumption of score 0 is wrong.

Premkumar-2004:
to clarify we're not assuming score 0, but rather that it may have been evaluated before dependencies were ready:

Premkumar-2004:
added the comments in code,
Added re-evaluation logic: when a nested composite was checked but didn't match, and a parent needs it, we re-evaluate because it may have been evaluated before its dependencies were ready.

PR #5817, opened 2026-01-03
fixed issue-5799

Premkumar-2004:


vstakhov:
I think it should be a more common approach - we have a structure of languages detected by unicode glyphs. We should just include flags there, as it'll be useful for other glyph based languages in general.

Premkumar-2004:
ok i'll do it and push it again

Premkumar-2004:
i have updated it

Premkumar-2004:
i have updated the changes can you please review it

vstakhov:
@droid review

PR #5819, opened 2026-01-04
added get_domain function

Premkumar-2004:


vstakhov:
This PR lacks any description of what problem it solves. Rspamd already has extensive domain extraction APIs (`rspamd_url:get_host()`, `get_tld()`, selectors like `from:domain`, etc.).

Before submitting PRs:
- Explain what problem you're solving
- Check if existing APIs already cover your use case
- Test your changes

Please don't send personal emails about PRs.
Closing.

PR #5820, opened 2026-01-04
added documentation

Premkumar-2004:


dragoangel:
this is very complex module with 1001 option to be defined, this 1 example not provide really any use, while [plugin docs](https://docs.rspamd.com/modules/metadata_exporter/) should be reviewed heavily before using this module

PR #5821, opened 2026-01-05
[Feature] WebUI: add backend API interaction error log

moisseev:
Add an error log modal with a responsive table providing:
- tracking of the last 50 errors using a circular buffer
- an "unseen since last view" counter on the badge in bottom-right corner
- copy-to-clipboard support with execCommand fallback for HTTP connections
- color-coded error types
- automatic column hiding on smaller screens


moisseev:
@droid review

PR #5837, opened 2026-01-13
[Fix] Refactor control pipe to prevent deadlocks and crashes

vstakhov:
## Summary

This PR fixes critical issues in the worker-to-main process communication via srv_pipe that could cause:
- SIGABRT crashes on Linux with FORTIFY_SOURCE=2 due to oversized messages
- Worker hangs/deadlocks when multiple commands are queued rapidly (e.g., during hyperscan compilation)
- Heartbeat loss leading to worker restarts

### Commit 1: Reduce control message size
- Remove redundant `cache_dir` fields from `hs_loaded`, `mp_loaded`, `re_map_loaded` command structures
- Cache directory is available from `worker->srv->cfg->hs_cache_dir`
- Reduces `rspamd_srv_command` size from ~8KB to ~500 bytes

### Commit 2: Refactor to queue-based architecture
- Replace per-request ev_io watchers with single watcher using khash for ID-based reply matching
- Add `rspamd_srv_pipe_ctx` with:
  - Single ev_io watcher for srv_pipe[1]
  - khash table for pending requests keyed by command ID
  - Send queue (utlist) for requests waiting to be sent
- Make srv_pipe non-blocking on both ends with proper EAGAIN handling
- Add EAGAIN handling to main process write path
- Fix bug: `r == ENOBUFS` → `errno == ENOBUFS`

## Root Cause

The original design assumed "messages are rare" and used blocking sockets with per-request watchers. The hyperscan refactoring added rapid `rspamd_worker_set_busy()` calls, creating multiple watchers on the same fd. With blocking sockets, when multiple watchers waited for replies, the second watcher's `recvmsg()` would block forever.

## Testing

- All unit tests pass (101 C++ tests, 754 Lua tests)
- Build succeeds on macOS

### DIFF
--- a/src/libserver/rspamd_control.h
+++ b/src/libserver/rspamd_control.h
@@ -63,7 +63,7 @@ enum rspamd_srv_type {
 enum rspamd_log_pipe_type {
 	RSPAMD_LOG_PIPE_SYMBOLS = 0,
 };
-#define CONTROL_PATHLEN MIN(PATH_MAX, PIPE_BUF - sizeof(int) * 2 - sizeof(int64_t) * 2)
+
 struct rspamd_control_command {
 	enum rspamd_control_type type;
 	union {


vstakhov:
@droid review

PR #5838, opened 2026-01-13
[Feature] Add --batch-wait option to dmarc_report

errror:
- Add -w/--batch-wait option to rspamadm dmarc_report to wait n seconds between sending batches. This can be used to avoid overloading SMTP server with too many messages in a short time frame.

### DIFF
--- a/lualib/rspamadm/dmarc_report.lua
+++ b/lualib/rspamadm/dmarc_report.lua
@@ -55,6 +55,11 @@ parser:option "-b --batch-size"
       :argname "<number>"
       :convert(tonumber)
       :default "10"
+parser:option "-w --batch-wait"
+      :description "Wait <batch-wait> seconds between batches"
+      :argname "<number>"
+      :convert(tonumber)
+      :default "0"
 
 parser:flag "-r --recheck-rua"
       :description "Re-check RUA addresses against exclude_rua_addresses map before sending"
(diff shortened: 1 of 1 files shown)


lualib/rspamadm/dmarc_report.lua
```diff
@@ -436,6 +441,10 @@ local function send_reports_by_smtp(opts, reports, finish_cb)
         if args.next_start > #reports then
           finish_cb(reports_sent, reports_failed)
         else
+           if opts.batch_wait then
+              lua_util.debugm(N, 'sleeping %d seconds before next batch', opts.batch_wait)
+              os.execute("sleep " .. opts.batch_wait)
```
vstakhov, line 446: No, please no `os.execute`. Use `rspamadm_ev_base:sleep()`.

errror: Thanks a lot, @vstakhov ! I did not know about this function. Will test and update the pull request.

errror: I just realized, that you added the `sleep()` function on 2026-01-19. It is not yet included in the latest release yet. I will wait for the next release to test and submit an updated pull request.

errror:
Ack.

I will do a new pull request once a new rspamd version including `rspamadm_ev_base:sleep()` was released.

PR #5839, opened 2026-01-13
Allow for use of Lua 5.5

bneumeier:
In Lua 5.5, the signature for lua_newstate changes (there is now an additional "seed" argument),
and the mechanism for adjusting garbage collection also changes.

I don't know whether the seed of 0 is ideal when using lua_newstate; probably, using a random seed would be better. This is a minimal patch that gets back to a working build.


vstakhov:
There are some formatting issues that I can clearly see in the diff. Otherwise it LGTM, thanks!

bneumeier:
Better in 9fc6971?

PR #5840, opened 2026-01-14
[Fix] Use proper rounding for symbol frequency statistics

moisseev:
- Replace incorrect floor() with round() in rounding functions to avoid losing small values
- Increase counters API frequency precision from 3 to 6 decimal places (need 5 to avoid rspamc displaying values as multiples of 0.06, need 6 for /counters endpoint itself - no additional overhead as JSON stores double anyway)
- Add frequency_stddev field to counters API output (fixes zero stdev in `rspamc counters` output)
- Clarify `rspamc counters` table header with "avg (stddev)" subheading
- Fix WebUI to preserve frequency precision before scaling

Example for symbol with frequency 0.004772 hits/sec:
- Before: /symbols returns 0.004772, /counters returns 0.004000, `rspamc counters` shows 0.240
- After:  /symbols returns 0.004772, /counters returns 0.004772, `rspamc counters` shows 0.286

### DIFF
--- a/interface/js/app/symbols.js
+++ b/interface/js/app/symbols.js
@@ -68,7 +68,8 @@ define(["jquery", "app/common", "footable"],
                         item.frequency = 0;
                     }
                     freqs.push(item.frequency);
-                    item.frequency = Number(item.frequency).toFixed(2);
+                    // Don't round yet, keep precision for scaling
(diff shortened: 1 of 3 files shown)


moisseev:
@droid review

PR #5842, opened 2026-01-15
[Fix] ratelimit: fix compatibility with old records

fatalbanana:


vstakhov:
That's good point: when I did that transition, I have assumed that records with old format would just expire quickly. It seems that is not the case for some of our users :)

PR #5846, opened 2026-01-17
[Minor] Fix WebUI symbols frequency column sorting

moisseev:
Previously, frequency values with exponential notation (e.g., "0.00e-5",
"389.40e-5") were compared as strings, causing incorrect sort order.


moisseev:
@droid review

PR #5849, opened 2026-01-19
Add module to save rejected mails for debugging

heptalium:
When installing and configuring a new server some ham mails might be rejected too. And as the mails are rejected there is no possibility to learn them as ham. By enabling the new `save_rejected` module all rejected mails are saved in a directory so `rspamc learn_ham` can be used to learn false positives as ham. In the corresponding merge request in the docs repository is an example for a systemd timer to remove older mails in the directory.

### DIFF
--- a/conf/modules.d/save_rejected.conf
+++ b/conf/modules.d/save_rejected.conf
@@ -1,0 +1,22 @@
+# Please don't modify this file as your changes might be overwritten with
+# the next update.
+#
+# You can modify 'local.d/save_rejected.conf' to add and merge
+# parameters defined inside this section
+#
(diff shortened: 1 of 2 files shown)


vstakhov:
What's wrong with https://docs.rspamd.com/modules/metadata_exporter#settings-send_mail-backend ? Why do we need an extra module?

heptalium:
I didn't know that it is possible with the metadata_exporter. And I don't yet completely understand how to configure it.
On the other hand, the new module addresses a problem that especially people might have, who are new to Rspamd.
And I think for those people it is much easier to enable an additional module then configuring the metadata_exporter and prevent the mailserver from looping back to Rspamd to make it work.

PR #5858, opened 2026-01-23
[Minor] Fix regex flags in parse_sa_regexp function

moisseev:
- Remove invalid flags 'g' and 'u' from regexp pattern that are not supported by PCRE inline modifiers (?flags).
PCRE inline modifiers only support 'imsx' flags.
- Fix duplicate 'i' flag in character class


moisseev:
@droid review

PR #5864, opened 2026-01-27
Add type check for builtin suspicious TLDs

dragoangel:
allow to disable builtin_suspicious wit null in local.d/url_suspect.conf

### DIFF
--- a/src/plugins/lua/url_suspect.lua
+++ b/src/plugins/lua/url_suspect.lua
@@ -408,14 +408,16 @@ function checks.tld_analysis(task, url, cfg)
   end
 
   -- Check built-in suspicious TLDs (5 TLDs, O(n) is fine)
-  for _, suspicious_tld in ipairs(cfg.builtin_suspicious) do
-    if tld == suspicious_tld or tld:sub(-#suspicious_tld) == suspicious_tld then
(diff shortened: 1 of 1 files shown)


dragoangel:
Without this condition we would get:
```
task; lua_metric_symbol_callback: call to (URL_SUSPECT_CHECK) failed (2): /usr/share/rspamd/plugins/url_suspect.lua:411: bad argument #1 to 'ipairs' (table expected, got userdata); trace: [1]:{[C]:-1 - ipairs [C]}; [2]:{/usr/share/rspamd/plugins/url_suspect.lua:411 - tld_analysis [Lua]}; [3]:{/usr/share/rspamd/plugins/url_suspect.lua:589 - analyze_url [Lua]}; [4]:{/usr/share/rspamd/plugins/url_suspect.lua:663 - <unknown> [Lua]};
```

and tbh - I would at all drop `builtin_suspicious` and instead introduce local map that contains this tlds in repo. Why this approach I like more: for map you can define `map = "file"` in local.d and this will override map from defaults, and you can define `map = ["file"]` and it will add your map to default map, at least AFAIK. @vstakhov what you think?

dragoangel:
Thx, forget to close, as proper fix exist

PR #5866, opened 2026-01-29
Enhance fuzzy blacklist handler with richer context

vstakhov:
## Summary
- Enhance `rspamd_fuzzy_maybe_call_blacklisted` to pass richer context to Lua handlers
- Add new vs existing rate limit distinction via `event_type` argument
- Pass bucket state (level, burst, rate, exceeded amount) via `ratelimit_info` table
- Include hash digest when session context is available
- Include extensions (domain, source IP) from fuzzy command

## Lua API

**Before (2 args):**
```lua
rspamd_config:add_fuzzy_blacklist_handler(function(ip, reason)
    rspamd_logger.info("Blocked: " .. tostring(ip))
end)
```

**After (6 args, backwards compatible):**
```lua
rspamd_config:add_fuzzy_blacklist_handler(function(ip, reason, event_type, ratelimit_info, digest, extensions)
    if event_type == "new" then
        rspamd_logger.info("NEW rate limit: " .. tostring(ip))
        if ratelimit_info then
            rspamd_logger.info("  Exceeded by: " .. tostring(ratelimit_info.exceeded_by))
        end
    end
end)
```

## Test plan
- [x] Build succeeds
- [x] Unit tests pass
- [ ] Manual testing with Lua handler logging all arguments

### DIFF
--- a/src/fuzzy_storage.c
+++ b/src/fuzzy_storage.c
@@ -337,6 +337,25 @@ struct rspamd_updates_cbdata {
 	gboolean final;
 };
 
+enum rspamd_ratelimit_event_type {
+	RATELIMIT_EVENT_NEW,
+	RATELIMIT_EVENT_EXISTING,
+	RATELIMIT_EVENT_BLACKLIST,
+};
+
+struct rspamd_ratelimit_callback_ctx {
+	rspamd_inet_addr_t *addr;              /* Client IP */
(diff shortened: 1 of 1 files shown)


vstakhov:
@droid review

vstakhov:
1 and 3 are invalid, backward compatibility is not needed. 2 is true but not very important in fact.

PR #5870, opened 2026-02-02
[Feature] Add mapstats utility for multimap statistics analysis

moisseev:
Introduce a new utility to analyze Rspamd logs and count matches for multimap module patterns. This helps identify ineffective map entries and optimize multimap configurations.

Features:
- Parse file-based multimap configurations from rspamadm configdump
- Load and validate map files with support for:
  * IP/CIDR patterns (IPv4/IPv6)
  * Regular expressions with PCRE flags (imsxurOL)
  * Plain string patterns (domains, hostnames, etc.)
  * Full-line and inline comments
- Process Rspamd logs (plain or compressed: bz2, gz, xz, zst)
- Match log entries against map patterns
- Generate statistics report grouped by map source files
- Show match counts and comments for each pattern
- Report unmatched symbol values for debugging

Installation:
- Added to CMake install rules (installs as 'mapstats')
- Added to RPM spec file
- Debian packaging works automatically via CMake

Usage:
mapstats -l /var/log/rspamd/rspamd.log
mapstats -l /var/log/rspamd -n 5        # Last 5 rotated logs
mapstats -l /var/log/rspamd --start 2024-01-01 --end 2024-01-31

Requirements:
- Perl 5.14+
- JSON::PP (core module)
- NetAddr::IP (optional, for type=ip maps only)

### DIFF
--- a/utils/mapstats.pl
+++ b/utils/mapstats.pl
@@ -247,8 +247,12 @@ exit;
 #-------------
 
 sub configdump {
-    my $cmd  = 'rspamadm configdump -C' . ( defined $_[0] ? " $_[0]" : '' );
-    my $json = `$cmd`;
+    my @cmd = ( 'rspamadm', 'configdump' );
+    push @cmd, '-C', $_[0] if defined $_[0];
+
+    open( my $fh, '-|', @cmd ) or die "Cannot execute rspamadm configdump: $!\n";
+    my $json = do { local $/; <$fh> };
+    close($fh);
 
     # Check command execution status
     if ( $? != 0 ) {
@@ -408,7 +412,7 @@ sub ProcessLog {
             }
 
             foreach my $s (@symbols_search) {
-                my @selected = grep /$s/, @symbols;
+                my @selected = grep /\Q$s\E/, @symbols;
                 next
                   unless ( scalar(@selected) > 0 );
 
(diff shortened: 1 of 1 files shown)


moisseev:
@droid review

moisseev:
**Fixed:**

1: Command Injection in configdump() ✅
•  Changed from backticks with string concatenation to list form open()
•  Eliminates shell interpretation of arguments

3: ReDoS in ProcessLog() ✅
•  Added \Q...\E to escape regex metacharacters in symbol names
•  Prevents catastrophic backtracking from symbol names like SYMBOL+++

5: Regex DoS documentation ✅
•  Added SECURITY CONSIDERATIONS section to POD
•  Documents trust model and limitations

**Not fixed:**

2: system("sleep") - Skipped to maintain sync with sibling script rspamd_stats.pl. Hardcoded string has no injection risk. Can be addressed in future unified refactoring.

4: Path Traversal - This is a diagnostic utility for system administrators. If an attacker controls Rspamd's multimap config, they already have system access. Adding path restrictions
would create false security theater and break legitimate use cases. Documented in SECURITY section.

6: Unvalidated flags - Flags are validated before use (line 345). Error message displaying received input is standard practice.

7: Resource exhaustion - Expected behavior for batch processing utilities (similar to grep, awk, sort). Users control input via -n option.

**Security Model:**

This utility follows UNIX philosophy: trusted input from system administrators. Similar trust assumptions as rspamadm, grep -f, and other diagnostic tools that process configuration files.

moisseev:
@droid review

PR #5871, opened 2026-02-02
fix dynamic bucket-specific rate-limits

JustKiddingCode:
In current rspamd, specifing a dynamic bucket-specific rate-limit is not possible.
This is due to the fact, that the configuration parser does not allow to set the keys in the bucket.
This pull request adds this to the configuration parser and also cleans up the ratelimit code.


vstakhov:
LGTM! @droid review

PR #5892, opened 2026-02-16
[Test] Don't install python libraries at runtime

fatalbanana:
Needs https://github.com/rspamd/rspamd-build-docker/pull/41


vstakhov:
Yeah, I should have done it before merge indeed.

PR #5893, opened 2026-02-16
[Feature] WebUI: Display classifier metadata in Bayesian statistics table

moisseev:
## Summary
Extend `/stat` endpoint and Bayesian statistics table to display classifier metadata (`name`, `type`, `per-user` flag) and statfile classes.

## Changes
- **Backend**: Add `class` field and `classifier` object with `name`, `type` (`binary`/`multi-class`), `per_user` flag to each statfile in `/stat` response
- **Frontend**: Add "Classifier name" and "Class" columns with type badges and rowspan grouping

## Example `/stat` response

**Before:**
```json
{
    "statfiles": [{
        "symbol": "BAYES_SPAM",
        "type": "redis",
        "revision": 28228,
        "users": 1
    }]
}
```

**After:**

```json
{
    "statfiles": [{
        "symbol": "BAYES_SPAM",
        "type": "redis",
        "revision": 28228,
        "users": 1,
        "classifier": {
            "name": "bayes_binary",
            "type": "binary",
            "per_user": false
        },
        "class": "spam"
    }]
}
```

Backward compatible with older Rspamd versions.

image

### DIFF
--- a/src/libstat/stat_process.c
+++ b/src/libstat/stat_process.c
@@ -1895,9 +1895,7 @@ rspamd_stat_statistics(struct rspamd_task *task,
 						ucl_object_frombool(classifier_per_user),
 						"per_user", 0, false);
 
-				if (!ucl_object_insert_key(elt, classifier_obj, "classifier", 0, false)) {
-					ucl_object_unref(classifier_obj);
-				}
+				ucl_object_insert_key(elt, classifier_obj, "classifier", 0, false);
 
 				if (st->stcf->class_name) {
 					ucl_object_insert_key(elt,


moisseev:
@droid review

src/libstat/stat_process.c
```diff
@@ -1780,6 +1863,47 @@ rspamd_stat_statistics(struct rspamd_task *task,
...
+						ucl_object_fromstring(classifier_name),
+						"name", 0, false);
+				ucl_object_insert_key(classifier_obj,
+						ucl_object_fromstring(classifier_type),
+						"type", 0, false);
+				ucl_object_insert_key(classifier_obj,
+						ucl_object_frombool(classifier_per_user),
+						"per_user", 0, false);
+
+				if (!ucl_object_insert_key(elt, classifier_obj, "classifier", 0, false)) {
```
vstakhov: Why do we need this condition here?

src/libstat/stat_process.c
```diff
@@ -1780,6 +1863,47 @@ rspamd_stat_statistics(struct rspamd_task *task,
...
+					while ((cur = ucl_object_iterate(elt, &it, true))) {
+						key = ucl_object_key(cur);
+						/* Skip classifier and class keys if they exist */
+						if (key && strcmp(key, "classifier") != 0 && strcmp(key, "class") != 0) {
+							ucl_object_insert_key(elt_copy, ucl_object_ref(cur), key, 0, true);
+						}
+					}
+				}
+				ucl_object_unref(elt);
+				elt = elt_copy;
```
vstakhov, line 1884: This is very fishy, before this change we were not unrefing the results of `st->backend->get_stat()` and now we actually do. I'm not sure it's safe.

src/libstat/stat_process.c
moisseev: The check prevents memory leak when ucl_object_insert_key() fails:

-  Success (true) → parent object elt takes ownership of classifier_obj and will free it
-  Failure (false) → classifier_obj remains orphaned → must be manually freed with ucl_object_unref() to avoid leak

Without this check, a failed insertion would leak the classifier_obj memory. 

However, should we also check the other insertions for "name", "type", "per_user", and "class"?

vstakhov: We don't need that, as there is no possibility to fail iirc.

src/libstat/stat_process.c
vstakhov: In fact, after investigation, I think this approach is safe.

PR #5894, opened 2026-02-18
[Feature] Multi-flag fuzzy hash support

vstakhov:
## Summary

- Support up to 8 flag slots per fuzzy hash (primary + 7 extra), allowing multiple rules to match against the same digest independently
- Move Redis update path from C MULTI/EXEC with embedded EVAL to Lua architecture following the Bayes pattern: `lualib/redis_scripts/fuzzy_update.lua` + `lualib/lua_fuzzy_redis.lua` functor called from C
- Extend wire protocol with epoch 12 carrying `rspamd_fuzzy_reply_v2` with extra flag entries

## Test plan

- [x] Build with `ninja -j8 install` passes
- [x] C++ unit tests pass (158/158)
- [x] Lua unit tests pass (818/821, 3 unassertive)
- [x] `luacheck` clean on new Lua files
- [x] Integration test: ADD hashes with different flags, verify HGETALL returns multi-slot layout
- [x] Integration test: DELETE and REFRESH with shingles
- [x] Integration test: epoch 12 reply carries extra flags to fuzzy_check plugin
- [x] Verify backward compatibility with epoch 11 clients

### DIFF
--- a/src/fuzzy_storage.c
+++ b/src/fuzzy_storage.c
@@ -1517,12 +1517,12 @@ rspamd_fuzzy_command_valid(struct rspamd_fuzzy_cmd *cmd, int r)
 	case 4:
 		if (cmd->shingles_count > 0) {
 			if (r >= sizeof(struct rspamd_fuzzy_shingle_cmd)) {
-				ret = RSPAMD_FUZZY_EPOCH11;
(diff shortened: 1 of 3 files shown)


vstakhov:
Where is my commit, dear Github?

PR #5898, opened 2026-02-20
[Fix] Fix multiclass Bayes learn cache and classifier isolation

dragoangel:
Propagate task->classifier from the HTTP header so class-targeted learning skips unrelated classifiers early. Validate the requested class exists in a statfile before tokenisation to fail fast with a clear 404 error on misconfigured setups.

Replace numeric class_id hash cache keys in learned_ids with direct class name strings throughout the Redis cache layer (C, Lua, Redis scripts) to fix uint64_t precision loss through Lua's 53-bit doubles, which caused equality checks to always fail for arbitrary multiclass class names.

Add RSPAMD_FLAG_CLASSIFIER_MULTICLASS at config load time to route probability lookups to the correct result type per classifier.

Store multiclass and binary Bayes results under per-classifier mempool keys (multiclass_result:<name>, bayes_prob:<name>) and update the set/get API to take a classifier_name parameter, preventing cross-contamination when multiple classifiers are configured.

Switch multiclass result allocation from heap (g_new0/g_new) to task memory pool, eliminating the need for rspamd_multiclass_result_free.

Inject can_learn_prob and can_learn_class into the mempool before invoking Lua learn conditions, replacing the legacy asymmetric spam_min/ham_max pair with a unified min_prob threshold.

Remove the cl_skipped gate in rspamd_stat_cache_check so the Redis learned_ids cache is always consulted before learn conditions run for proper logging and response handling on relearning same emails.

### DIFF
--- a/src/libstat/learn_cache/redis_cache.cxx
+++ b/src/libstat/learn_cache/redis_cache.cxx
@@ -152,6 +152,46 @@ rspamd_stat_cache_redis_runtime(struct rspamd_task *task,
 	return (void *) ctx;
 }
 
+/*
+ * Map a class name to the string stored in Redis.
+ * ham → "0", spam → "1" (preserves existing cache entries),
+ * anything else (multiclass) → the literal class name.
+ */
+static const char *
+rspamd_stat_cache_class_to_str(const char *class_name)
+{
+	if (class_name == nullptr) {
+		return "0";
+	}
+	if (strcmp(class_name, "spam") == 0) {
+		return "1";
+	}
+	if (strcmp(class_name, "ham") == 0) {
+		return "0";
+	}
+	return class_name;
+}
+
+/*
+ * Reverse of the above: translate a cached Redis string back to a canonical
+ * class name for comparison.
+ * "0" → "ham", "1" → "spam", anything else → returned as-is.
+ */
+static const char *
+rspamd_stat_cache_str_to_class(const char *cache_str)
+{
+	if (cache_str == nullptr) {
+		return nullptr;
+	}
+	if (strcmp(cache_str, "0") == 0) {
+		return "ham";
+	}
+	if (strcmp(cache_str, "1") == 0) {
+		return "spam";
+	}
+	return cache_str;
+}
+
 static int
 rspamd_stat_cache_checked(lua_State *L)
 {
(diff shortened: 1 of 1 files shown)


dragoangel:
changes needed inside docs: `min_prob = 0.95,` instead of 
```
    spam_min = 0.95,
    ham_max = 0.05,
```
spam_min left as fallback* for backward compatibility

Also I think that in general `probability_check` should be not inside `classifier.autolearn.options` - we use this not just in autolearn, @vstakhov right? It used to verify any learning - including manual one, so having it there is quite missleading.

Or maybe I not properly understand `autolearn` section (from tests I do it right)? If we not have `spam_threshold, junk_threshold & ham_threshold` in it - this means there will be no autolearning?

dragoangel:
@vstakhov we also have to mention in docs that it's **required** to use different redis servers (or at least databases - not sure due to bayes expiry can struggle here) for different classifiers, otherwise there will be not enough level of segregation of classifiers between each other.

moisseev:
@droid review

src/libstat/learn_cache/redis_cache.cxx
```diff
@@ -191,7 +164,12 @@ rspamd_stat_cache_checked(lua_State *L)
 	auto res = lua_toboolean(L, 2);
 
 	if (res) {
-		auto val = lua_tointeger(L, 3);
+		/* The cached value is the class name string (e.g. "spam", "ham", "transactional").
+		 * Previously this was stored as a 64-bit integer hash, but uint64_t values
+		 * larger than 2^53 lose precision when passed through Lua doubles, causing
+		 * the equality check to always fail and forcing UNLEARN instead of
+		 * ALREADY_LEARNED for multiclass classifiers. */
+		const char *cached_class = lua_tostring(L, 3);
```
vstakhov: so how compatibility is preserved? From my point of view, this change breaks binary classifiers behaviour - it was 0/1 previously, but now it will be class name. So I guess this approach is incorrect.

dragoangel: From my view it's relatively small one-time change and will not impact much real workloads, so I ignored backward compatibility in this place. If you want - we can detect spam & ham classes and explicitly set them to 0 & 1

vstakhov: Yes, let's do that.

dragoangel: @vstakhov please check latest changes 😊

PR #5902, opened 2026-02-21
[Fix] Properly count rspamd-proxy-milter conns count

dragoangel:
This PR will fix issue in rspamd-proxy-milter:
```
rspamadm control stat -j | jq .total.conns
4294967222
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# <a href="https://rspamd.com"><img src="https://rspamd.com/img/rspamd_logo_black.png" alt="Rspamd" width="220px"/></a>
+# <a href="https://rspamd.com"><img src="https://rspamd.com/rspamd_logo_black.png" alt="Rspamd" width="220px"/></a>
 
 [![GHA](https://github.com/rspamd/rspamd/actions/workflows/ci.yml/badge.svg)](https://github.com/rspamd/rspamd/actions/workflows/ci.yml)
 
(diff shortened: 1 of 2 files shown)


moisseev:
@droid review

vstakhov:
## Analysis of the refcount changes

### Hunk 2 (`client_sock = -1`) — Correct fix

This is a legitimate fix for a real double-close bug. In the milter path, `rspamd_milter_handle_socket(nfd, ...)` takes ownership of the fd, but the session also stores `client_sock = nfd`. When the last session in the chain is destroyed, the destructor both calls `rspamd_milter_session_unref()` (which closes the fd via the milter library) and then `close(session->client_sock)` — a double-close. Setting `client_sock = -1` immediately after handing the fd to the milter library is the right thing to do, and it propagates correctly through `proxy_session_refresh` (which copies `client_sock` to nsession).

### Hunk 1 (`REF_RETAIN`) — Incorrect fix, introduces a memory leak

The added `REF_RETAIN(session)` in the milter finish handler has **no matching `REF_RELEASE`**. Here's the lifecycle:

**Without the PR (current code):**
```
REF_INIT_RETAIN(session)                          → refcount = 1
proxy_backend_master_finish_handler → REF_RELEASE  → refcount = 0 → dtor
```

**With the PR:**
```
REF_INIT_RETAIN(session)                          → refcount = 1
REF_RETAIN(session) [new]                         → refcount = 2
proxy_session_refresh() steals milter_conn, sets client_sock = -1
proxy_backend_master_finish_handler → REF_RELEASE  → refcount = 1
... nobody releases again → session leaked forever
```

After `proxy_session_refresh`, the old session is orphaned with `client_milter_conn = NULL` and `client_sock = -1` — no callbacks are registered, nothing will ever bring the refcount to 0. The comment claiming the matching release is in `proxy_backend_master_finish_handler` / `proxy_backend_master_error_handler` is incorrect — those releases already match the original `REF_INIT_RETAIN`.

This "fixes" the `nconns` underflow only because **leaked sessions are never destroyed**, so `nconns--` in `proxy_session_dtor` is never reached for them. It trades a counter underflow for a memory leak (one full `rspamd_proxy_session` + mempool + mirror_conns per milter message).

### The actual root cause of the `nconns` underflow

The problem is that `worker->nconns++` happens **once** per TCP accept (`proxy_accept_socket`), but `session->worker->nconns--` happens in `proxy_session_dtor` for **every** session destroyed — including all sessions created by `proxy_session_refresh` for multi-message milter connections.

For a milter connection carrying N messages:
```
accept            → nconns++                         (nconns = 1)
msg1 backend done → proxy_session_refresh → old session dtor → nconns--  (nconns = 0)
msg2 backend done → proxy_session_refresh → old session dtor → nconns--  (nconns = -1, underflow!)
...
close             → final session dtor → nconns--    (nconns = -(N-1))
```

The reported value `4294967222` = `(uint32_t)(-74)` is consistent with many multi-message milter connections accumulating extra decrements.

### Suggested fix

Instead of leaking sessions, compensate for the extra `nconns--` in `proxy_session_refresh` by incrementing when creating the replacement session:

```c
// In proxy_session_refresh, after creating nsession:
REF_INIT_RETAIN(nsession, proxy_session_dtor);
nsession->worker->nconns++;  // compensate: old session's dtor will decrement
```

This way each refreshed session pre-increments `nconns` to cancel the old session's destructor decrement, maintaining the invariant of one net increment per TCP connection.

dragoangel:
@vstakhov thanks for highlight 😊, please check updated version of patch

Issue #5904, opened 2026-02-23
[BUG] expand_keys corrupts binary msgpack data in EVALSHA keys ("Missing bytes in input")


Olen:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

Bayes classification and learning intermittently fail with `ERR user_script:1: Missing bytes in input` when Redis Lua scripts attempt to `cmsgpack.unpack()` the token data passed by rspamd. The error is **not caused by corrupted data in Redis** — it reproduces on a completely empty Redis instance with zero Bayes keys.

The error affects both `bayes_classify.lua` (`cmsgpack.unpack(KEYS[3])`) and `bayes_learn.lua` (`cmsgpack.unpack(KEYS[5])`), and occurs on both proxy (milter) and controller workers. The failure rate is approximately 30–40% of all Bayes operations.

### Steps to Reproduce

1. Configure per-user Bayes with `new_schema = true` and Redis backend
2. Train sufficient messages to exceed `min_learns` threshold (I used `min_learns = 10`)
3. Send multiple messages to the trained user via normal mail delivery (proxy/milter path)
4. Observe rspamd log — some messages classify successfully (BAYES_HAM/BAYES_SPAM symbol appears), others fail with "Missing bytes in input"

Alternatively, batch learning via the controller HTTP API shows the same error:
```bash
for msg in /var/mail/example.com/user/cur/*; do
    curl -s -H "Deliver-To: user@example.com" --data-binary "@$msg" http://127.0.0.1:11334/learnham
done
# ~30-40% of requests fail with "Missing bytes in input"
```

**Key observations:**
[...]

PR #5906, opened 2026-02-24
Fix return value for length greater than 65535 in Redis backend

dragoangel:
Found in #5904


dragoangel:
Close as implemented in https://github.com/rspamd/rspamd/commit/30bd75dbcbf56b73010ebfd9f2f341a9e16d7be2

PR #5908, opened 2026-02-24
fix: properly set bayes class labels

dragoangel:
PR #5898 introduced multi-class bayes fixes, but had side effect that statfiles
configured for new `class = "spam\ham";` or old `spam = true\false;` caused the
Redis backend to use the full class name ("spam"/"ham") as the token hash field
name instead of the historical short labels ("S"/"H"). This would significantly 
impact existing Redis training data.

Changes:
- `redis_backend.cxx` (`get_class_label`): always map class names "spam"→"S"
  and "ham"→"H" so that token hashes use the same field names regardless of
  whether the statfile was configured with `spam = true` or `class = "spam"`.
- `bayes_learn.lua`: add defensive S/H normalisation for class labels "spam",
  "ham", "true", "false" before writing token occurrence fields to Redis.
- `bayes_expiry.lua`: normalise "S"→"spam" and "H"→"ham" when reading token
  hash fields via HGETALL so that occurrence distribution statistics and
  class-significance decisions use canonical class names consistently.

After this fix `class = "spam"` / `class = "ham"` is fully interchangeable with
`spam = true` / `spam = false` at the Redis data level — all reads and writes
use the same "S"/"H" field names as before PR #5898 in `RS_<hash>` keys,
and `<sha1>_occurrence_<class_name>` rendered as `spam` and `ham` as it was.

### DIFF
--- a/src/libserver/html/html.cxx
+++ b/src/libserver/html/html.cxx
@@ -1397,6 +1397,7 @@ struct rspamd_html_url_query_cbd {
 	khash_t(rspamd_url_hash) * url_set;
 	struct rspamd_url *url;
 	GPtrArray *part_urls;
+	uint32_t parent_flags; /* Flags from outer URL to propagate */
 };
 
 static gboolean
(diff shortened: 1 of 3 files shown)


moisseev:
@droid review

PR #5910, opened 2026-02-25
[Feature] Fix SIGSEGV when sign_headers is configured as UCL array

Jesssullivan:
Howdy!

I initially worked around this by using colon-delimited string instead of array: sign_headers = "from:to:subject:date:message-id";

The issue, AFAICT:

ucl_object_tostring() returns NULL for UCL_ARRAY type objects, but the DKIM signing code passes this NULL to strlen() without checking, causing a segfault on the first signing attempt.

Note, the UCL parser silently accepts array syntax for sign_headers, and some example configs use this format, making the crash difficult to diagnose without core dumps or ASAN builds.   I am in a bit over my head with ASAN.  

Fixes: SIGSEGV in dkim.c:strlen(headers) when headers is NULL

Following @vstakhov's stuff from November:

- **Config loader** (`dkim_check.c`): New `rspamd_dkim_ucl_sign_headers()` helper detects `UCL_ARRAY` and joins elements with `:` delimiter. Logs deprecation warning.

- **C crash site** (`dkim.c`): NULL guard before `strlen(headers)` — returns descriptive `GError` with `DKIM_SIGERROR_EMPTY_H` instead of crashing.

- **Lua FFI bridge** (`lua_ffi/dkim.lua`): Handles `table` type (UCL arrays arrive as Lua tables) via `table.concat(dkim_headers, ':')`.

### DIFF
--- a/lualib/lua_ffi/dkim.lua
+++ b/lualib/lua_ffi/dkim.lua
@@ -90,6 +90,11 @@ local function create_sign_context(task, privkey, dkim_headers, sign_type)
 
   if not dkim_headers then
     dkim_headers = default_dkim_headers
+  elseif type(dkim_headers) == 'table' then
+    -- UCL arrays arrive as Lua tables; join into colon-delimited string
+    dkim_headers = table.concat(dkim_headers, ':')
+  elseif type(dkim_headers) ~= 'string' then
+    return nil, 'sign_headers must be a colon-delimited string or array, got ' .. type(dkim_headers)
   end
 
   if not sign_type then
--- a/src/libserver/dkim.c
+++ b/src/libserver/dkim.c
@@ -3627,6 +3627,16 @@ rspamd_create_dkim_sign_context(struct rspamd_task *task,
 	nctx->common.is_sign = true;
 
 	if (type != RSPAMD_DKIM_ARC_SEAL) {
+		if (headers == NULL) {
+			g_set_error(err,
+						DKIM_ERROR,
+						DKIM_SIGERROR_EMPTY_H,
+						"sign_headers is NULL (check config: "
+						"use a colon-delimited string, not an array)");
+
+			return NULL;
+		}
+
 		if (!rspamd_dkim_parse_hdrlist_common(&nctx->common, headers,
 											  strlen(headers), true,
 											  err)) {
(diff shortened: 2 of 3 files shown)


vstakhov:
I'm not convinced why do we need this change - sign_headers is expected to be a string. Probably it's not good to crash but my point is simple: shit on input => shit on output. With array support we apparently get nothing useful.

vstakhov:
In fact, something like 
```
sign_headers = {
  "Content-Type" = "oversign";
}
```

is way better:
1) it provides guarantees for header names to be unique
2) it explicitly requires type of signing (e.g. oversign, or ignore, or oversign on absence)

vstakhov:
Ah, and also small note - please don't use `[CritFix]` tag, it's reserved for the category of bugs where normal functioning of Rspamd is compromised and it's like "Release new version and upgrade immediately". A simple `Fix` or even `Feature` (if we go with the object support + type checks) will be way better :)

Jesssullivan:
Title updated, noted 🫡  sorry for the jump scare there. 

On the `sign_headers` being expected as a string — definitely!  The core issue I'm trying to address is working back from the crash.  Your point about the map-based sign_headers (with oversign header) is a much better than array support anyway, glad to drop array handling wholesale if that'd be better.  The motivation here is to get a handle on the crash.  

in my case, mail stack running in kubernetes.  The rspamd config like the rest of the stack is generated via templates.  When I initially wrote the DKIM signing config, I used array syntax because UCL accepts it and it felt natural: sign_headers = ["from", "to", "subject", "date", "message-id"];

 UCL parsed it without complaint --> rspamd starts- only on signing attempt does the SIGSEGV happen.  In Kubernetes, this manifests as a CrashLoopBackOff with no useful diagnostics. I had to pull core dumps from the container to find it was strlen(NULL) at dkim.c:3629.

This boils down to:
- dkim.c: NULL guard before strlen(headers) — return GError with DKIM_SIGERROR_EMPTY_H
  instead of crashing. This should prevent a SIGSEGV regardless of how the bad input arrived.
-  Some kind of config validation: reject non-string types for sign_headers at config load
  time with a error message.

-Jess

Jesssullivan:
Revisiting this;

> 2. it explicitly requires type of signing (e.g. oversign, or ignore, or oversign on absence)

AFAICT `ignore` is not parsed yet (though I am probably missing something 🌁 ) but no matter- setting oversign type still resolves the observed DKIM crash even if the parser accepts the UCL array.

PR #5914, opened 2026-02-27
[Feature] Shard migration and multi-class Bayes in statistics_dump

vstakhov:
## Summary

- Add `rspamadm statistics_dump migrate` subcommand for migrating per-user Bayes data between Redis shards after the Jump Hash → Ketama hash transition (commit `4ea750466`)
- Fix multi-class Bayes support in dump/restore: now handles arbitrary statfile classes (not just binary spam/ham) with proper class label mapping (`class_labels` config, `tbl.class` field)
- Fix dump to iterate all shards via `all_upstreams()` instead of connecting to a single round-robin shard

### Migrate subcommand

```
rspamadm statistics_dump migrate [--dry-run] [--no-delete] [--batch-size N] [-c config]
```

- Scans all Redis shards, identifies misplaced prefix keys via `get_upstream_by_hash()`
- Moves data in batches using Redis Lua scripts (SCAN+HGETALL → msgpack → HMSET)
- `--dry-run`: show what would migrate without writing
- `--no-delete`: copy to target without removing from source
- Idempotent: safe to re-run after interruption

### Multi-class changes

- `check_redis_classifier()` now builds a `symbols` list with `{symbol, class_name, label}` for each statfile
- Class determined by priority: explicit `class` config → `spam` boolean → heuristic from symbol name
- CDB packing uses dynamic class labels instead of hardcoded S/H

## Test plan

- [x] `luacheck` passes clean
- [x] C++ unit tests pass (158/158)
- [x] Lua unit tests pass (818/821, 3 unassertive pre-existing)
- [x] Manual test: `rspamadm statistics_dump dump` with multi-class config
- [x] Manual test: `rspamadm statistics_dump migrate --dry-run` with sharded Redis
- [x] Manual test: full migration on test Redis cluster

### DIFF
--- a/lualib/rspamadm/statistics_dump.lua
+++ b/lualib/rspamadm/statistics_dump.lua
@@ -780,6 +780,14 @@ local function restore_handler(opts)
 
   local batch = {}
   local pending_cmds = 0
+  local total_lines = 0
+  local total_cmds = 0
+  local total_batches = 0
+  local start_time = os.time()
+  local last_report_time = start_time
+
+  rspamd_logger.messagex("starting restore (batch_size=%s, pipeline_max=%s, mode=%s)",
+      opts.batch_size, pipeline_max, opts.mode or 'add')
 
   for _, f in ipairs(files) do
     local fd
@@ -788,6 +796,7 @@ local function restore_handler(opts)
       io.input(fd)
     end
 
+    rspamd_logger.messagex("processing file: %s", f)
     local cur_line = 1
     for line in io.lines() do
       local ucl_parser = ucl.parser()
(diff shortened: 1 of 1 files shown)


dragoangel:
1 user taking ~ 1 minute to process meaning potentially 100k users migration can take up to 70 days, and unfortunately script somewhere stores data on controller which he doesn't cleanup, due to which its consistently takes more RAM usage, most likely at minimum x1.5 (up to x2) of Redis RAM.

My modified version of backup script which collect all keys user-index keys names per-shard in memory and then proccess all keys in shard by batches of 500 keys using x2 of 1 Redis RAM shard on backup too - but just at a peak, on new shard memory is fully cleaned. All my backup job takes in my specific case only 48 minutes to run. Unfortunately restore without significant modifications not runs successfully, it faces:
```
PANIC: unprotected error in call to Lua API (/usr/share/rspamd/lualib/rspamadm//statistics_dump.lua:840: cannot resize stack to fit 0 commands)
```
right after trying to read backup file (~6.5G). To note: backup file taking much less space that actual data used by Redis**, and backup file not compressed even.

vstakhov:
@droid review

PR #5922, opened 2026-03-05
[Minor] Skip empty In-Reply-To header in replies check

moisseev:
An empty `In-Reply-To` header value ("") is truthy in Lua, bypassing the `nil` check. In `replies_check` this caused a misleading log entry "ignoring reply to  as no recipients are matching hash ". In `replies_check_cookie` it triggered an unnecessary `decrypt_cookie` call.


moisseev:
@droid review

PR #5927, opened 2026-03-07
[Fix] Skip recipient check when no hash found in Redis

moisseev:
When a key is not found in Redis, lua_redis returns a redis.null userdata (not nil), which is truthy and caused check_recipient() to be called unconditionally, logging a misleading "no recipients are matching hash" message despite no hash being stored.


moisseev:
@droid review

PR #5937, opened 2026-03-14
feat: allow utilize own rspamd mirror via env variable

dragoangel:


vstakhov:
Very nice, I like it!

PR #5938, opened 2026-03-15
[Feature] Add missing Jinja2 filters and tests to Lupa

vstakhov:
## Summary
- Add 10 new filters to Lupa template engine: `split`, `trim`, `tojson` (via UCL), `unique`, `min`, `max`, `items`, `keys`, `values`, `wordwrap`
- Add 15 new tests: `is_in` (string/table), `is_startswith`, `is_endswith`, `is_match` (Lua patterns), `is_boolean`, `is_true`, `is_false`, `is_integer`, `is_float`, `is_eq`, `is_ne`, `is_lt`, `is_le`, `is_gt`, `is_ge`
- Closes the gap with standard Jinja2 built-in filters and tests

## Test plan
- [ ] Build succeeds with modified lupa.lua
- [ ] Existing Lua tests pass (`rspamd-test -p /rspamd/lua`)
- [ ] Verify `split` filter with separator and without (whitespace split)
- [ ] Verify `tojson` filter produces valid JSON via UCL
- [ ] Verify `is_in` works for both substring matching and table element lookup
- [ ] Verify `is_match` works with Lua patterns

### DIFF
--- a/contrib/lua-lupa/lupa.lua
+++ b/contrib/lua-lupa/lupa.lua
@@ -1146,6 +1146,18 @@ function M.filters.int(value)
 end
 
 ---
+-- Returns a list of {key, value} pairs from table *t*.
+-- @param t The table to get items from.
+-- @usage expand('{% for k, v in {a=1, b=2}|items %}{{ k }}={{ v }} {% endfor %}')
+-- @name filters.items
+function M.filters.items(t)
+  assert(t, 'input to filter "items" was nil instead of a table')
+  local result = {}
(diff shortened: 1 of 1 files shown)


dragoangel:
is so cool, if we would have add `is_json` check and `tojson(string,0 for compact and 2-4 etc for intent)` - then we can fully parse `env.VAR={"var":["1","2",..]}` and other complex cases, question only with `[]` lists, as `tojson` need to understand that `["1","2"]` is not `{'1','2'}`

PR #5944, opened 2026-03-17
[Fix] Register redis and http as known hs_helper worker options

dragoangel:
The redis and http configuration blocks in the hs_helper worker section were not registered via rspamd_rcl_register_worker_option, causing rspamadm configdump to emit "unknown worker attribute: redis" warnings. The Lua backend reads these blocks at runtime through the full UCL options object, so they worked correctly despite not being registered.

Add proper RCL registration for both redis and http as ucl_object_t fields so the config schema recognizes them as valid worker attributes.


vstakhov:
Interesting that we even don't use those thingies.

PR #5952, opened 2026-03-25
[Fix] Self-healing hyperscan cache

vstakhov:
## Summary

- When workers fail to deserialize a cached hyperscan blob (version mismatch, corruption), they now **delete the stale cache entry** and **send a recompile request** to hs_helper
- Added `rspamd_hs_cache_lua_delete_async()` C wrapper for the Lua backend's existing `delete` method (file/redis/http)
- New `RSPAMD_SRV_RECOMPILE_REQUEST` control command: worker → main → broadcast `RSPAMD_CONTROL_RECOMPILE` → hs_helper
- 5-second debounce in `rspamd_hs_helper_reload` to prevent thrashing from multiple workers
- Removed dead `needs_recompile` flag (was set on worker's `re_class` but checked in hs_helper — different process, never worked)

Previously the system would get permanently stuck in PCRE fallback because hs_helper's `exists` check would pass for stale entries and skip recompilation.

## Test plan

- [x] Build succeeds (`ninja -j8 install`)
- [x] C++ tests pass (170/170)
- [x] Lua tests pass (948/951, 3 unassertive, 0 failed)
- [ ] Manual: corrupt an hs cache file, verify logs show deletion + recompile + successful reload

### DIFF
--- a/src/hs_helper.c
+++ b/src/hs_helper.c
@@ -249,6 +249,7 @@ struct hs_helper_ctx {
 	ucl_object_t *redis_config; /* Redis backend configuration */
 	ucl_object_t *http_config;  /* HTTP backend configuration */
 	int lua_backend_ref;        /* Lua reference to backend object */
+	ev_tstamp last_recompile_time; /* Timestamp of last recompile start for debounce */
 };
 
 /* Parse cache_backend string to enum */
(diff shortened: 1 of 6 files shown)


vstakhov:
@droid review

PR #5956, opened 2026-03-28
[Feature] Split attachment filenames into sub-tokens for Bayes classifier

moisseev:
Instead of a single whole-filename token, generate a structured set:
- #f:  original filename (case-preserved, backward-compatible)
- #fe: last extension + double extension when present (e.g. #fe:gz, #fe:tar.gz)
- #fp: meaningful parts split by delimiters, CamelCase and letter/digit
    boundaries; purely numeric and single-character parts discarded

Improves generalization for emails with similar attachment names differing
only in dates, IDs or other numeric suffixes, and provides a strong signal
for malware disguise patterns (e.g. pdf.exe, doc.vbs).

### DIFF
--- a/lualib/lua_stat.lua
+++ b/lualib/lua_stat.lua
@@ -927,6 +927,8 @@ local function get_stat_tokens(task, cf)
   return res
 end
 
+exports._filename_to_tokens = filename_to_tokens
+
 exports.gen_stat_tokens = function(cfg)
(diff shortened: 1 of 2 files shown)


moisseev:
@droid review

vstakhov:
I do agree, that's very nice idea, thank you!

PR #5958, opened 2026-03-30
[Fix] Skip HELO IP address literals for SPF domain resolution

moisseev:
Per RFC 7208 §2.3, IP address literals (e.g. [192.0.2.1]) cannot be used as SPF domains. When MAIL FROM is empty, rspamd falls back to HELO as the SPF domain.

With an IP literal in HELO, rspamd_spf_resolve() failed (DNS name rejected by the resolver), but spf_cred was non-NULL, so the result was incorrectly set to R_SPF_DNSFAIL instead of R_SPF_NA.

Skip building spf_cred from HELO when it starts with '[', so rspamd_spf_get_cred() returns NULL and the result is R_SPF_NA.


moisseev:
@droid review

PR #5977, opened 2026-04-10
[Fix] Separate greylisting period from Redis connection timeout in gr…

moisseev:
…eylist

settings.timeout (greylisting period, 5 min) was being picked up by lua_redis as the Redis connection timeout, inflating the symbol's augmentation timeout to 300s. Add redis_timeout (default 1.0s) and explicitly set redis_params.timeout after parse_redis_server.

### DIFF
--- a/conf/modules.d/greylist.conf
+++ b/conf/modules.d/greylist.conf
@@ -20,7 +20,8 @@ greylist {
   ];
 
   expire = 1d; # 1 day by default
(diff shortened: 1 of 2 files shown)


moisseev:
@droid review

moisseev:
@droid review

moisseev:
@droid review

PR #5978, opened 2026-04-10
[Fix] Warn on task_timeout less than symcache symbol timeout

moisseev:
Call rspamd_worker_check_and_adjust_timeout during configtest so misconfigured plugin timeouts are reported at configuration validation time. Elevate the diagnostic from info to warning level.

Fix rspamd_symcache_add_symbol_augmentation to parse "key=value" format in augmentation strings, allowing numeric timeout augmentations from Lua plugins to be stored and compared correctly.

Clarify mx_check timeout comment to explain that the effective symbol timeout includes dns.timeout in addition to the configured value.

### DIFF
--- a/conf/modules.d/mx_check.conf
+++ b/conf/modules.d/mx_check.conf
@@ -21,7 +21,9 @@
 # You also need to define redis servers for this module
 
 mx_check {
-  # connection timeout in seconds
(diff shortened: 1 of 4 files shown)


moisseev:
@droid review

moisseev:
@droid review

PR #5984, opened 2026-04-13
[Fix] Skip ICU conversion for x-binaryenc charset in all detection paths

moisseev:
x-binaryenc is a synthetic name returned by CED (Google Compact Encoding Detection) to signal binary content, not a real text encoding.  ICU has no converter for it and always fails with U_FILE_ACCESS_ERROR, producing a misleading warning in the logs.

Add an early exit in rspamd_mime_text_part_maybe_convert() for all three detection paths: announced-charset missing, announced-charset unknown, and rspamd_mime_charset_utf_check content-heuristic. When the detected charset is x-binaryenc, mark the part as raw binary immediately, consistent with what the existing fallback already does.

Extract set_part_binary() helper and RSPAMD_BINARYENC_CHARSET constant to eliminate the resulting code duplication.

Fixes the spurious "cannot open converter for x-binaryenc" warning seen when processing messages with binary MIME parts.

### DIFF
--- a/src/libmime/mime_encoding.c
+++ b/src/libmime/mime_encoding.c
@@ -31,6 +31,7 @@
 #include <math.h>
 
 #define UTF8_CHARSET "UTF-8"
+#define RSPAMD_BINARYENC_CHARSET "x-binaryenc"
 
 #define RSPAMD_CHARSET_FLAG_UTF (1 << 0)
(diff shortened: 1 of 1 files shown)


moisseev:
@droid review

moisseev:
@droid review

PR #5985, opened 2026-04-14
[Feature] Add --batch-wait option to dmarc_report

errror:
Hi @vstakhov ,
This is an updated version of my previous pull request #5838 . This one does not use `os.execute` any more but now `rspamadm_ev_base:sleep()` as suggested by you.

Sorry for the long delay. I wanted to test my code which was not possible until the latest release.

Feel free to squash my commits into one. If you have questions, feel free to ask me here or on SLAC26 in Berlin.

Best regards,
Patrick


lualib/rspamadm/dmarc_report.lua
```diff
@@ -436,6 +441,10 @@ local function send_reports_by_smtp(opts, reports, finish_cb)
         if args.next_start > #reports then
           finish_cb(reports_sent, reports_failed)
         else
+           if opts.batch_wait then
```
vstakhov: `0` is true in case of `lua`, so the correct thing is to compare with 0, like `if opts.batch_wait > 0 then`

lualib/rspamadm/dmarc_report.lua
```diff
@@ -675,6 +684,11 @@ local function process_report_date(opts, start_time, end_time, date)
 
     -- Force garbage collection after each batch to release Redis connections
     collectgarbage("collect")
+
+    if opts.batch_wait then
```
vstakhov: Same

lualib/rspamadm/dmarc_report.lua
vstakhov: We can also skip last wait and we should do it: 
`if batch_end < #results and opts.batch_wait > 0 then`

vstakhov (changes requested):
Please also fix indentation - it's 3 spaces now not 2

lualib/rspamadm/dmarc_report.lua
errror: > We can also skip last wait and we should do it: `if batch_end < #results and opts.batch_wait > 0 then`

Agreed, but only for the second case. In 436, this is already handled with `if args.next_start > #reports then ... else`.

errror:
Hi @vstakhov ,

thanks very much for your suggestions, I updated the PR accordingly (and learned some new about lua ;-) ).

PR #5991, opened 2026-04-21
[Test] Test saving of DMARC reports

fatalbanana:
- Provokes trouble with PUC-Rio Lua


vstakhov:
Yeah, we should probably fix the compatibility with plain Lua in that place.

PR #6002, opened 2026-04-26
[Minor] Log map's description and improve empty static maps handling

moisseev:
This pull request improves how empty static maps are handled in the Lua map infrastructure, particularly to avoid unnecessary error logs and clarify error messages when maps are intentionally empty or misconfigured. The most important changes are grouped below:

**Error message clarity in C map code:**

* In `map.c`, error messages for map loading failures now include the map's description, making it easier to identify which map is misconfigured. This applies to various failure cases, such as empty URL lists, missing elements, empty object lists, no valid backends, or invalid value types.

**Lua map handling improvements:**

* In `lua_maps.lua`, when an empty table is provided for a map (such as a default whitelist), the code now creates a static empty map without invoking the C map infrastructure, preventing spurious error logs. A warning is still logged for visibility.
* The default whitelist in `lua_cta.lua` is now omitted entirely if empty, relying on the new static empty map logic rather than always creating a map from an empty table.

### DIFF
--- a/lualib/lua_cta.lua
+++ b/lualib/lua_cta.lua
@@ -70,10 +70,6 @@ local DEFAULT_STOPWORDS = {
   "mailerlite", "mailerq", "mailrelay", "mailup", "omnisend", "clickdimensions", "dotdigital", "pepipost"
 }
 
-local DEFAULT_WHITELIST = {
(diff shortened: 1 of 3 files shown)


moisseev:
@droid review

Issue #6007, opened 2026-04-28
[BUG] INVALID_MSGID false positive for valid internationalized Message-IDs when enable_mime_utf8 is enabled


NickOvt:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
When a message arrives with a UTF-8 local part in the envelope sender (SMTPUTF8) and `enable_mime_utf8 = true` is set in rspamd options, the `INVALID_MSGID` symbol is triggered on the Message-ID header even though the Message-ID is valid for an SMTPUTF8 message.

The Message-ID RHS (right-hand side domain) contains a valid internationalized domain (e.g. `wildduck.äää.test`) which is legitimate in an EAI/SMTPUTF8 context, but rspamd's MSGID validation appears to apply strict ASCII-only rules regardless of whether SMTPUTF8 is in use.

Additionally, `enable_mime_utf8` is completely undocumented in the official rspamd documentation (or I'm just blind). The only reference to this option was found by searching through pull requests in the rspamd repository. This makes it very difficult to configure rspamd correctly for SMTPUTF8/EAI deployments. The official docs should mention this option explicitly under the options configuration page.

### Steps to Reproduce

1. Run rspamd (latest Docker image `rspamd/rspamd`) with `enable_mime_utf8 = true` in `/etc/rspamd/local.d/options.inc`
2. Send an email via SMTP with `SMTPUTF8` extension where:
- Envelope sender has a UTF-8 local part, e.g. `йокйцалцй@wildduck.äää.test`
- Message-ID header domain is an internationalized domain, e.g. `<49faaad5-3d90-2713-583b-6cb8a5d06345@wildduck.äää.test>`
[...]

PR #6011, opened 2026-04-29
[Fix] Honor mime_utf8 option in INVALID_MSGID rule

moisseev:
Two related issues caused INVALID_MSGID false positives on valid EAI/SMTPUTF8 Message-IDs (RFC 6532):

* The sane_msgid regexp unconditionally rejected bytes \x80-\xff, even when mime_utf8 was enabled. Relax the regexp in that case while keeping structural checks intact.

* The configuration option was registered only as enable_mime_utf, but the corresponding Lua API is rspamd_config:is_mime_utf8(), so users naturally try enable_mime_utf8. That spelling silently had no effect because the parser did not bind it to any field. Register enable_mime_utf8 as an alias mapped to the same struct field so configs using it actually take effect.

Add a functional test (configs/mid_utf8.conf, messages/mid_eai_utf8.eml, cases/107_mid_utf8.robot) that exercises both fixes via the new option name and verifies that structurally invalid Message-IDs are still flagged.

Issue #6007

### DIFF
--- a/rules/regexp/headers.lua
+++ b/rules/regexp/headers.lua
@@ -533,7 +533,13 @@ reconf['FORGED_MUA_POSTBOX_MSGID_UNKNOWN'] = {
 }
 
 -- Message id validity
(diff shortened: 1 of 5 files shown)


moisseev:
@droid review

PR #6014, opened 2026-05-01
[Feature] Add chain-aware cache and intermediate hop injection in url_redirector, improve timeouts handling

dragoangel:
### Goal
Catch cloaker/rotator phish chains where a malicious intermediate hop
serves a benign tail on second visit. Without this, the module saw only
orig and final, so the malicious middle URL was invisible to
phishing/SURBL/regexp modules and the cached "tail =
innocent.com" silently confirmed a benign verdict for everyone after.

### Changes

* Inject every saved intermediate hop into the task, gated by new
  setting `save_intermediate_redirs = { redirectors, non_redirectors }`.
  Default `{ redirectors=false, non_redirectors=true } -- redirector
  chains are noise; non-redirector intermediates are where rotator
  infrastructure surfaces.` For users who not explicitly defined
  `redirectors_only = false` behavior of module will not change in general.
* Cache layout becomes one Redis entry per adjacent pair: hash(prev)
  -> next_url with markers '^hop:' (intermediate, walk further),
  '^nested:' (gave up, can extend later) or no marker (terminal).
  Reader walks via successive GETs until terminal or dead end.
* On '^nested:' cache hit, hand off to a fresh live HTTP walk with
  full nested_limit budget. If extension finalizes, the upstream
  '^nested' marker is overwritten with '^hop:' -- chain self-heals
  across scans, growing up to nested_limit hops per scan.
* Cache walks are unbounded (Redis is cheap); only HTTP consumes
  nested_limit. Cycle protection is a per-walk seen-set keyed by URL
  string.
* Symbol option for redirector_symbol shows the full host path
  (host1->host2->...->hostN), not just orig->final.
* ZINCRBY on top_urls uses the canonical URL string with no markers.

Split timeouts so the symbol budget isn't read as a per-call timeout:

* timeout: total budget for URL_REDIRECTOR_CHECK augmentation (8s).
* http_timeout: per-HEAD timeout, number or table with connect/ssl/
  write/read components (4s).
* redis_timeout: per-Redis-call timeout (2s); applied via redis_opts
  swap before parse_redis_server, with the nested redis{} block
  handled too.
* Reservation lock TTL on hash(orig) is settings.timeout -- worst
  case time the symbol could legitimately hold the marker.

### DIFF
--- a/conf/modules.d/url_redirector.conf
+++ b/conf/modules.d/url_redirector.conf
@@ -24,10 +24,10 @@ url_redirector {
   #   read_timeout = 3s;
   # };
   redis_timeout = 2s; # redis timeout for cache operations (redis.conf module has higher priority)
-  nested_limit = 5; # how many redirects to follow
+  nested_limit = 2; # how many redirects to follow
   key_prefix = "rdr:"; # default hash name
   check_ssl = false; # check ssl certificates
-  max_urls = 5; # how many urls to check (СTA checked in first place)
+  max_urls = 5; # how many urls to check (CTA checked in first place)
   max_size = 10k; # maximum body to process
   redirectors_only = true; # follow only known redirectors
   top_urls_key = "rdr:top_urls"; # key for top urls
--- a/conf/modules.d/url_redirector.conf
+++ b/conf/modules.d/url_redirector.conf
@@ -14,11 +14,28 @@
 
 url_redirector {
   expire = 1d; # 1 day by default
-  timeout = 10; # 10 seconds by default
-  nested_limit = 1; # How many redirects to follow
+  timeout = 8s; # total timeout of module
+  http_timeout = 4s; # HTTP HEAD timeout per redirect hop Either a number (whole-request duration)
+  # or a table with .connect_timeout, .ssl_timeout, .write_timeout, .read_timeout for granular control.
+  # http_timeout = {
+  #   connect_timeout = 2s;
+  #   ssl_timeout = 2s;
+  #   write_timeout = 1s;
+  #   read_timeout = 3s;
+  # };
+  redis_timeout = 2s; # redis timeout for cache operations (redis.conf module has higher priority)
+  nested_limit = 5; # how many redirects to follow
   key_prefix = "rdr:"; # default hash name
   check_ssl = false; # check ssl certificates
+  max_urls = 5; # how many urls to check (СTA checked in first place)
   max_size = 10k; # maximum body to process
+  redirectors_only = true; # follow only known redirectors
+  top_urls_key = "rdr:top_urls"; # key for top urls
+  top_urls_count = 200; # how many top urls to save
+  save_intermediate_redirs = {
+    redirectors = false;
+    non_redirectors = true; # inject non-redirector hops by default since they can hide cloaker phishing urls
+  }
 
   .include(try=true,priority=5) "${DBDIR}/dynamic/url_redirector.conf"
   .include(try=true,priority=1,duplicate=merge) "$LOCAL_CONFDIR/local.d/url_redirector.conf"
--- a/src/plugins/lua/url_redirector.lua
+++ b/src/plugins/lua/url_redirector.lua
@@ -297,8 +297,12 @@ local http_walk
 -- hash(chain[#chain]); pass nil to issue the GET. seen is the shared
 -- per-scan URL-string set (cache walk and http_walk write to it so
 -- cycles crossing both layers are caught with one extra Redis GET at
--- worst).
-step = function(task, orig_url, chain, seen, data)
+-- worst). ntries is the count of HTTP HEADs already issued in this
+-- scan -- threaded through cache hops without change so the
+-- ^nested-bridge below hands http_walk the correct remaining budget,
+-- not a fresh nested_limit. Defaults to 0 for top-level cache walks.
+step = function(task, orig_url, chain, seen, data, ntries)
+  ntries = ntries or 0
   if data == nil then
     local last = chain[#chain]
     local next_key = cache_key_for_url(tostring(last))
@@ -310,7 +314,7 @@ step = function(task, orig_url, chain, seen, data)
             apply_redirect_chain(task, chain)
             return
           end
-          step(task, orig_url, chain, seen, d)
+          step(task, orig_url, chain, seen, d, ntries)
         end,
         'GET', { next_key })
     if not ret then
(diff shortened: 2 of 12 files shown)


vstakhov:
## Review

Thanks for the work — the chain-aware cache design is reasonable and the code is well-commented. A few items below; nothing here is a safety blocker, but several should land before merge.

### Compliance

- Title format `[Feature] …` ✓
- GPG-signed ✓
- `luacheck` clean ✓
- 2-space indentation ✓
- No new tests added ✗

### Behavior changes worth calling out in the PR description

The PR bundles several default-config / API-surface changes that aren't mentioned in the body:

1. `conf/modules.d/url_redirector.conf` raises `nested_limit` from `1` → `5` — 5× expansion of follow depth for anyone running the shipped config.
2. Default `timeout` cut from `10` → `8`.
3. `redirector_symbol` option format changes from `host(orig)->host(final)` to `host1->host2->…->hostN`. Any consumer parsing the option string will break.
4. `URL_REDIRECTOR_NESTED` option format also changes (loses the `:cached` suffix; see #2 in bugs).
5. `set_redirected` is now called on **every adjacent pair** rather than only `orig → final`, flagging every intermediate URL with `RSPAMD_URL_FLAG_REDIRECTED`. Other modules keying off that flag will see more URLs than before.
6. `ZINCRBY` to `top_urls_key` now fires per-link, not per-chain — `top_urls` now mixes popular tails with popular intermediates.

### Real bugs / regressions

**1. Debug log lost for the string-UA path.** The diff moves `lua_util.debugm(N, task, 'select user agent %s', ua)` inside the `else` branch:

```lua
if type(settings.user_agent) == 'string' then
  ua = settings.user_agent
else
  ua = settings.user_agent[math.random(#settings.user_agent)]
  lua_util.debugm(N, task, 'select user agent %s', ua)
end
```

Before, it fired regardless. Move the call back out of `else`.

**2. `URL_REDIRECTOR_NESTED` no longer fires on cached `^nested` hits.** Old code emitted the symbol immediately on cache hit. New `step()` for `prefix == 'nested'` jumps straight into `http_walk` to extend the chain without raising the symbol first. The symbol now only fires if the live extension *also* exhausts `nested_limit`. If that's deliberate, please document the new semantics — but I'd lean toward emitting the symbol once on encounter, since a chain parked at `^nested` for the whole `expire` window is otherwise silently downgraded.

**3. Cyrillic letter in user-facing comment.** Both `conf/modules.d/url_redirector.conf` and `src/plugins/lua/url_redirector.lua`:

```
max_urls = 5; # how many urls to check (СTA checked in first place)
```

That `С` is U+0421 CYRILLIC CAPITAL LETTER ES, not Latin `C`. Fix to Latin and consider rewording to `(CTA URLs checked first)`.

**4. Lock TTL margin gone.** Old: `floor(timeout * 2)` = 20s. New: `floor(settings.timeout)` = 8s — exactly the augmentation timeout that bounds the symbol. If the symbol runs to its full budget, the lock can expire just as the symbol finishes, allowing a duplicate live walk by the next scan. Not catastrophic, but the old `* 2` margin handled this. Suggest `floor(settings.timeout) + 1` or restore `* 2`.

**5. `apply_http_timeout` silently drops unknown keys.** Only copies the four granular timeouts. `http_timeout = { timeout = 4 }` (a likely typo since the scalar form uses `timeout`) is silently ignored and falls back to defaults. Either warn on unrecognized keys, or accept `timeout` in the table form as a whole-request fallback.

**6. `chain_append` dedups only against the tail.** The cache walk has a per-walk `seen[val]` set; `http_walk` does not. With `redirectors_only = false`, a redirector loop A→B→A→B is bounded only by `nested_limit` and bloats the chain (and cache) with `nested_limit` entries of a 2-cycle. A 1-line `seen[tostring(url)]` guard inside `http_walk` would mirror the cache walk.

### Test coverage

`test/functional/cases/162_url_redirector.robot` only exercises a 2-hop chain and one cached re-scan. None of the new logic is covered:

- `^hop` vs `^nested` markers and the cache-walk algorithm
- Intermediate-hop injection (`task:inject_url` for non-orig, non-final hops)
- `save_intermediate_redirs.{redirectors,non_redirectors}` gating
- Self-healing path: `^nested` → live extension → upstream link rewritten as `^hop`
- New `redis_timeout` / `http_timeout` table form
- Chain-host symbol option format

`dummy_http.py` already supports `/redirect2` → `/redirect1` → `/hello`, so a 3-hop test is cheap. Given how many behavior surfaces this touches, I'd ask for at least one new case asserting all intermediate URLs are injected and a re-scan reproduces the chain from cache.

### Style nits

- `step(local_chain, data)` overloads two distinct entry points (issue GET when `data == nil`, process when not). Splitting into `walk_step(chain)` / `process_step(chain, data)` would help.
- Some comments lost their leading capitalization in the rewrite (`# how many redirects to follow`, etc.) — minor inconsistency with surrounding `# 1 day by default`.
- The length-1 self-loop write in `cache_chain_to_redis` is subtle; the cache reader interaction (the `seen[val]` cycle check kicks in on the second hop) deserves a sibling comment in `step()`.

### Asks before merge

1. Add a functional test covering chain caching + intermediate-hop injection.
2. Fix the Cyrillic `С`.
3. Restore the `select user agent` debugm for the string-UA case.
4. Clarify or fix `URL_REDIRECTOR_NESTED` semantics on cached `^nested` (bug #2).
5. Document the behavior changes (nested_limit default, symbol option formats, per-hop ZINCRBY, intermediate `set_redirected`) in the PR body.
6. Optional: `seen` guard in `http_walk`, restore lock TTL margin.

dragoangel:
Hi @vstakhov,
regarding comments:
1. code vs conf had misalignment - code said 5, conf said 1, I think default rspamd settings are already quite strict - it walk only via known redirectors and limiting it to 1 sounds too strict, so what was set in lua sounds more correct
2. default timeout change mentioned in code, and yes 8 seconds fills reasonable, on practice is rarely extends 2-3s even with 5 redirects
3. redirector symbol will have `host1->host2->…->hostN` and that's the point, but with default settings it will be applied only for non-redirector hosts, which not enabled by default, and user can disable this logic by defining `save_intermediate_redirs.non_redirectors=false` and he get exactly - `host(orig)->host(final)` as it was
4. `URL_REDIRECTOR_NESTED` yes - because it will not be cached, I improved logic of how nested limit works - now limit applies only to `HTTP HEAD` requests in scope of 1 task, not to cache requests, because redis is cheap and we can easilly go more then that and see loops if any too. It means that on 1st scan you can face nested limit, but on second scan when you hit cache - you will need just 1 HEAD to really finish resolving to end - and you not get cached nested, but actually do that HEAD and get proper resolved end url - I think this is good improvement?
5. 5 & 6 yes and this is intentional from my view: I think it useful to see in top urls not only `innocent.com` but also `evil.com` & `evil.org` & `evil.net` which all are intermediate links that are on X visit start redirecting to `innocent.com` instead of displaying evil page.

regarding bugs:
1. debug log on string (the only 1 existing) UA with wording `select user agent` sounds wrong, it clear if that the only UA set - it was used, so I moved that debug inside condition on purpose
2. `URL_REDIRECTOR_NESTED` no longer fires on cached `^nested` hits - this aligned with points explained in comments N4.
3. yeah, will fix
4. if task reached timeout - nobody will fill that symbol, I can add `+ 1` second to timeout for TTL lock, but `* 2` sounds way too much, what you think?
5. `apply_http_timeout` silently drops unknown keys because otherwise it would allow override non-timeout related settings which should not be there. `http_timeout = { timeout = 4 }` is not allowed and comment section says that format is `http_timeout = 4` (which is default for non glanular settings) or `http_timeout = { connect_timeout = 1, ssl_timeout = 2, ... }`.
6. chain_append dedups only against the tail. The cache walk has a per-walk seen[val] set; http_walk does not - I will double-check that.

dragoangel:
@vstakhov I think I will also add check of cache on 30x inside of `http_walk` - it would add extra 1 redis get, but can minimize need of another `http_walk` when intermediate link is already cached, f.e.:
```
1. short.ly/1 -> short.ty/1 -> dst.com - frist query - no cache
2. short.ly/2 -> short.ty/1 -> dst.com - only 1st HEAD will be real, second will be get from cache
```

PR #6017, opened 2026-05-04
[Fix] url_redirector: cache write missing on splice, brittle mid-walk

dragoangel:
Fix cache write missing on splice, brittle mid-walk handling, silent stale-lock drops

- Persist live-resolved chain when http_walk splices into step (terminal path now calls finalize_chain instead of apply_redirect_chain only). Previously the new link from orig_url to a cached redir_url was never written, leaving 'processing' marker at hash(orig_url) until 13s expiry
- Resume via http_walk on true cache miss mid-walk instead of giving up with a truncated chain (covers expired/evicted downstream links in cached chains)
- Differentiate cache miss from 'processing' lock mid-walk: lock means another worker is resolving (apply partial), miss means cache gone (extend via HTTP)
- Surface redis errors during chain walk via dedicated rspamd_logger.errx
- Add debug log when SET NX lock claim fails (held by another worker or stale processing marker after crash); previously it was a silent drop
- Add debug log 'no URLs matched redirector_hosts_map' at handler exit when message had URLs but selected=0, exposing cold-start window where redirector_hosts_map multimap has not finished loading

### DIFF
--- a/src/plugins/lua/url_redirector.lua
+++ b/src/plugins/lua/url_redirector.lua
@@ -293,6 +293,17 @@ local redirection_codes = {
 local step
 local http_walk
 
+-- Terminal exit for step(): write back if we extended via HTTP this scan,
+-- else just apply. Hoisted as a free function so step()'s recursive cache
+-- hops don't allocate a fresh closure per call.
+local function step_finish(task, chain, http_extended)
+  if http_extended then
+    finalize_chain(task, chain, nil)
+  else
+    apply_redirect_chain(task, chain)
+  end
+end
+
 -- Walk a cached redirect chain. data is the Redis value for
 -- hash(chain[#chain]); pass nil to issue the GET. seen is the shared
 -- per-scan URL-string set (cache walk and http_walk write to it so
@@ -313,39 +324,31 @@ step = function(task, orig_url, chain, seen, data, ntries, http_extended)
   ntries = ntries or 0
   http_extended = http_extended or false
 
-  -- Exit helper: write back if we extended via HTTP this scan, else just apply.
-  local function finish()
-    if http_extended then
-      finalize_chain(task, chain, nil)
-    else
-      apply_redirect_chain(task, chain)
-    end
-  end
-
   if data == nil then
     local last = chain[#chain]
-    local next_key = cache_key_for_url(tostring(last))
+    local last_str = tostring(last)
+    local next_key = cache_key_for_url(last_str)
     local ret = lua_redis.redis_make_request(task,
         redis_params, next_key, false,
         function(e, d)
           if e then
             rspamd_logger.errx(task,
-                'redis error during chain walk at %s: %s', last, e)
-            finish()
+                'redis error during chain walk at %s: %s', last_str, e)
+            step_finish(task, chain, http_extended)
           elseif d == 'processing' then
             -- Another worker is currently resolving this hop; their write
             -- will populate the cache when they finish. Apply what we have
             -- and don't duplicate their HTTP work.
             lua_util.debugm(N, task,
-                'cache lock at %s mid-walk, applying partial chain', last)
-            finish()
+                'cache lock at %s mid-walk, applying partial chain', last_str)
+            step_finish(task, chain, http_extended)
           elseif type(d) ~= 'string' then
             -- True cache miss mid-walk: a previous chain link points to a
             -- URL whose own cache entry is gone (TTL expired or evicted).
             -- Resume live HTTP from this dead end so the chain rebuilds and
             -- gets re-cached, instead of giving up with a truncated chain.
             lua_util.debugm(N, task,
-                'cache miss for %s mid-walk, extending with live HTTP', last)
+                'cache miss for %s mid-walk, extending with live HTTP', last_str)
             http_walk(task, orig_url, last, ntries + 1, chain, seen)
           else
             step(task, orig_url, chain, seen, d, ntries, http_extended)
@@ -354,7 +357,7 @@ step = function(task, orig_url, chain, seen, data, ntries, http_extended)
         'GET', { next_key })
     if not ret then
       rspamd_logger.errx(task, 'cannot make redis request to walk chain')
-      finish()
+      step_finish(task, chain, http_extended)
     end
     return
   end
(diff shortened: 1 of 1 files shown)


vstakhov (commented):
Commit is GPG-signed; `[Fix]` tag conforms; luacheck is clean. The bug analysis in the commit message is precise and the fix is correct: when `http_walk` splices into `step` on a cache hit at the redirect target, the new prefix→redir link must go through `finalize_chain` (which writes back) and not just `apply_redirect_chain` (which only links + injects), otherwise `hash(orig_url)` stays stuck at `'processing'` until the 13-second TTL expires.

Below focuses on non-canonical Lua-API patterns. This PR is small and the new code is mostly correct; the issues are minor.

## Non-canonical Lua API usage

### 1. `tostring(ndata)` is redundant
`url_redirector.lua:621-623`:
```lua
lua_util.debugm(N, task,
    'failed to claim lock for %s (held by another worker or stale processing marker, ndata=%s); skipping this scan',
    orig_url, tostring(ndata))
```
`rspamd_logger`'s `%s` is the universal placeholder for all argument types; `tostring()` is done internally. (Confirmed by every other `%s` usage in this file passing URL objects directly via `__tostring`.) Drop the wrapper:
```lua
orig_url, ndata)
```

### 2. `task:get_urls() or {}` — `or {}` is dead code
`url_redirector.lua:717-718`:
```lua
if #selected == 0 then
  local task_urls = task:get_urls() or {}
  if #task_urls > 0 then
```
`task:get_urls()` (`src/lua/lua_task.c:2649-2652`) always pushes a table when the task is valid (empty when the task has no message), and errors otherwise. The `or {}` never fires. Just:
```lua
local task_urls = task:get_urls()
if #task_urls > 0 then
```

Also worth considering: this debug branch only runs when `#selected == 0`, but always issues `task:get_urls()` for a debug-level message. `task:get_urls()` materialises the full URL set into a Lua table (allocations + a `qsort` if `sort=true` was set elsewhere — though here it isn't). Wrap in a `lua_util.debugm`-only path or guard with the existing N-debug enabled check, so production traffic doesn't pay for the table allocation just to find out the count.

A cheaper alternative if available is `task:get_urls_count()`-style metadata, or just iterate a small probe — but the simplest is gating on debug enable. (Minor.)

### 3. `finish` closure constructed per `step()` invocation
`url_redirector.lua:317-323`:
```lua
local function finish()
  if http_extended then
    finalize_chain(task, chain, nil)
  else
    apply_redirect_chain(task, chain)
  end
end
```
`step()` is recursive on `^hop:` chains and on `data == nil` re-entry, so each cache hop allocates a fresh closure capturing `task`, `chain`, `http_extended`. For very long cached chains this is non-zero overhead. Two ways out:
- Hoist the dispatch inline at each terminal: `if http_extended then finalize_chain(task, chain, nil) else apply_redirect_chain(task, chain) end` — verbose.
- Pre-bind once at the entry frame and pass `finish` as a parameter to recursive `step`/`http_walk` calls — cleaner.

Closure allocation per hop is dominated by the Redis round-trip in practice, so this is a minor optimisation. Flagging only because it's a hot-path-adjacent pattern that's easy to get right when first written.

### 4. `last` is `__tostring`'d up to twice per Redis callback
`url_redirector.lua:327, 333, 340, 348`:
```lua
local next_key = cache_key_for_url(tostring(last))
...
function(e, d)
  if e then
    rspamd_logger.errx(task, '... at %s: %s', last, e)
  elseif d == 'processing' then
    lua_util.debugm(N, task, 'cache lock at %s ...', last)
  elseif type(d) ~= 'string' then
    lua_util.debugm(N, task, 'cache miss for %s ...', last)
```
`tostring(last)` is computed once outside (for the cache key) and then each error/debug branch passes `last` as `%s` again, which re-invokes `__tostring` (the slow path that percent-encodes whitespace via `lua_url_push_encoded` in `src/lua/lua_url.c:339-376`). Since `tostring(last)` is already in scope at the outer frame, capture it once:
```lua
local last = chain[#chain]
local last_str = tostring(last)
local next_key = cache_key_for_url(last_str)
local ret = lua_redis.redis_make_request(task,
    redis_params, next_key, false,
    function(e, d)
      if e then
        rspamd_logger.errx(task, 'redis error during chain walk at %s: %s', last_str, e)
        ...
```
Hot path (cache hit) doesn't pay for it; only matters on the err / processing / miss branches. Minor.

## Correctness observations

### `http_extended` propagation — verified correct
- top-level `step` from `redis_get_cb` (`:605`) passes no flag → defaults to `false` → cache-only walk → `apply_redirect_chain`. ✓
- `http_walk` splice at `:527` always passes `true` → terminal → `finalize_chain` writes the new link. ✓ (this is the fix)
- recursive `step` on `^hop:` (`:386`) preserves the flag through the hop chain. ✓
- new `^hop:` chain that hits a true cache miss mid-walk (`:349`) hands off to `http_walk` and discards the flag — fine, because `http_walk` always finalises on its own terminals. ✓

### New "true cache miss mid-walk" branch — same shape as existing `^nested` extension
The new branch at `:342-349` calls `http_walk(task, orig_url, last, ntries + 1, chain, seen)` after `last` was already added to `seen` by a prior iteration. This mirrors the pre-existing `^nested` bridge at `:400` which has the same shape. The cycle guard at `http_walk:438` (`if ntries > 1 and seen[url_str] then ... cycle`) hinges on the `ntries == 1` skip-comment to avoid false-firing — which holds at top-level (`ntries=0` → `+1=1`, skip applies) but not when `step` was entered via an `http_walk` splice with `ntries > 0`. This is a pre-existing concern not introduced by this PR (the same theoretical false-positive applies to the `^nested` path), and only fires if `tostring(rspamd_url.create(val))` round-trip-stably equals `val`. Worth a follow-up if you ever observe `'cycle in http walk'` debug logs you can't explain — but not blocking this PR.

### `step` parameter list now at 7 positional args
`task, orig_url, chain, seen, data, ntries, http_extended`. Adding more state will hurt; consider folding `(ntries, http_extended)` into a small `walk_state` table at the next change. Not actionable now.

## Verdict
LGTM modulo the `tostring(ndata)` and `or {}` nits. The fix is correct and well-explained. Other items are micro-polish.

dragoangel:
Thanks for details review, will fix mentioned stuff 👍

dragoangel:
Done, please review :)

PR #6018, opened 2026-05-04
[Feature] elastic: log Reply-To, received IPs, URL metadata, and pre-result module

dragoangel:
Changes:
- reply_to_user / reply_to_domain: parsed from Reply-To via rspamd_util.parse_mail_address, mirroring the from / mime_from split.
- received_ips: list of IPs from Received headers
- urls and urls_cta with the new collect_urls config block: per-URL records {url, etld, host, protocol, flags, count} plus aggregate metrics {total, unique, max_repeats, repeat_ratio}. CTA URLs are collected via text_part:get_cta_urls({original=true}) and walked via :get_redirected so url_redirector-resolved hops are captured, then either kept inline at the top of urls (sorted ahead of non-CTA so they survive max_urls truncation) or emitted into a dedicated urls_cta when separate_cta is on
- action_forced: the module name from task:has_pre_result(), so logs show which prefilter short-circuited the pipeline (or 'no force').

Renames get_received_delay to get_received_info (returns delay + ips in one pass over the received chain) and replaces the local merge_settings helper with lua_util.override_defaults — the two are functionally equivalent recursive deep-merges, but override_defaults is the project-wide maintained helper.

### DIFF
--- a/conf/modules.d/elastic.conf
+++ b/conf/modules.d/elastic.conf
@@ -49,6 +49,7 @@ elastic {
     headers_count_ignore_above = 5; # Record only the first N same-named headers, add "ignored above..." if the limit is reached; set 0 to disable the limit
     headers_text_ignore_above = 2048; # Strip specific header value and add "..." to the end; set 0 to disable the limit
     symbols_nested = false;
+    urls_nested = false;
     empty_value = "unknown"; # Empty numbers, IPs and IP nets are not customizable; they will always be 0, :: and ::/128 respectively
   };
   index_policy = {
@@ -97,6 +98,21 @@ elastic {
   # "Precedence";
   # "List-Id";
   extra_collect_headers = [];
+  collect_urls = {
+    enabled = true;
+    max_urls = 0; # 0 = no limit; otherwise truncate (CTAs and most-repeated URLs are placed first so they survive truncation)
+    separate_cta = true; # If true, also emit urls_cta with just CTA URLs from HTML parts
+    full_urls = false; # Log full URL text vs just host
+    full_cta_urls = false; # Log full URL text vs just host (for CTA URLs)
+    # Params passed verbatim to task:get_urls(); see https://docs.rspamd.com/lua/rspamd_task/#m1860f
+    get_url_params = {
+      content = true; # Include URLs found inside text/html content
+      images = false; # Include URLs from <img src> attributes (noisy)
+      emails = false; # Include mailto: URLs (rarely useful for phishing analysis)
+      sort = true; # Deterministic order, stable dedup, reproducible doc layout
+      protocols = ["http", "https", "ftp"]; # Protocols to collect
+    };
+  };
   geoip = {
     enabled = true;
     managed = true;
--- a/src/plugins/lua/elastic.lua
+++ b/src/plugins/lua/elastic.lua
@@ -157,7 +157,6 @@ local settings = {
       content = true, -- include URLs found inside text/html content
       images = false, -- include URLs from <img src> attributes (noisy)
       emails = false, -- include mailto: URLs (rarely useful for phishing analysis)
-      sort = true, -- deterministic order, stable dedup, reproducible doc layout
       protocols = { 'http', 'https', 'ftp' }, -- protocols to collect
     },
   },
@@ -332,14 +331,13 @@ local function get_received_info(received_headers)
       end
       -- '...' is rspamd's "valid but unknown" placeholder; elastic's t_ip mapping rejects it,
       -- so guard against it here even though :is_valid() returns true for that case.
-      local ip_str
       local real_ip = received_header['real_ip']
-      if real_ip and real_ip:is_valid() and tostring(real_ip) ~= '...' then
-        ip_str = tostring(real_ip)
-      end
-      if ip_str and not seen_ips[ip_str] then
-        seen_ips[ip_str] = true
-        table.insert(ips, ip_str)
+      if real_ip and real_ip:is_valid() then
+        local s = tostring(real_ip)
+        if s ~= '...' and not seen_ips[s] then
+          seen_ips[s] = true
+          table.insert(ips, s)
+        end
       end
     end
   end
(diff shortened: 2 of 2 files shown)


vstakhov (commented):
Both commits are GPG-signed; tags conform; luacheck is clean. The feature is well-scoped and the inline comments are unusually thoughtful (e.g. the `'...'` placeholder note, the rationale for `original=true` on `get_cta_urls`).

Below focuses on non-canonical Lua-API patterns. Most are micro-issues, but several execute on every URL of every scanned message.

## Non-canonical Lua API usage

### 1. `tostring(u:get_text())` is redundant — twice
`elastic.lua:363` and `elastic.lua:383`:
```lua
rec.url = tostring(u:get_text())
...
return tostring(u:get_text())
```
`url:get_text()` already returns a Lua string (`lua_pushlstring` in `src/lua/lua_url.c:319`). It only returns userdata when called with `as_text=true`. The wrapping `tostring()` is dead weight on every URL we serialize.

### 2. `tostring(u)` is the **expensive** way to dedup URLs
`elastic.lua:472` (in `add_with_chain`):
```lua
local key = tostring(u)
```
This invokes the URL `__tostring` metamethod (`lua_url_tostring`), which percent-encodes whitespace via `lua_url_push_encoded` (a two-pass walk). For dedup keys you don't need the encoded form; `u:get_text()` is a single `lua_pushlstring` over the raw buffer.

### 3. `tostring()` around a number is redundant in concatenation
`elastic.lua:387`:
```lua
tostring(u:get_flags_num())
```
Lua's `..` already coerces numbers to strings. Drop it.

### 4. `or nil` is a no-op
`elastic.lua:478`:
```lua
u = u:get_redirected() or nil
```
For any value `v`, `v or nil` evaluates to `v`. Just `u = u:get_redirected()`.

### 5. `u:get_host()` called twice per record
`elastic.lua:357-358`:
```lua
etld = u:get_tld() or u:get_host(),
host = u:get_host(),
```
Cache it once.

### 6. Redundant `if on then` guard on the flag table
`elastic.lua:365-371`. `url:get_flags()` (`src/lua/lua_url.c:1085-1108`, `PUSH_FLAG`) only inserts a key when the flag bit is set — every value in the returned table is `true`. The `if on then` branch is unreachable-as-false:
```lua
for name in pairs(u:get_flags() or {}) do
  table.insert(flag_list, name)
end
```
`u:get_flags()` is the heaviest URL accessor (allocates a table, pushes a string per set bit). It's only called after sorting/truncation here, which is good.

### 7. `tostring(real_ip)` called twice
`elastic.lua:337-338`:
```lua
if real_ip and real_ip:is_valid() and tostring(real_ip) ~= '...' then
  ip_str = tostring(real_ip)
end
```
Cache the first `tostring`. (Same anti-pattern at `:484`/`:495` is pre-existing — not your bug, but worth not propagating.)

### 8. `url_key(u, false)` computed twice per URL when `full_urls=false`
`elastic.lua:403,405`. Compute once, reuse for the `cta_keys` lookup; only recompute if `full_urls` is true:
```lua
local nfkey = url_key(u, false)
local is_cta = cta_keys and cta_keys[nfkey] or false
if not (is_cta and cta_mode == 'skip') then
  local key = full_urls and url_key(u, true) or nfkey
  ...
```

### 9. `task:get_urls(... sort = true ...)` is wasted work here
`elastic.lua:861`. `lua_task_get_urls` (`src/lua/lua_task.c:2675-2694`) `qsort`s all URLs when `sort=true`. The downstream code immediately puts them into a hash for dedup and re-sorts the deduped entries by count — the C-level sort doesn't survive into the output. The comment `Deterministic order, stable dedup, reproducible doc layout` also overstates what the flag does: `table.sort` over `order` is unstable in standard Lua, so ties between equal-count URLs aren't actually pinned regardless.

## Other observations

### Type inconsistency on `received_ips`
`elastic.lua:854-858`. When populated, this is `["1.2.3.4", ...]`; when empty, the literal string `"::"`. ES tolerates both, but the sentinel makes "no Received IPs" look identical to "Received IP was the unspecified address." Consider omitting the field when empty, or sending `[]`.

### Behavior change in `get_received_info` — verified equivalent
The inner `if timestamp == 0` preserves "first valid wins" for `delay` while letting the loop continue to harvest IPs. ✓

### `lua_util.override_defaults` swap — verified equivalent
Functionally identical to the removed local `merge_settings` for this call site. One semantic difference: `override_defaults` returns a new table and you reassign `settings`. All closures capture `settings` as an upvalue (a Lua local), so the rebind is propagated — safe. ✓

### Queue fix (commit `c07cefc`)
Correct. Wrap-around only on full drain, so partial pops don't churn indices. ✓

## Suggested patch (concentrates 1, 2, 3, 4, 5, 6, 7, 8)

```lua
local function url_to_record(u, full_urls, count)
  local host = u:get_host()
  local rec = {
    etld = u:get_tld() or host,
    host = host,
    protocol = u:get_protocol(),
    count = count,
  }
  if full_urls then
    rec.url = u:get_text()
  end
  local flag_list = {}
  for name in pairs(u:get_flags() or {}) do
    table.insert(flag_list, name)
  end
  if #flag_list > 0 then
    rec.flags = flag_list
  end
  return rec
end

local function url_key(u, full_urls)
  if full_urls then
    return u:get_text()
  end
  return (u:get_host() or '') .. '|' ..
      (u:get_protocol() or '') .. '|' ..
      u:get_flags_num()
end
```

…and inside `add_with_chain`:
```lua
local key = u:get_text()
...
u = u:get_redirected()
```

…and inside `get_received_info`:
```lua
local real_ip = received_header['real_ip']
if real_ip and real_ip:is_valid() then
  local s = tostring(real_ip)
  if s ~= '...' and not seen_ips[s] then
    seen_ips[s] = true
    table.insert(ips, s)
  end
end
```

## Verdict
Approve in spirit; request the Lua-API tightening above before merge. None of it is structural — single-line cleanups in newly added hot-path code that runs per-URL per-message.

dragoangel:
Thank you, will adjust PR to fix issues :)

dragoangel:
@vstakhov done, please review.

One note regarding point: Type inconsistency on received_ips - `::` is done on purpose, this way it already done for ip & sender_ip - elastic has problems to find fully "skipped" fields, so to properly able to aggregate data `::` was been chosen as "empty ip" indicator, in same way we fill "empty ipnet" as `::/128` and "empty" `unknown` for any other string.

PR #6020, opened 2026-05-05
[Feature] lua_extras: structured custom lua loader

vstakhov:
## Summary

- Add `lualib/lua_extras` with `register_selector` / `register_map` / `register_regexp` helpers and a `load_dir(cfg, dir, kind)` directory loader.
- Wire structured loading into `rules/rspamd.lua`: files in `$LOCAL_CONFDIR/lua.local.d/{selectors,maps,regexps}/*.lua` are loaded before `rspamd.local.lua`. Each file returns a `{ name = def }` table whose entries are dispatched to the matching helper.
- Ship `example.lua.example` files in each new subdirectory documenting the expected contract.

## Why

Today, distributions and add-ons that want to ship custom selectors, maps or regexp rules have no place to put them other than `rspamd.local.lua` or free-form `lua.local.d/*.lua` — both of which end users routinely edit. Selectors specifically had no convenient registration helper at all (`lua_selectors.register_extractor` is callable, but adding maps/regexps required separate boilerplate).

The new tree gives a stable, structured location that vendor code can use without colliding with end-user customization, while keeping the existing free-form behaviour intact.

## Load order

1. `lua.local.d/{selectors,maps,regexps}/*.lua` — structured (NEW)
2. `rspamd.local.lua` (unchanged)
3. `lua.local.d/*.lua` at root (existing free-form)
4. `local.d/rspamd.lua` (unchanged)

End users can still override anything shipped via the structured tree from either `rspamd.local.lua` or a root-level free-form file.

## File contract

Each structured file must `return` a table; the loader iterates it and calls the matching helper:

- `selectors/*.lua` — `{ name = fn }` or `{ name = { get_value = fn, description = ..., type = ... } }`
- `maps/*.lua` — `{ name = add_map_args_table }`; the resulting map is stored in `rspamd_maps[name]`
- `regexps/*.lua` — `{ SYMBOL = { re = ..., score = ..., description = ..., condition = ... } }`

Errors in any single file are logged with the file path via `rspamd_logger.errx` and skipped — they never abort startup.

## Reload

Picked up on `rspamadm reload` / SIGHUP, same as the existing `lua.local.d/*.lua` files.

## Test plan

- [x] `luacheck lualib/lua_extras.lua rules/rspamd.lua` — clean
- [x] `ninja install` — `lua_extras.lua` installed via existing `GLOB_RECURSE` rule, no CMake changes needed
- [x] `test/rspamd-test -p /rspamd/lua` — no new failures (unchanged: 1003/1014, 8 pre-existing errors in `lua_redis.prepare_setup.lua`)
- [x] `rspamadm configtest` with example selector/regexp files dropped into the runtime config dir — symbols register, `configdump` shows them with re/score/description, ends with `syntax OK`
- [ ] Reviewer sanity check on file contract / naming

### DIFF
--- a/conf/lua.local.d/maps/example.lua.example
+++ b/conf/lua.local.d/maps/example.lua.example
@@ -1,0 +1,24 @@
+-- Custom map definitions loaded by lua_extras at config time.
+--
+-- Each file in $LOCAL_CONFDIR/lua.local.d/maps/*.lua must return a table
+-- where keys are map names and values are the argument table accepted by
+-- rspamd_config:add_map{...}, see https://rspamd.com/doc/lua/config.html.
+--
+-- Registered maps are stored in the global `rspamd_maps` table so other
+-- lua code can look them up by name.
+
+return {
+  -- A simple set map loaded from a file (URL can also be http(s):// etc).
+  example_local_domains = {
+    type        = 'set',
+    description = 'Locally hosted domains',
+    url         = '${LOCAL_CONFDIR}/local_domains.list',
+  },
+
+  -- A regexp map of keys to string values.
+  example_brand_regexps = {
+    type        = 'regexp',
+    description = 'Brand name regexps',
+    url         = '${LOCAL_CONFDIR}/brand_regexps.list',
+  },
+}
--- a/conf/lua.local.d/regexps/example.lua.example
+++ b/conf/lua.local.d/regexps/example.lua.example
@@ -1,0 +1,28 @@
+-- Custom regexp rules loaded by lua_extras at config time.
+--
+-- Each file in $LOCAL_CONFDIR/lua.local.d/regexps/*.lua must return a table
+-- where keys are symbol names and values are regexp rule definitions equivalent
+-- to assigning into config['regexp'][SYMBOL] from rspamd.local.lua.
+--
+-- See https://rspamd.com/doc/developers/writing_rules.html for the full DSL.
+
+local re_from_foo = 'From=/foo@/H'
+local re_subject_blah = '/blah/P'
+
+return {
+  EXAMPLE_FROM_FOO = {
+    re          = re_from_foo,
+    score       = 1.2,
+    description = 'Sender contains foo@',
+  },
+
+  EXAMPLE_FROM_FOO_NO_BLAH = {
+    re          = string.format('(%s) && !(%s)', re_from_foo, re_subject_blah),
+    score       = 0.5,
+    description = 'foo@ sender without "blah" in subject',
+    -- Optional condition gating the rule:
+    condition   = function(task)
+      return task:has_from('mime')
+    end,
+  },
+}
--- a/conf/lua.local.d/selectors/example.lua.example
+++ b/conf/lua.local.d/selectors/example.lua.example
@@ -1,16 +1,21 @@
 -- Custom selector definitions loaded by lua_extras at config time.
 --
 -- Each file in $LOCAL_CONFDIR/lua.local.d/selectors/*.lua must return a table
--- where keys are selector names and values are either:
---   * a function(task) returning the value, or
+-- where keys are selector names and values are one of:
+--   * a function(task) returning the value (shortcut for { get_value = fn });
 --   * a full selector definition table:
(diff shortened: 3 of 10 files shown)


vstakhov:
Pushed `7d69711ce` adding a **two-phase preload→resolve loader** so selectors can declare real cross-kind dependencies.

### What changed
- **Two-phase loading** (`lua_extras.load_extras(cfg, base_dir)`): phase 1 reads every file in `lua.local.d/{maps,selectors,regexps}/*.lua` and collects entries into per-kind buffers; phase 2 registers them in dependency order (maps → selectors → regexps).
- **`lua_extras.deferred(factory_fn)`**: opt-in wrapper. Wrapped entries are evaluated during phase 2 *after* all earlier-kind entries are registered, so a selector can capture `rspamd_maps[name]` or compile an `rspamd_regexp` from map data at definition time.
- **`re_selector`** field on selector defs. When `true` (or `{ selector = "...", delimiter = " " }`), the helper also calls `cfg:register_re_selector(...)` so the selector name is usable in the regexp DSL: `name=/regex/{selector}`.
- `rules/rspamd.lua` now calls `load_extras` once instead of looping per kind. `load_dir` is retained for callers that only need one kind.

### End-to-end proof — selector consuming a map fires a regexp symbol

Test files dropped into `$LOCAL_CONFDIR/lua.local.d/`:

```lua
-- maps/test_extras.lua
return { example_local_domains = {
  type = 'set', url = '/tmp/lua_extras_test/local_domains.list' } }
```
```lua
-- selectors/test_extras.lua
local lua_extras = require "lua_extras"
return {
  test_extras_from_local_domain = lua_extras.deferred(function(_cfg)
    local domains = rspamd_maps['example_local_domains']  -- live ref
    return {
      re_selector = true,
      get_value = function(task)
        local from = task:get_from('mime')
        local dom = from and from[1] and from[1].domain
        return dom and domains:get_key(dom) and dom or nil, 'string'
      end,
    }
  end),
}
```
```lua
-- regexps/test_extras.lua
return { TEST_EXTRAS_LOCAL_FROM = {
  re = 'test_extras_from_local_domain=/.+/{selector}', score = 0.1 } }
```

Map data: `example.com\nlocal.test\n`.

Scan results:

```
=== From: alice@example.com (in map) ===
Symbol: TEST_EXTRAS_LOCAL_FROM (0.10)

=== From: alice@unknown.test (NOT in map) ===
(symbol absent — correct)
```

Log lines from the run:

```
added map /tmp/lua_extras_test/local_domains.list
TEST_EXTRAS deferred factory: domains=rspamd{map}(0x114a547c8)
lua_config_register_r registered regexp selector test_extras_from_local_domain
TEST_EXTRAS from selector: domain=example.com hit=true
TEST_EXTRAS from selector: domain=unknown.test hit=nil
```

The deferred factory captured a live `rspamd{map}` ref before the selector was registered, the regexp symbol fired exactly when the captured map said the domain matched, and stayed silent when it didn't.

vstakhov:
Added a functional test in `ebd16ed33`: `Functional.Cases.001_Merged.271_Lua_Extras`.

`test/functional/lua/lua_extras_test.lua` stages a tree under `$TMPDIR` with `maps/`, `selectors/`, `regexps/` subdirectories, then calls `lua_extras.load_extras` on it. The selector is wrapped in `lua_extras.deferred(...)` so its factory captures `rspamd_maps[name]` at registration time, exercising the maps → selectors → regexps phase-2 ordering and the `re_selector` auto-binding into the regexp DSL.

Two cases assert the regexp symbol fires only when the From-domain is in the map:

```
$ env RSPAMD_INSTALLROOT=/usr/local python3 -mrobot -v RSPAMD_USER:vstakhov \
    --suite "Functional.Cases.001_Merged" --test "*lua_extras*" test/functional/

Functional.Cases.001 Merged.271 Lua Extras
==============================================================================
Lua_extras deferred selector fires regexp when From in map            | PASS |
Lua_extras deferred selector silent when From not in map              | PASS |
2 tests, 2 passed, 0 failed
```

Adjacent suite (`*selector*` filter, 6 tests including 270_selector) still passes after wiring the new lua line into `merged.conf`.

PR #6026, opened 2026-05-08
Allow to add urls to redis history export

JustKiddingCode:
Add new flag include_urls (default false), when set to true, the redis history export will have a key urls as an array containing the urls of the mail.


vstakhov:
Please don't add any more stuff to history redis - it has never been designed for that. We have Clickhouse and Elastic modules that are exactly designed for that sort of data. I want to keep history_redis as minimal as I can (meaning as minimal as it is so far).

Issue #6003, opened 2026-04-27
[BUG] Web Interface Fuzzy delete tries to delete from all storages


poettig:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**
When trying to delete a fuzzy hash from a specific storage + flag, when there are different storages using the same flag, rspamd tries to delete from all storages that use the flag instead of only from the storage + flag selected in the dropdown. This causes a "503" error showing up if one of them is not writable.

Example: We have a local fuzzy storage and bulk-rspamd.com:FUZZY_BULK both use the flag 1, which should not be a problem as those are different storages in different rules. As FUZZY_BULK ist not writable, an error shows up.

This makes sense as the web interface only seems to send the email as request body and the flag number as request header "Flag".

I derive that all matching storages are hit with a delete from the rspamd log when I press "delete":
```
<8ef1d1>; task; fuzzy_controller_io_callback: deleted fuzzy hash (txt) 411be9c391d58756bed439ce8e32ec4427f8c76742d0f4626d32a4e4e3f73a45231d916f0dde6a234407ad642f8afea5a1eabfec9d8b00d3e43d68bde57edb51, list: FUZZY_MANUAL:1 for message <1305a9ff-e902-4cd6-ad50-d21218da1542@redacted.com>
<8ef1d1>; task; fuzzy_controller_io_callback: deleted fuzzy hash (bin) ff3aa6edd656bc9cb12974a202ab5ac05c607663299bbf6a57af23e1678519a05a07d83370e744ffcb8456e56e3d767d89b2bf54feb0a1430b03e990513bed31, list: FUZZY_MANUAL:1 for message <1305a9ff-e902-4cd6-ad50-d21218da1542@redacted.com>
[...]

PR #6027, opened 2026-05-08
[Minor] Warn on fuzzy flag collisions across writable rules

moisseev:
Issue: #6003
PR: rspamd/docs.rspamd.com#99


moisseev:
@droid review

vstakhov:
Maybe we should just keep a global set of all known flags, so we won't need any temporary hash tables on rules parsing logic.

vstakhov:
Thanks!

PR #6029, opened 2026-05-09
[Minor] Add backend none for Prometheus

bruflu:
The metric_exporter plugin provides an endpoint for Prometheus. If there is no graphite present on the system metric_exporter logs this message every time Prometheus scrapes the metrics:
```
May  8 00:11:50 vultr rspamd[57711]: <>; lua; metric_exporter.lua:137: Push failed: Socket error detected: Connection refused
```
I add the backend `none` which only sets the timestamp of the last export. With this backend I can use Prometheus to scrape metrics without filling the logs with the message above.

### DIFF
--- a/src/plugins/lua/metric_exporter.lua
+++ b/src/plugins/lua/metric_exporter.lua
@@ -142,10 +142,33 @@ local function graphite_push(kwargs)
   })
 end
 
+local function none_config()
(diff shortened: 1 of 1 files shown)


fatalbanana:
`metric_exporter` is not needed for Prometheus metrics to work

PR #6035, opened 2026-05-11
[Minor] url_redirector: skip non-HTTP(S) URLs in http_walk

moisseev:
Non-HTTP(S) schemes (such as tel:, mailto:, etc.) cannot have HTTP redirects. Attempting to follow them in http_walk is unnecessary and could potentially lead to errors. This change skips these URLs early in the redirect chain walk.


moisseev:
@droid review

moisseev:
@droid review

moisseev:
@droid review

Issue #6032, opened 2026-05-09
[Feature] Improvements for MX check module and TCP library


dragoangel:
## Summary

`mx_check` is the Rspamd module that probes a sender domain's MX records to verify the host actually accepts SMTP — the signal helps detect forged, disposable, and misconfigured envelopes. The module has been in the tree for years and works, but several aspects have accumulated rough edges that affect signal quality and operational behaviour.

This issue proposes a coherent set of improvements:

- A **three-layer Redis cache** (`d:` / `m:` / `i:`) that lets domains sharing MX hosts (every G-Suite / M365 tenant, every ESP customer) reuse a single set of probe verdicts.
- A **finer error/outcome taxonomy** that splits the current `MX_INVALID` into specific symbols per actually-distinguishable network outcome (`MX_REFUSED`, `MX_TIMEOUT_CONNECT`, `MX_TIMEOUT_READ`, `MX_ERROR` for SMTP-level rejections, `MX_INVALID` for non-SMTP listeners), plus DNS-level signals (`MX_NXDOMAIN`, `MX_NULL`, `MX_BROKEN`) and IP-class signals (`MX_LOCAL_*`, `MX_BOGON_*`).
- A **per-source symbol family** (`MIME_FROM_*`, `REPLY_TO_*` alongside envelope-From) so operators can score the same outcome differently depending on which header surfaced the domain.
- **Per-layer trust/skip maps** (`exclude_domains`, `exclude_mxs`, `exclude_ips`) for whitelisting at the granularity that matches operator intent.
- A small enabling change in `lua_tcp` — phase-specific timeouts and a connect-phase error callback — that makes the probe shapes clean to express.

The proposal is layered: most of it is changes inside `mx_check.lua` itself. A smaller part is the `lua_tcp` change. Without the latter, two of the `mx_check` improvements (clean per-phase timeouts, distinct connect-vs-read failure callbacks) can't be expressed cleanly.

## Problem

The current `mx_check` has several interacting limitations:

1. **Cache is keyed by sender domain only.** Every distinct sender domain costs a fresh DNS + TCP probe even if it shares MX hosts with thousands of domains we've already verified. SMTP MX targets cluster heavily (`aspmx.l.google.com` for every G-Suite tenant, `*.outlook.com` for every M365 tenant, large hosting providers' shared MTAs, etc.), so this is a lot of avoidable work.
[...]

PR #6039, opened 2026-05-14
[Rework] mx_check: three-layer cache, finer outcomes, IP-class classification (#6032 Phases A & C)

vstakhov:
## Summary

Reworks the `mx_check` core (Phase A) and adds IP-class classification, per-layer trust maps, and run-scope toggles (Phase C) from the plan in #6032. **Operator-visible scoring is unchanged**: the primary symbols (`MX_GOOD`, `MX_INVALID`, `MX_MISSING`, `MX_WHITE`) keep today's defaults and emission rules. Every new symbol is emitted alongside at score 0 so deployments can collect tuning data before Phase B turns on real default scores.

Depends on the `lua_tcp` enabling change in #6034 (already merged) — phased timeouts and the connect-phase `on_error` callback.

Phase B (two-path `MX_*` / `MX_A_*` matrix, per-source `MIME_FROM_*` / `REPLY_TO_*` families, retiring `MX_MISSING`) and IPv6 probing remain out of scope, landing separately per #6032.

## Phase A — core rework

**Three-layer Redis cache** under `<key_prefix>:` (default `rmx:`):

| Key               | Value examples                                             |
|-------------------|------------------------------------------------------------|
| `rmx:d:<domain>`  | `mx:host1:10,host2:5` / `mx_miss:1.2.3.4` / `nxd` / `null` |
| `rmx:m:<mx_host>` | `1.2.3.4,5.6.7.8` / `nxd`                                  |
| `rmx:i:<ip>`      | `good` / `refused` / `timeout_connect` / `timeout_read` / `invalid` / `error:<code>` |

Two domains pointing at a shared MX host reuse the `m:` and `i:` entries — the second hits cache at every step. Old single-layer keys age out under their own TTL; no migration needed.

**Probe-shape split** using #6034's options: `probe_connect_only` (TCP-connectivity check, distinguishes refused / timeout / unreachable via `on_error`) and `probe_with_greeting` (banner parse with RFC 5321 §4.2.1 multi-line support and reply-code validation).

**Greeting controls.** `wait_for_greeting` (conflated "read the banner" with "send QUIT", and emitted `MX_GOOD` on *any* CRLF-terminated line, including mid-banner) is replaced by `verify_greeting` (validate the reply code: 2xx → good, 4xx/5xx → `MX_ERROR`) and a separate `send_quit` flag that waits for the final banner line.

**New outcome symbols** (score 0; primary `MX_INVALID` still fires alongside): `MX_REFUSED`, `MX_TIMEOUT_CONNECT`, `MX_TIMEOUT_READ`, `MX_ERROR` (4xx/5xx, code as parameter), `MX_NXDOMAIN`, `MX_NULL` (RFC 7505 Null MX), `MX_BROKEN` (all MX RRs unresolvable).

**Deprecation aliases**, mapped at config load with a warning: `timeout` → `connect_timeout`, `wait_for_greeting` → `verify_greeting`. The shipped `modules.d/mx_check.conf` deliberately does not pre-set the new keys, so a legacy alias in `local.d` is honoured.

## Phase C — IP-class classification, trust maps, run scope

**IP-class classification.** Resolved MX-target IPs are partitioned PUBLIC / LOCAL / BOGON against fixed RFC range sets. LOCAL (RFC1918 / CGNAT / ULA) is unprobeable from our vantage point — probing it reaches our own LAN, not the publisher's. BOGON (loopback, link-local, TEST-NET, multicast, reserved) has no legitimate meaning as a published MX target and is a packet-injection footgun. Only PUBLIC IPs are probed; the rest emit `MX_LOCAL_ONLY` / `MX_LOCAL_MIX` and `MX_BOGON_ONLY` / `MX_BOGON_MIX`. The range sets are a correctness invariant — not operator-tunable.

**Per-layer trust/skip maps.** `exclude_mxs` (glob map — a bare hostname matches exactly, an explicit `*.foo` matches subdomains) short-circuits the whole check with `MX_WHITE` when a trusted MX host is seen. `exclude_ips` (radix) drops matched IPs from the probe set; if the set empties as a result, `MX_SKIP` fires. These join the existing `exclude_domains`.

**Run-scope toggles.** `check_authorized` and `check_local` (both default `false`) control whether authenticated and local-network senders are checked — useful for ESP / outbound deployments. The default (skip both) preserves today's behaviour.

**`test_mode`** (testing only, default `false`): lifts loopback out of the bogon set so the probe path can be exercised against a local listener. Never enable it in production.

## Caveats

- The IPv4-mapped range `::ffff:0:0/96` is intentionally left out of the bogon set: rspamd's radix stores every IPv4 address as its v4-mapped form, so listing that prefix would classify all IPv4 traffic as bogon. v4-mapped IPv6 MX targets can only arise once AAAA probing lands and must be handled outside the radix test.
- `lua_tcp_request` still validates `callback` even with `read = false`; the connect-only probe passes a no-op stub.
- UCL doesn't merge multiple top-level `options{}` blocks; the functional test config splits its `fake_records` into a separate include with `duplicate=merge`.

## Test plan

- `luacheck src/plugins/lua/mx_check.lua` — clean.
- `cases/167_mx_check.robot` — 11/11 pass: Null MX, NXDOMAIN, broken MX, NODATA-is-not-NXDOMAIN, connect-refused, A-fallback, `MX_LOCAL_ONLY`, `MX_BOGON_ONLY`, `exclude_mxs` → `MX_WHITE`, `exclude_ips` → `MX_SKIP`, local-sender skip.
- `cases/230_tcp.robot` — passes (regression check on `lua_tcp` callers).

Refs #6032.

### DIFF
--- a/conf/modules.d/mx_check.conf
+++ b/conf/modules.d/mx_check.conf
@@ -21,19 +21,53 @@
 # You also need to define redis servers for this module
 
 mx_check {
-  # TCP connection timeout in seconds. The effective symbol timeout is
-  # timeout + dns.timeout (see options.inc), so task_timeout must be set
-  # higher than their sum to avoid forced task termination.
-  timeout = 1.0;
-  # symbol yielded if no MX is connectable
+  # Per-phase TCP timeouts (seconds). The effective symbol budget is
+  # connect_timeout + read_timeout + dns.timeout (see options.inc);
+  # task_timeout must be higher than that sum to avoid forced termination.
+  connect_timeout = 1.0;
+  # read_timeout is only used when verify_greeting = true.
+  read_timeout = 5.0;
+
+  # Legacy `timeout` is still accepted and is mapped to connect_timeout
+  # at module load with a deprecation warning. Don't set both.
+
+  # SMTP greeting validation. When verify_greeting = true the probe reads
+  # the SMTP banner, validates the reply code (220 -> good, 4xx/5xx ->
+  # MX_ERROR, other -> MX_INVALID) and handles multi-line banners
+  # correctly. send_quit, when also true, issues `QUIT` after the final
+  # banner line for a clean shutdown.
+  verify_greeting = false;
+  send_quit = false;
+
+  # Cache TTLs (seconds).
+  expire = 86400;          # successful outcomes
+  expire_novalid = 7200;   # hard failures (refused, invalid, nxdomain, null, broken, error)
+  expire_timeout = 1800;   # transient timeouts (so recovery surfaces quickly)
+
+  # Force-reject behaviour. Defaults are off; flip these only after
+  # auditing your operator scores and DNS reliability.
+  reject_nxdomain = false;
+  reject_null_mx = false;
+
+  # Symbol names (primary surface; today's behaviour preserved).
   symbol_bad_mx = "MX_INVALID";
-  # symbol yielded if no MX is found
   symbol_no_mx = "MX_MISSING";
-  # symbol yielded if MX is connectable
   symbol_good_mx = "MX_GOOD";
-  # lifetime of redis cache - 1 day by default
-  expire = 86400;
-  # prefix used for redis key
+  symbol_white_mx = "MX_WHITE";
+
+  # Finer outcome symbols (Phase A: score 0, informational).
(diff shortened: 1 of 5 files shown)


vstakhov:
@droid review

PR #6050, opened 2026-05-21
[Fix] autolearnstats: fix table formatting crash and add sorting/grouping options

moisseev:
- Replaces LuaJIT `string.format` with custom `pad()` and `cell()` helpers to fix a crash on table column widths $\ge$ 100 and truncate overlong values. 
- Introduces `--sort-by` and `--group` CLI options to better organize output rows.


moisseev:
@droid review

PR #6055, opened 2026-05-22
[Feature] mx_check: three-layer cache rewrite

dragoangel:
This is the comprehensive implementation behind issue #6032. The single-layer cache from the previous shape is replaced by a three-layer Redis design (`d:<domain>` / `m:<mxhost>` / `i:<ip>`) under `<key_prefix>:`. Short-code wire formats minimise Redis footprint; per-layer validators (`is_valid_cache_value`) treat unrecognised entries as a cache miss; the resolve / probe path that follows then issues a fresh `cache_set` at the same key, overwriting the bad entry in place.

## Probe coordination

- `SET NX EX` claims the `i:<ip>` probe lock; a post-claim `GET` disambiguates held lock, already-published verdict, and corrupted-value-needing-heal cases. A separate `force_claim_probe_lock` path overwrites corrupted values to break the SET NX loop without leaking refcounts.
- Redis errors during the lock claim surface as `MX_REDIS_ERROR`; lock held by another worker surfaces as `MX_INFLIGHT` and skips duplicated TCP connections which under high load would result in DoS-like activity from the target side and most likely negatively impact rspamd's user IP/ASN/Org reputation.

## DNS / probe model

- Dual-stack via `probe_ipv4` / `probe_ipv6` / `prefer_ipv6` with family-tagged cache values (`v4:` / `v6:` / `v64:`) and coverage checks so flipping the probe-family set re-resolves only as needed.
- Real DNS path failures (SERVFAIL / REFUSED / timeout) are distinguished from authoritative NXDOMAIN / NOREC via `is_dns_real_failure`; the former surface as `MX_DNS_FAIL` (cached as `df`) so a recovered resolver path can be re-tried promptly. NXDOMAIN / NOREC collapse into `MX_NONE`.
- step3 partitions resolved IPs into PUBLIC / LOCAL (RFC1918 / CGNAT / ULA) / BOGON (loopback, TEST-NET, multicast, link-local, etc.). Only PUBLIC IPs reach the TCP probe. `MX_LOCAL_ONLY` / `MX_LOCAL_MIX` / `MX_BOGON_ONLY` / `MX_BOGON_MIX` fire with the offending IPs as options. `test_mode` lifts loopback out of the bogon set so the probe path can be exercised against 127.0.0.1.

## Symbol surface

- Multi-source: `check_from` / `check_mime_from` / `check_reply_to` with envelope > reply-to > mime-from priority dedup when the same domain is hitting MX checks from different sources. Per-source prefixes (`symbol_prefix_from` / `symbol_prefix_mime_from` / `symbol_prefix_reply_to`) fan every `MX_*` symbol across the three sources at registration time.
- A-fallback path (no MX RR, A used as implicit MX per RFC 5321 §5.1) has its own `MX_A_*` symbol family so operators can score it independently of the MX-RR path.
- Per-outcome greylist and reject gates (`greylist_invalid` / `greylist_none` / `greylist_broken` / ..., `reject_null_mx` with `reject_authorized` / `reject_local` kill switches); null-MX domains can now trigger a real `set_pre_result`. `reject_nxdomain_mx` removed as a bad option to serve — practically NXDOMAIN reject only makes sense at eTLD+1 granularity.
- Probe-outcome symbols (`MX_GOOD` / `MX_TIMEOUT_*` / `MX_REFUSED` / `MX_INVALID` / `MX_ERROR` / `MX_INFLIGHT`) populate the option field with the MX hostname; IP-class symbols still carry IPs since that's where IP information is the point. `MX_REDIS_ERROR` has no option (it's a module-internal signal).
- New punishment maps: `bad_mxs` (glob on MX hostnames) and `bad_ips` (radix on resolved IPs). Any hit short-circuits with `MX_BAD` / `MX_IP_BAD` before any TCP probe runs, which allows penalising domains that share the same MX infra.
- **Per-entry weight multiplier for `bad_mxs` / `bad_ips`**: an optional numeric token after the pattern is parsed via `tonumber` and used as a weight multiplier on top of the symbol's group score. `trapmx.example.com 3` triples the `MX_BAD` weight for that hostname; `1.2.3.4 0.5` halves the `MX_IP_BAD` weight for that IP. Default multiplier is `1.0` (no value, or non-numeric value). Lets operators tier confidence within a single map instead of maintaining several.

## Scoring

- `set_metric_all_sources` ships sensible defaults for every symbol. Operators can tune any weight through the new `mx` group in `conf/groups.conf` via `local.d/mx_group.conf` or `override.d/mx_group.conf` without touching the module.

## Functional tests

- `167_mx_check.robot` refreshed for the new symbol set; `MX_NONE` replaces `MX_NXDOMAIN`/`MX_MISSING`, `MX_A_REFUSED` covers the closed-port A-fallback case, and `MX_BAD` / `MX_IP_BAD` have dedicated assertions.
- `168_mx_check_greeting.robot` covers `verify_greeting=true` / `send_quit=false`: silent listener → `MX_TIMEOUT_READ`; continuation `220-` with no follow-up held past `read_timeout` → `MX_GOOD` (a regression that re-queued reads under `send_quit=false` would surface as `MX_TIMEOUT_READ`); 5xx greeting → `MX_ERROR`; non-SMTP line → `MX_INVALID`.
- `169_mx_check_greeting_quit.robot` covers `verify_greeting=true` / `send_quit=true`: proper multi-line timing → `MX_GOOD` plus dummy status file `QUIT_AFTER_FINAL` (catches a regression where QUIT is sent before the final `220 ` line, which rspamd's verdict alone cannot detect); slow second line → `MX_TIMEOUT_READ`.
- `util/dummy_smtp.py` mock with `silent` / `error` / `messy` / `greeting_single` / `greeting_multi` modes and a `--status-file` argument for out-of-band timing verification.


vstakhov:
@review

vstakhov (changes requested):
Reviewed the full 1847-line `mx_check.lua`, both new config files, all three Robot suites, and the Python mock. Overall a substantial and well-structured rewrite — the lock protocol, the family-tagged cache, and the test design (especially the out-of-band status-file timing verification) are all above the median for module changes. One finding is a real defect with a concrete fix; the rest are suggestions.

## Findings

### 🔴 Defect — envelope-from variants are not registered as virtual children of the mx_check parent

`src/plugins/lua/mx_check.lua:1671-1682`:

```lua
local function register_all_sources(base_name)
  local prefixes = {
    settings.symbol_prefix_from,        -- ""        (envelope-from)
    settings.symbol_prefix_mime_from,   -- "MIME_FROM_"
    settings.symbol_prefix_reply_to,    -- "REPLYTO_"
  }
  for i, prefix in ipairs(prefixes) do
    if i ~= 1 then
      rspamd_config:register_symbol({ name = prefix .. base_name, type = 'virtual', parent = id })
    end
  end
end
```

The `i ~= 1` guard skips the envelope-from variant **for every symbol**, but the rationale (skip the callback-parent to avoid double registration) is only valid for `symbol_bad_mx` — that's the only symbol whose envelope-from variant *is* the callback parent (registered at line 1661 as `name = settings.symbol_prefix_from .. settings.symbol_bad_mx = "" .. "MX_INVALID"`).

For every other call (`register_all_sources(symbol_good_mx)`, `symbol_white_mx`, `symbol_mx_refused`, `symbol_mx_timeout_*`, `symbol_mx_error`, `symbol_mx_none`, `symbol_mx_null`, `symbol_mx_broken`, `symbol_mx_dns_fail`, `symbol_mx_local_*`, `symbol_mx_bogon_*`, `symbol_mx_skip`, `symbol_mx_bad`, `symbol_mx_ip_bad`, `symbol_mx_inflight`, `symbol_mx_redis_error`, all six `symbol_mx_a_*`) the envelope-from variant — `MX_GOOD`, `MX_WHITE`, `MX_NONE`, `MX_BAD`, `MX_IP_BAD`, `MX_A_GOOD`, … — is **never registered** as a child of the mx_check parent.

**Why the tests still pass:** `task:insert_result` tolerates unknown symbol names and pairs them with whatever `set_metric_symbol` defined for that name. So the symbol *appears* in the result and the `Expect Symbol MX_GOOD` assertions pass.

**What breaks silently:**
- `symbols_enabled = [...]` filters and group-level enable/disable don't include the envelope-from variants in the mx_check parent's scope.
- Symbol metadata (description, group binding) is decoupled from the callback parent.
- Symbol dependency declarations against e.g. `MX_GOOD` can't find a registered symbol.
- `rspamc -d` symbol listing shows them as orphans rather than virtual children of `MX_INVALID`.

**Fix:** parametrize the skip:

```lua
local function register_all_sources(base_name, skip_envelope)
  local prefixes = {
    settings.symbol_prefix_from,
    settings.symbol_prefix_mime_from,
    settings.symbol_prefix_reply_to,
  }
  for i, prefix in ipairs(prefixes) do
    if not (skip_envelope and i == 1) then
      rspamd_config:register_symbol({ name = prefix .. base_name, type = 'virtual', parent = id })
    end
  end
end

register_all_sources(settings.symbol_bad_mx, true)  -- envelope-from variant is parent
register_all_sources(settings.symbol_good_mx)        -- register all 3 variants
register_all_sources(settings.symbol_white_mx)
-- ...all other calls unchanged...
```

---

### 🟡 Minor — `bad_mxs` / `bad_ips` weight multiplier accepts negative numbers

`mx_check.lua:1081` and `:1206`:

```lua
local mult = (type(m) == 'string') and tonumber(m) or nil
done('bad_ip', { key = ip, weight_mult = mult })
```

`tonumber("-5")` returns `-5`, which inverts the punishment into a reward in `emit_outcome`:

```lua
task:insert_result(p .. settings.symbol_mx_ip_bad,
  info.weight_mult or 1.0, info.key or mx_domain)
```

A misconfigured map entry like `1.2.3.4 -3` silently turns `MX_IP_BAD` into a negative-score symbol. Suggest clamping:

```lua
local mult = (type(m) == 'string') and tonumber(m) or nil
if mult and mult < 0 then mult = nil end  -- punishment maps must not emit negative weights
```

---

### 🟡 Minor — IP-class symbols fire alongside subsequent short-circuits

In `step3` (`mx_check.lua:1062-1085`):

```lua
if has_local or has_bogon then
  emit_ip_class_symbols(task, mx_domain, local_ips, bogon_ips, has_public, src)  -- (1)
end
if not has_public then
  done('ip_class_skipped', ctx); return
end
if bad_ips then
  for _, ip in ipairs(public_ips) do
    local m = bad_ips:get_key(ip)
    if m then
      done('bad_ip', { ... }); return                                              -- (2)
    end
  end
end
```

If a domain has LOCAL + PUBLIC mix **and** the public IP is in `bad_ips`, both `MX_LOCAL_MIX` (1) **and** `MX_IP_BAD` (2) fire. Same with `exclude_ips` → `MX_LOCAL_MIX` + `MX_SKIP`. Default scores would sum to 9.0 in the bad_ips case (3.0 + 6.0), which may not be the intended semantics for `MX_BAD` / `MX_IP_BAD` described in the PR description as "short-circuits before any TCP probe runs". If short-circuit was meant to suppress everything past it, move the IP-class emission after the bad/exclude checks. If both signals are intentional (operators see the full shape), please make that explicit in a code comment so a future reader doesn't "fix" it.

---

### 🟡 Minor — `step3` returns the first cached verdict, even if a later IP has a better one

`mx_check.lua:1158-1188` walks `ips[]` and returns the first cache hit, including negative ones (`tc`, `rf`, `tr`, `inv`). If `ips = [A, B]` with `cache[A] = 'tc'` but `cache[B] = 'gd'` (the host moved its preferred IP), the worker reports `tc` until A's cache expires (`expire_timeout = 2h` by default). This is a deterministic / Redis-traffic-friendly choice, but in practice the cache becomes self-reinforcing: as long as the slot has *any* verdict, B is never consulted.

Not a defect — defensible — but worth a sentence in the docstring noting that the first cached IP wins, and operators should size `expire_timeout` accordingly.

---

## Positives worth calling out

- **The lock protocol (`SET NX EX` + post-claim `GET` + `force_claim` heal path)** is genuinely well thought out. The state diagram in the comments at lines 413–420 documents the four post-claim cases cleanly and the code implements all four.
- **Cache-value validation that treats unknown entries as a miss + lets the natural resolve/probe path overwrite the bad value** is elegant — avoids a separate `DEL` write op and is migration-safe across schema bumps.
- **Family-tagged IP-list cache values (`v4:` / `v6:` / `v64:`) with `family_coverage_ok`** correctly handles the "flip `probe_ipv6` on" case without staleness — without the tag, you'd either spuriously re-resolve every cached entry or serve stale incomplete data.
- **`emit_ip_class_symbols` separating `_MIX` / `_ONLY`** lets operators score differently for "publisher mistakenly listed a private IP alongside the real one" vs "publisher's entire MX setup is unreachable from outside".
- **A-fallback `MX_A_*` symbol family** is the right call — operationally a no-MX-but-A-listening domain is a different signal than an MX-RR domain that responds.
- **`is_dns_real_failure` distinguishing SERVFAIL/timeout from NXDOMAIN/NOREC** (via `MX_DNS_FAIL` at score 0.0) prevents penalizing senders for *our* resolver path issues.
- **The test mock's status-file timing verification** in `dummy_smtp.py` is the right way to test QUIT ordering — the verdict alone cannot tell you whether QUIT landed before or after the final banner line. The `OK_NO_QUIT` / `QUIT_AFTER_FINAL` / `QUIT_BEFORE_FINAL` / `EOF_BEFORE_FINAL` distinction catches regressions the rspamd-side assertion cannot.
- **The `220-` continuation regression test** at line 31-39 of `168_mx_check_greeting.robot` is a clever asymmetric test: a single `220-` (continuation marker) with no follow-up should still fire `MX_GOOD` under `send_quit=false`. A future change that always waits for `220 ` (with space) would surface here as `MX_TIMEOUT_READ`.

---

## Summary

- **Required:** Fix `register_all_sources` so the envelope-from variants of every symbol other than `symbol_bad_mx` are properly registered as virtual children of the mx_check parent.
- **Recommended:** Clamp negative `weight_mult` to `nil`/`1.0` in `step2` (bad_mxs) and `step3` (bad_ips).
- **Recommended:** Clarify the intended semantics of IP-class symbols coexisting with `MX_BAD` / `MX_IP_BAD` / `MX_SKIP` short-circuits — in a code comment regardless of which way you decide.
- **Optional:** One-line note about first-cached-IP-wins semantics in `step3` so operators understand how `expire_timeout` interacts with backend address changes.

dragoangel:
Thank you for review, I will check points today at evening or tomorrow and will fix them

dragoangel:
- ⛔ Defect — envelope-from variants are not registered as virtual children of the mx_check parent - real issue, will fix.
- ✅ Minor — IP-class symbols fire alongside subsequent short-circuits - this is by design. MIX symbols fire always when found, and breaking it by bad ips - would drop details, as operator would expect to punish bad ips by listing public subnets, not local/bogon subnets that already in place. Not a bug. If user wants to allow broken MX - exlude_mxs MUST be used - which totally drops Bogons/Local checks or exlude_domains which drops even at higher level. Design is clear from my view.
- ✅ Minor — step3 returns the first cached verdict, even if a later IP has a better one. This also not a bug and is by design - "better is by MX prio" in MX checker is not the case, better to have less probes - and by fully utilize cached probes, if domain X has MX01 and MX02 - we not have any cache - we miss both MX01/MX02 get_lookups and issue probe for MX01 only once, but another domain Y - has same IPs behind his MXs - but IPs prio is swapped (loadbalancing as example), we check both IPs from cache - find one with lower prio and use results without probing anything, less TCP probes - less complains.
- ✅ Minor — bad_mxs / bad_ips weight multiplier accepts negative numbers - tbh I intentionally leave this this way, not see issue in "putting negative scores", yes it would mess up wording - we can rename to `mxs_multiplayer` but I can't get good pharasing for it, @vstakhov I would glad to rename it if you have ideas. If user want put HAM score for "bad MX" - let him do it, why not - he can changer "BAD MX/IPS" symbol to "MODIFIER" in his setup. About proposed fix with forcing `1.0` or totally dropping records with negative score would be messed-up logic from operator standpoint - no Rspamd module had such logic with multiplayers in any map.

dragoangel:
@vstakhov can you please trigger rerun of CI/CD - I fixed lack of executable flag and resolved symbols registration, expect that now tests should pass ☺️

dragoangel:
```
All three mx_check suites PASSED on ubuntu_amd64:
Cases.167 Mx Check — PASSED in 4.8s
Cases.168 Mx Check Greeting — PASSED in 6.7s
Cases.169 Mx Check Greeting Quit — PASSED in 6.4s
```
All mx check tests passed. What failed:
```
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0870714Z Controller SSL - stat :: Fetch /stat over HTTPS from the controlle... | FAIL |
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0872579Z Controller SSL - errors :: Fetch /errors over HTTPS from the contr... | FAIL |
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0880897Z Cases.001 Merged.440 Ssl Server                                       | FAIL |
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0881475Z 6 tests, 4 passed, 2 failed
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0882577Z Cases.001 Merged                                                      | FAIL |
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0883186Z 6 tests, 4 passed, 2 failed
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0884231Z Cases                                                                 | FAIL |
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.0884788Z 6 tests, 4 passed, 2 failed
ubuntu_amd64 / test  UNKNOWN STEP  2026-05-26T08:16:28.1110090Z 674 tests, 672 passed, 2 failed, 0 skipped.
```

PR #6061, opened 2026-05-24
[Feature] Add fixed-point formatting to fpconv

moisseev:
- Add FPCONV_PRECISION_ALL sentinel for trim-trailing-zeros mode with compile-time guard (static_assert > 17 significant digits)
- Implement %.Nf rounding with carry (round_at, trim_trailing_zeros)
- Fix %.0f carry detection for numbers like 9.9 -> 10
- %f/%F/%g/%G use FPCONV_PRECISION_ALL instead of hardcoded literals
- Add C++ unit tests for fpconv precision and rounding

### DIFF
--- a/.github/workflows/ci_rspamd.yml
+++ b/.github/workflows/ci_rspamd.yml
@@ -78,17 +78,84 @@ jobs:
         run: |
           sudo mv /usr/bin/miltertest /usr/bin/miltertest.is.broken.on.fedora || true
 
+      - name: Install pabot
+        # --break-system-packages: Ubuntu 24.04 ships Python 3.12 with PEP 668
+        # marking the system interpreter as externally-managed. Older pip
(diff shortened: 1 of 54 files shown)


moisseev:
@review

moisseev:
@review

moisseev:
@review

vstakhov:
It looks good in general but I have some gut feeling that we should not touch the original fpconv code, as it's a 3-rd party library. Can we probably make a shim layer to add fixed point formatting out of the number produced by girsu2?

moisseev:
@review

moisseev:
@review

vstakhov:
I think that since this shim is actually managed code, it should be somewhere in libutil and not in contrib. Sorry for not mentioning that earlier.

moisseev:
@review

Issue #6092, opened 2026-06-09
[Feature] Improve task_timeout vs. symbols cache timeout warning


phonon112358:
## Summary

In Rspamd 4.1.0, the warning `configured task_timeout X is less than maximum symbols cache timeout Y` calculates the threshold by summing all symbol timeouts sequentially. Since most symbols run in parallel, this produces an unrealistically high value.

## Motivation

Example (Rspamd 4.1.0):
```
configured task_timeout 30.00 is less than maximum symbols cache timeout 40.00
top symbols: CLAM_VIRUS(15.00); GREYLIST_CHECK(5.00); RATELIMIT_CHECK(5.00);
GREYLIST_SAVE(5.00); HISTORY_SAVE(5.00); RATELIMIT_UPDATE(5.00)
```
The 40s is a sequential sum, but these symbols execute concurrently. Raising `task_timeout` to 40s to silence the warning seems excessive for a single hung ClamAV scan.

## Ideas for Resolution

Two possible fixes:
- Calculate based on parallel execution groups rather than sequential sum
- Or at minimum clarify the message: *"theoretical maximum assuming sequential execution"*

moisseev:
**Root cause:** In `symcache_impl.cxx`, the `pre_postfilter_iter` lambda (used for prefilters, postfilters, and idempotent filters) initializes `saved_priority` to `-1`. Since `init()` sorts these vectors
by priority (values ≥ 0), the first item in each phase vector always triggers a false "priority change" on the second iteration, splitting it into its own single-item group. This causes its timeout to be
counted individually instead of being grouped with items at the same priority.

**Important:** The algorithm does **not** simply sum all symbol timeouts sequentially. It already accounts for parallelism within each phase:
- **Prefilters / postfilters / idempotent:** grouped by priority — only the **maximum** timeout per priority group is taken.
- **Normal filters:** the **longest dependency chain** is taken (not the sum of all filters).
- Phases themselves (prefilters → filters → postfilters → idempotent) are summed, since they execute sequentially.

The issue was a bug in the grouping logic that prevented it from working correctly for the first item in each phase.

**In your example:**

| Phase | Symbols | Priority | Buggy result | Correct result |
|---|---|---|---|---|
| Prefilters | `GREYLIST_CHECK(5)`, `RATELIMIT_CHECK(5)` | both 5 | 5 + 5 = **10** | max(5, 5) = **5** |
| Filters | `CLAM_VIRUS(15)` | — | 15 | 15 |
[...]

PR #6094, opened 2026-06-10
[Fix] symcache: fix timeout inflation in pre_postfilter_iter grouping

moisseev:
The `saved_priority` initialization to -1 caused the first item in each phase vector (prefilters/postfilters/idempotent) to be split off from its priority group and counted individually, inflating the computed maximum symbols cache timeout. Initialize to the first item's priority instead so items at the same priority are correctly grouped.

Issue: #6092


moisseev:
@review

PR #6114, opened 2026-06-24
[FEATURE]: src/rspamadm: allow reading password from env

stefanbenten:
Currently the CLI only allows reading the password from tty. This makes it challenging to check the password via automation, ie. used by continous deployment to ensure a clean state.

This change adds the functionality to read from env, so that passwords can be safely passed in and checked.

### DIFF
--- a/src/rspamadm/pw.c
+++ b/src/rspamadm/pw.c
@@ -269,8 +269,21 @@ rspamadm_pw_check(void)
 			exit(EXIT_FAILURE);
 		}
 
-		plen = rspamd_read_passphrase(test_password, sizeof(test_password),
-									  0, NULL);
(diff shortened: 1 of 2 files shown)


vstakhov:
I think it's better to use environment or .env for that. On one hand, stdin is at least history safe (so password won't be stored in history), on the other hand it promotes bad practices like hardcoding passwords in scripts.

stefanbenten:
The idea is not to hardcode passwords anywhere, but being able to use rspamadm in IaC pipelines to ensure the currently deployed password is matching the desired safely stored one during deployments.
Either way, i have adjusted to use ENVs, for my purpose both ways work just fine.

Issue #5463, opened 2025-05-10
[Feature] Find a FooTable replacement for the WebUI


jeromelebleu:
## Summary

FooTable seems unmaintained since 2017 and could be replaced by another recent library.

## Motivation

FooTable is probably doing the job fine, but unfortunately it is almost the last part of the WebUI which is still using Glyphicons - see #5364. The removal of Glyphicons will conclude the FontAwesome transition - and will lighten the page loading by the way.

## Describe alternatives you've considered

It seems there is at least those two alternatives:
* Bootstrap Table: https://bootstrap-table.com/
* Grid.js: https://gridjs.io/

Any feedback, other alternatives and preferences are welcome to choose!

moisseev:
Though Glyphicons has already been replaced with FontAwesome via https://github.com/rspamd/rspamd/pull/5702, I'd like to reopen this issue as it remains relevant. FooTable is still unmaintained (last updated in 2017), and replacing it would help:

1. **Improve maintainability** - Using an actively maintained library will provide better long-term support, security updates, and compatibility with modern browsers
2. **Modernize the WebUI** - FooTable's age (8+ years without updates) means we're missing out on improvements in table libraries, including better responsive design, accessibility features, and performance optimizations
3. **Reduce technical debt** - Continuing to rely on an abandoned dependency increases risk for future maintenance

PR #6120, opened 2026-07-02
WebUI: migrate from FooTable to Tabulator

moisseev:
## Summary

Replaces the unmaintained [FooTable](https://fooplugins.github.io/FooTable/) library with [Tabulator](https://tabulator.info/) 6.5 across all WebUI data tables, then removes FooTable entirely.

Ref #5463 — find a maintained FooTable replacement for the WebUI.

## Motivation

FooTable has been unmaintained since 2017 and was the last WebUI component still depending on bundled Glyphicons assets (#5364). Dropping it removes the final Glyphicon dependency and completes the
FontAwesome transition, while removing one legacy JS/CSS bundle from page load.

## What changed

All five FooTable-driven tables are rebuilt on Tabulator, migrated one at a time:

- **Errors log** (`history.js`)
- **Throughput / rrd-table** (`graph.js`)
- **Symbols** (`symbols.js`)
- **History + Scan** (`history.js` / `upload.js`, shared via `libft.js`) — the most complex: action filter with negation, column-options dropdown (visible / hidden / row-mode), per-column responsive
collapse, and symbol rendering with fuzzy-hash buttons.

FooTable is then removed wholesale — `footable.min.js`, `footable.standalone.min.css`, `FooTable.FontAwesome.css`, and the `footable-fontawesome` module are deleted, along with every `<link>`/`require`
reference and the now-dead `.footable` CSS.

## Notes

- Tabulator 6.5 vendored at `interface/js/lib/tabulator.min.js` and `interface/css/tabulator_bs5.min.css`.
- Responsive layout uses Tabulator's `collapse` mode, preserving the existing UX (columns collapse in order; a row click expands the detail view).
- Cross-table UI helpers (single-page footer hiding, scroll preservation, click-to-toggle rows) live in a new shared `js/app/tab-utils.js`.
- History/Scan keep a query-language global search — `foo bar` (AND), `OR`, `-neg`, `"phrases"` — ANDed with per-column header filters.
- Light/dark themes follow automatically: Tabulator's hardcoded BS5 colours are remapped to Bootstrap CSS variables that flip with `data-bs-theme`.
- The legacy history path is retained for older servers that don't report a `version` field.
- The recipients ("To/Cc/Bcc") column gained a per-column filter that matches against the full recipient list, not the truncated cell value.


moisseev:
@review

moisseev:
@review

PR #6124, opened 2026-07-05
[Rework] Remove jQuery from the WebUI

moisseev:
Remove jQuery from the WebUI entirely: the vendored `jquery-3.7.1.min.js`,
the `jquery.stickytabs` plugin, and the Bootstrap 5 jQuery bridge are all
gone. Every app module now loads without jQuery.

The work is split into reviewable commits along the dependency graph
(common.js first, since everything imports it):

1. Vendored chart libs — D3Evolution 2.0.4→2.1.0 and D3Pie 1.1.2→1.2.0,
   which dropped their own `$.extend` usage; the stale jQuery shim deps
   for them are removed.
2. Sticky tabs — the `jquery.stickytabs` plugin is replaced with an inline
   vanilla implementation on the BS5 Tab API. URL-hash persistence and the
   initial-tab fallback are preserved.
3. BS5 bridge — imperative `.modal()/.popover()/.trigger("click.bs.dropdown")`
   become `bootstrap.{Modal,Popover,Dropdown}.getOrCreateInstance(...)`.
4. AJAX spine (`common.js`) — `query`/`queryServer` rewritten on
   `XMLHttpRequest`, preserving the neighbours fan-out/aggregation,
   per-server error dispatch, `statusCode`, timeout, jqXHR-compatible
   callback arguments, and the NProgress download-progress hook. Adds the
   shared vanilla helpers (`el`, `delegate`, `data`, `onAjaxStart/Complete`,
   `setAjaxTimeout`) reused by the rest of the migration.
5. Leaf modules (`graph`, `selectors`, `history`, `symbols`).
6. `config`, `upload`, `stats`, `main` (rspamd.js), `libft` — one commit
   each; see the commit messages for details.
7. Final — drop jQuery from the RequireJS config (paths, app/bootstrap
   shims) and delete `jquery-3.7.1.min.js`. Drive-by: `data-backdrop` →
   `data-bs-backdrop` on `#modalDialog` (the BS4 attribute name was
   silently inert under BS5).

Notable translation choices:
- jQuery's delegated `mouseenter`/`mouseleave` (which don't bubble
  natively) is simulated on `mouseover`/`mouseout` with a
  `relatedTarget`-containment check (`stats.js`).
- `$.Deferred` / `$.when().always` → `Promise` / `Promise.all().finally`.
- jQuery event namespaces (`hidden.bs.dropdown.rspamd`) → a
  `common.data`-tracked `addEventListener` that removes the previous
  handler before rebinding.
- The `/stat` connect probe and the legacy `/auth` fan-out keep a raw XHR
  to avoid `common.query`'s automatic error logging on their expected
  non-2xx paths.


moisseev:
@review

interface/js/app/stats.js
```diff
@@ -267,20 +337,22 @@ define(["jquery", "app/common", "d3pie", "d3"],
 
             function addFuzzyStorage(server, storages) {
                 let i = 0;
-                $.each(storages, (storage, hashes) => {
+                const fuzzyTbody = document.querySelector("#fuzzyTable tbody");
+                Object.entries(storages).forEach(([storage, hashes]) => {
```
vstakhov: **Must fix:** `Object.entries(storages)` throws `TypeError` when the server's stat data has no `fuzzy_hashes` key — the old `$.each(undefined, …)` was a silent no-op. The key is genuinely optional: the controller only emits it when fuzzy_check recorded stats, and an unreachable neighbour contributes `data = {}`. Any install with fuzzy_check disabled — or a cluster with one node down — throws on every Status refresh inside the `Promise.all(...).finally()` callback: the bayes/fuzzy tables render partially and the pie chart never draws.

`addStatfiles` above got exactly this guard (`Array.isArray` at line 305); this needs the same, e.g. `Object.entries(storages || {})` (and `Object.keys(storages || {})` for the rowspan below).

interface/js/app/common.js
```diff
@@ -456,45 +688,65 @@ define(["jquery", "nprogress"],
...
-                $el.slideUp(400, function () {
-                    $(this).addClass("d-none");
-                });
-            } else {
-                $el.addClass("d-none");
+            for (const el of toElements(selector)) {
+                if (anim) {
+                    const height = el.offsetHeight;
+                    el.style.overflow = "hidden";
+                    el.animate(
```
vstakhov: **Must fix:** these animations use `fill: "forwards"` and are never cancelled. A forwards-filling animation keeps applying its end value indefinitely — clearing `el.style.height` in `onfinish` doesn't remove it. Same issue in `ui.show` below.

Concrete failure: close the fuzzy-hashes card via its `.card-close-btn` (upload.js:198, animated hide → fill holds `height: 0`), then run another hash lookup — `common.show("#hash-card", true)` (upload.js:151) removes `d-none`, measures `offsetHeight` **while the old fill still forces height 0**, and animates 0→0. The card never reappears. Even a lone animated `show` pins the element at the measured pixel height, so content that grows later gets clipped.

Fix: capture the animation and cancel it once the final state is committed:
```js
const anim = el.animate([...], {...});
anim.onfinish = () => {
    el.classList.add("d-none");
    el.style.height = "";
    el.style.overflow = "";
    anim.cancel();
};
```
(and the equivalent in `show`).

interface/js/app/common.js
```diff
@@ -40,20 +38,163 @@ define(["jquery", "nprogress"],
...
+
+        // Resolve a string selector, Element, NodeList, Array, or jQuery wrapper
+        // into a plain Array of Element. Accepts jQuery objects during the staged
+        // removal so callers still passing $(...) keep working.
+        function toElements(selector) {
+            if (!selector) return [];
+            if (typeof selector === "string") {
+                return Array.from(document.querySelectorAll(selector));
+            }
+            if (typeof selector.length === "number") {
```
vstakhov, line 100: Latent trap: any Element with a numeric `.length` is treated as a collection here — `HTMLSelectElement` and `HTMLFormElement` both have one, so passing a select/form to `hide`/`show`/`toggle` operates on its options/controls instead of the element. There's already one masked instance: `common.show(pickerEl)` at upload.js:538 un-hides the `<option>` rather than the `<select>` (harmless today only because `toggleWidgets(true, …)` at :531 already showed it).

Check `selector instanceof Element` (or `nodeType === 1`) before the `length` duck-test.

interface/js/app/rspamd.js
```diff
@@ -500,29 +594,28 @@ define(["jquery", "app/common", "stickytabs", "visibility",
...
+        checked_server = e.currentTarget.value;
         if (checked_server === "All SERVERS") {
-            $("#learnServers").removeClass("invisible");
+            document.getElementById("learnServers").classList.remove("invisible");
         } else {
-            $("#learnServers").addClass("invisible");
+            document.getElementById("learnServers").classList.add("invisible");
         }
-        tabClick("#" + $("#tablist > .nav-item > .nav-link.active").attr("id"));
+        tabClick("#" + document.querySelector("#tablist > .nav-item > .nav-link.active").id);
```
vstakhov: `querySelector(...)` returns `null` when no tab is active, and this throws — the old `.attr("id")` returned `undefined` and degraded to a harmless `tabClick("#undefined")` no-op. Reachable: open the UI with a stale/foreign hash (e.g. `/#foo`) so `initStickyTabs` activates nothing, log in, then switch the server in `#selSrv` → uncaught TypeError kills the change handler. Same pattern at line 122. Needs a null guard.

interface/js/app/common.js
```diff
@@ -403,7 +638,7 @@ define(["jquery", "nprogress"],
                 queryServer(neighbours_status, 0, "neighbours", {
                     success: function (json) {
                         const [{data}] = json;
-                        if (jQuery.isEmptyObject(data)) {
+                        if (Object.keys(data).length === 0) {
```
vstakhov: `Object.keys(data)` throws if `data` is `null` — `parseBody` returns `null` for an empty 200 body with a JSON content-type, and the old `jQuery.isEmptyObject(null)` returned `true`. Cheap guard: `if (!data || Object.keys(data).length === 0)`.

interface/js/app/rspamd.js
```diff
@@ -251,87 +378,37 @@ define(["jquery", "app/common", "stickytabs", "visibility",
...
-        $.ajax({
-            type: "GET",
-            url: "stat",
-            success: function (data) {
+        const xhr = new XMLHttpRequest();
+        xhr.open("GET", "stat", true);
+        const ajaxTimeout = common.getAjaxTimeout();
+        if (ajaxTimeout > 0) xhr.timeout = ajaxTimeout;
+        xhr.onload = () => {
+            if (xhr.status >= 200 && xhr.status < 300) {
```
vstakhov, line 376: Nit: the probe treats 304 as failure, while your own `queryServer` deliberately keeps `|| xhr.status === 304` (common.js:524). Practically near-unreachable (browsers surface revalidation as 200), but internally inconsistent — same in the legacy `/auth` fan-out (stats.js:487).

interface/js/app/rspamd.js
```diff
@@ -207,41 +258,117 @@ define(["jquery", "app/common", "stickytabs", "visibility",
...
             },
             errorMessage: "Cannot get server status",
             server: "All SERVERS"
         });
     }
 
 
+    // Parse JSON, returning {} for an empty or non-JSON body (the /stat probe
+    // must not throw and strand the connect flow).
+    function parseJsonSafe(text) {
```
vstakhov: Nit: a 200 `/stat` response with malformed JSON now silently becomes `{}` and proceeds to `displayUI()` — you get an unauthenticated, empty main UI with "Cannot get server status" alerts. The old content-type-driven parse raised `parsererror` → error path → login form. Consider treating a parse failure of a `Content-Type: application/json` body as probe failure instead of swallowing it.

interface/js/app/upload.js
```diff
@@ -14,41 +14,42 @@ define(["jquery", "app/common", "app/libft"],
...
-                        common.alertMessage("alert-info", jqXHR.statusText);
-                    }
-                    deferred.resolve();
-                },
-                complete: () => deferred.resolve(),
-                server: server()
+                common.query(url, {
+                    data: data,
+                    params: {
+                        processData: false,
```
vstakhov: Nit: `params: {processData: false}` is dead now — the new `queryServer` ignores it (the body is already a string). Same at lines 60 and 157; worth removing so it doesn't imply it does something.

interface/js/app/common.js
```diff
@@ -221,100 +363,191 @@ define(["jquery", "nprogress"],
...
+                    return xhr.responseText ? JSON.parse(xhr.responseText) : null;
+                }
+                return xhr.responseText;
+            }
+
+            // jQuery fires statusCode handlers once per request for the final
+            // status, in addition to success/error.
+            function runStatusCode() {
+                if (o.statusCode && typeof o.statusCode[xhr.status] === "function") {
+                    o.statusCode[xhr.status](xhr.responseText, xhr.statusText, xhr);
```
vstakhov, line 465: Nit: for error statuses jQuery passed `(jqXHR, textStatus, errorThrown)` to `statusCode` handlers, not `(responseText, statusText, xhr)`. All current callers (upload.js) take no arguments, so nothing breaks — just document the new contract here so future callers don't assume jQuery's shape.

vstakhov (changes requested):
Impressive work — the staged commit structure is exemplary, and I traced the whole diff old→new (including the subtle spots: the sticky-tabs rewrite, the `hidden.bs.dropdown` rebind and fuzzy-button delegation cleanup in libft.js, the mouseenter/mouseleave simulation in stats.js, `$.when().always` → `Promise.all().finally`, and both vendored chart-lib bumps). The overwhelming majority of the translation is faithful, and ESLint/stylelint pass. The Playwright race fix is correct: `Promise.all` registers both listeners before the click.

Requesting changes for two real regressions (inline): the `fuzzy_hashes` `TypeError` in stats.js that breaks the Status tab for any install with fuzzy_check disabled or a neighbour down, and the fill-forwards animation pinning in `common.hide`/`common.show` that leaves the fuzzy-hashes card permanently collapsed after it's closed once. The rest of the inline comments are smaller hardening/consistency points.

Noted, no action needed:
- The `data-bs-backdrop` drive-by is a user-visible change (map modal no longer closes on outside click), but it's declared in the PR description as restoring the markup's intent — fine by me.
- `return false` → `preventDefault()` in selectors.js drops `stopPropagation`, so clicks on those controls now bubble to the body-level popover-dismiss handler and close an open settings popover — matches that handler's purpose.

Since the two must-fix paths (Status tab with fuzzy disabled, hash-card close/reopen) evidently have no E2E coverage, a small Playwright case for either would be a nice follow-up, not a blocker.

vstakhov (approved):
But we have some merge conflicts now I suppose.

moisseev:
I believe the easiest way is to revert commit 8d934a9790956fd21b89532e5ecfe2e00115285c

PR #6130, opened 2026-07-10
fix(lua-lupa): Use local loop variable

AmateurECE:
Lua 5.5 makes `for` expression variables constant. The recent upgrade to lua 5.5 broke lua-lupa, which previously attempted to modify the loop variable. This caused an unrecoverable error on startup:

```
rspamd_lua_require_function: require of lua_util.jinja_template failed: error loading module 'lupa' from file '/usr/share/rspamd/lualib/lupa.lua':\x0A\x09/usr/share/rspamd/lualib/lupa.lua:1858: attempt to assign to const variable 'word'
```


vstakhov:
This change is correct in general but it is not enough to fix lua 5.5 compatibility with regard to this language shift.

PR #6140, opened 2026-07-16
elasticsearch exporter, add elastic common schema mode

JustKiddingCode:
This PR adds a new mode to the elasticsearch exporter to export the data in [ECS](https://www.elastic.co/docs/reference/ecs).
Using ECS, index mapping can be greatly simplified and usage is consistent across different log ingestors.
Currently ecs mode should be considered beta, so field names can be changed.


JustKiddingCode:
Rebased to newest rspamd version.

PR #6145, opened 2026-07-22
[Rework] WebUI: drop Font Awesome framework

moisseev:
Replace Font Awesome's "SVG with JS" framework with a local subset SVG sprite. Delete fontawesome.min.js and solid.min.js (~900 KB of JS) and svg-with-js.min.css, and drop the fontawesome/fontawesome_solid RequireJS paths and shim.

app/icons.js renders the glyphs: an initial sweep plus a MutationObserver replace <i class="fas fa-<name>"> with inline <svg class="svg-inline--fa"> <use href="img/icons.svg#fa-<name>"/></svg>, carrying over non-class attributes (id) and modifier classes (fa-spin, spacing). setIcon() swaps an icon in place; the theme toggle (rspamd.js) and the map-modal header (config.js) move off class toggling and [data-fa-i2svg] selectors. Each icon keeps its natural aspect ratio: rspamd.css sizes the box at 1em square by default, and icons.css (generated) overrides the width of non-square glyphs. rspamd.css also adds the fa-spin keyframes svg-with-js.min.css used to provide.

The sprite (img/icons.svg, 35 glyphs) and icons.css are generated by icons-build.mjs (repo root) from icons-manifest.txt; v5 aliases (e.g. times -> xmark) resolve at build time. Font Awesome (@fortawesome/fontawesome-free, pinned 6.6.0) is a devDependency; CI (ci_eslint) runs "npm run check:icons" - icons-check.mjs verifies every fa-<name> in the markup is listed in the manifest, then rebuilds and fails on any diff (stale files or Font Awesome version drift).


moisseev:
@review

PR #6146, opened 2026-07-22
[Fix] Redact credentials from map error logs

moisseev:
Map URIs may carry HTTP basic-auth userinfo ("user:pass@host"). It was logged verbatim in CRITICAL (msg_err_*) messages, which feed the /errors ring buffer exposed via the controller, leaking credentials to anyone reading it. Add a redacted uri_log copy (userinfo replaced with "***@") used in error logs; the real uri is kept intact as it serves as the "maps { <uri> {} }" configuration key and carries the userinfo needed for the Authorization header.


moisseev:
@review

vstakhov:
We have conflicts now.

PR #6147, opened 2026-07-23
elastic: export message size in rspamd_meta metadata

### LINKED ISSUE #6136: [Feature] elastic: add message size (bytes) to exported metadata
## Summary

The elastic Lua module (src/plugins/lua/elastic.lua) exports a rich set of per-message metadata to Elasticsearch/OpenSearch via get_general_metadata(task) (fields like scan_time, score, symbols, language, received_delay, etc.), but it does not currently include the raw message size in bytes. I'd like to see a size field added to the exported rspamd_meta document, populated from the existing task:get_size() API (which already returns the task size in bytes, including headers + parts).

## Motivation

Message size is a common data point for spam/mail-flow analysis: spotting oversized attachments, correlating size with specific spam/phishing campaigns, capacity planning, and building dashboards in Kibana/OpenSearch Dashboards. Since the elastic module already collects and indexes a broad set of per-message metadata for exactly this kind of analysis, having size available directly on the document — rather than needing to join it in from another data source — would make it significantly easier to build such dashboards and queries. The change itself is minimal: task:get_size() already exposes the value, so it's just a matter of adding it to the metadata table that's already being assembled:

r.size = task:get_size() or 0

This would be picked up automatically by create_bulk_json() and pushed as part of the existing bulk payload, and could be mapped as a numeric field (e.g. long) in the module's managed index template so it's aggregatable/sortable out of the box.

## Describe alternatives you've considered

Correlating Postfix/mail log size fields with the queue ID after the fact. This works, but requires joining two separate data sources (Postfix logs and the elastic module's index) instead of having the size directly on the same document as the rest of the scan metadata.
Patching elastic.lua locally. Also works, but the elastic module is a stock plugin without a local.d/override.d extension point for adding arbitrary custom fields to the exported metadata (unlike, say, collect_headers/extra_collect_headers, which only cover headers). A local patch would need to be reapplied on every Rspamd upgrade, which isn't sustainable for a production setup.

Neither alternative is a good substitute for having the field natively available in the module, since the point is to keep all per-message data (score, symbols, size, etc.) together on one indexed document.

## Additional context

If maintainers would prefer a more general mechanism instead of hardcoding size — e.g. an extension point for adding arbitrary task-level fields to the exported rspamd_meta object, similar in spirit to collect_headers/extra_collect_headers but not limited to headers — I'd be fine with that approach as well, as long as size ends up being one of the fields that can be added.

TomFuch:
This PR adds the message size (in bytes) to the metadata exported by the elastic Lua module.

The size field is populated using task:get_size(), which includes the complete message size (headers and body parts).

This allows Elasticsearch/OpenSearch users to analyze message sizes, detect oversized messages or attachment-heavy campaigns, and improve mail-flow dashboards.

Fixes #6136


dragoangel:
You not added index change, its very bad thing to do. I shipped proper fix + changes for new fuzzy options in #6154

TomFuch:
Thanks, closing this as it's now in master

PR #6154, opened 2026-07-26
[Feature] elastic: structured fuzzy results and message size

dragoangel:
- `fuzzy`: replace flat `fuzzy_hashes` with per-match objects from `task:get_fuzzy_results()` (rule, symbol, found, queried, type, prob, flag, exact); `fuzzy_nested` toggles nested type
- `size`: `task:get_size()`, requested in #6136

 #6147 added `size` to the document only, but `rspamd_meta` is `dynamic: true`, so without
the mapping the type is inferred per index instead of declared.


vstakhov (changes requested):
The structured fuzzy-result path and Elasticsearch mapping look sound, but this drops compatibility with external fuzzy implementations that populate only the legacy fuzzy_hashes mempool variable. task:get_fuzzy_results() is empty for those implementations, so fuzzy data will silently disappear from Elasticsearch after this change. The milter exporter explicitly preserves this fallback. Please retain fuzzy_hashes when the structured list is empty, or provide a compatible conversion path.

vstakhov:
In fact, that's bullshit, ignore it :)

vstakhov (approved):

PR #6156, opened 2026-07-27
[Minor] Expose Errors history to read-only mode

moisseev:
The /errors endpoint and its WebUI table were gated behind the enable (privileged) password. Lower the privilege check so any authenticated user can view them.

This is safe: the endpoint returns only Rspamd's internal operational error log (timestamp, pid, level, module, message) — no email content, PII, or secrets — and read-only users already see strictly more sensitive data in the History tab (sender/ recipient IPs, scores, symbols). Authentication remains required and the endpoint is purely informational (no mutation).

### DIFF
--- a/interface/index.html
+++ b/interface/index.html
@@ -741,7 +741,7 @@
 						</div>
 					</div>
 				</div>
-				<div class="card shadow my-3 ro-hide" id="errors-history">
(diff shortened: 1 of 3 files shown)


moisseev:
@review

PR #6162, opened 2026-07-30
[CritFix] Gate privileged file and shared memory message source inputs

vstakhov:
## Summary

The `File`, `Path`, `Shm`, `Shm-Offset` and `Shm-Length` headers, their v3 metadata equivalents (`file`, `shm`, `shm_offset`, `shm_length`) and the proxy's `File` query argument make rspamd open a path or map a shared memory object that the **client** names. They were accepted on any transport, so any client that could reach a scanner, controller or proxy port could have an arbitrary file readable by the rspamd user parsed and reported back to it.

This adds a per-worker `allow_file_and_shm_inputs` option, gates every privileged message-source control behind it, and hardens the file and shared-memory input paths themselves.

**This release stays compatible: the code default and every shipped default are `true`.** The default becomes `false` in the next major release, which is stated in the shipped config comments. Operator-facing documentation is maintained separately on docs.rspamd.com.

## The capability model

The decision is derived **solely from the transport a connection was accepted on**:

```
allowed = (accepted transport is AF_UNIX) || allow_file_and_shm_inputs
```

Unix sockets allow these inputs independently of the option, since access is already controlled by the socket's filesystem permissions. Nothing the client sends — `User-Agent`, forwarded headers, query arguments, request metadata — takes part in the decision.

Authentication is a **separate axis**: a controller password, an `enable_password`, a `secure_ip` match and an encrypted connection do not unlock filesystem or shared memory access. This is verified by test, including over an encrypted `rspamc --key` connection.

A prominent startup warning is emitted per TCP listener while the option is enabled, escalated for a non-loopback listener.

## Gating

Rejection happens before the named object is opened, mapped or statted, and returns a genuine HTTP 400:

- v2 headers are rejected at the request-header choke point, so a privileged control is never even stored on the task;
- the v3 raw-`Shm` request body path is gated separately — it maps a segment without going through `rspamd_task_load_message()`;
- the v3 metadata keys are gated before any header is synthesised;
- `rspamd_task_load_message()` is the backstop for every caller;
- the proxy rejects `File` and `Path` as header and as query argument.

### Two vectors found during the work

- **`rspamd_protocol_handle_url()` converts query arguments into request headers at the upstream**, so `?Path=…` / `?Shm=…` in a proxy URL reached a scanner that trusts the proxy's loopback connection. The proxy now strips those arguments.
- **`fuzzy_check` builds its own task** for `/fuzzyadd` and `/fuzzydel`, which bypassed the controller's gate. It now inherits the capability from the accepted transport.

## Input hardening

- Paths are validated before any syscall (empty, embedded NUL, control characters, overlong), and a **truncated** path is rejected rather than acted upon.
- File input opens first and validates via `fstat` (no pathname `stat`), requires a regular file, rejects unrepresentable sizes, enforces `cfg->max_message` before reading, and takes a **bounded snapshot** handling short reads and `EINTR` instead of parsing a `MAP_SHARED` mapping that can fault if the caller truncates the file.
- Shared memory maps only the required **page-aligned window** rather than the whole backing object, maps nothing at all for a zero-length window, and snapshots the window before unmapping — so a mutable caller-owned object cannot fault the parser.
- `Shm`/`Shm-Offset`/`Shm-Length` are treated as reserved hop-by-hop headers: the proxy strips every client-supplied instance at ingress, and the internal triplet is inserted after removing any existing instances rather than appended behind them. When shared-memory transport is not permitted for a TCP upstream, a bounded inline body is forwarded instead. Encrypted and unencrypted forwarding behave identically.
- `Filename` remains ordinary descriptive metadata and is explicitly not treated as the privileged `File` control.

## Mapping lifecycle

Both `munmap` sites in the HTTP layer derived their length from a fresh `fstat`, which **over-unmaps whenever the segment grew after a copy mapped it** (a proxy copy maps `st_size` at copy time while the owner keeps growing the same object). They now use the length captured at `mmap` time. Failed `ftruncate`/`mmap`/`dup`/`fstat` operations close descriptors, release names and leave no dangling `begin` pointer, and an immutable body can no longer be routed into the ftruncate-and-remap path.

## Admission limits

The scanner's `max_tasks` counted only requests whose complete body had arrived, so connections between accept and body-complete were uncounted; it also used `>` and returned without accepting, which spun on the level-triggered listener. It now counts accepted and body-pending connections, is enforced with `>=` before any request or body state is allocated, and accepts-then-closes when over the limit. Accounting is idempotent so every early error and pre-task disconnect decrements exactly once.

Controller and proxy gain `max_connections` and `max_connections_per_source` (default `0` = unlimited), enforced before pool, session, KDF, file or shared-body allocation.

## Related corrections

- `RSPAMD_TASK_PROTOCOL_FLAG_LOCAL_CLIENT` was set whenever the client claimed `User-Agent: rspamc`. It is now derived from the accepted transport and peer identity. It had no readers, so it granted nothing, but it must not become a privilege source.
- `encrypted_only` exempted peers via `rspamd_inet_address_is_local()`, which also matches IPv6 link-local and site-local, letting any host on the segment skip the encryption requirement. Both the scanner and proxy now use a strict loopback/unix test.
- Reply writing folds error codes into the 5xx range, which would report a client error as `500 Internal Server Error`. An opt-in flag lets these rejections emit their real status without changing any existing error's mapping.

## Testing

- **C++ unit suite** (`test/rspamd_cxx_unit_task_input.hxx`): page-aligned windows at awkward offsets, a small window of a large backing object, zero-length windows, `max_size`, the combined offset+length overflow check, path validation, and snapshot stability across a concurrent truncate and grow. RAII cleanup so a failing assertion cannot leak an object. LeakSanitizer output is byte-identical with and without the new suite.
- **Four functional suites** covering the transport-derived behaviour: rejection over TCP for v2 and v3, the proxy query argument, unix socket retaining the capability, compatibility when enabled, client `Shm*` never overriding proxy values, inline forwarding, `max_message`, and pending connections counting toward all three admission limits with counters released on disconnect.
- **Each new suite has a verified negative control** — flipping only the config value makes exactly the intended tests fail and no others.
- Existing suites pass: 384/384 merged (checkv3, controller, scanner, compression, encrypted protocol) and 23/23 others.
- Build clean under ASAN+LSAN; C++ 302/302; Lua 1173 passed / 0 failed; luacheck clean; all modified files clang-format clean.

Tests use only benign temporary files and shared memory objects they create and clean up themselves.

## Notes for reviewers

- This macOS build has `HAVE_SANE_SHMEM` undefined, so the local shm tests exercise the `open()` flavour. **The `shm_open` path is only exercised on Linux CI** — worth confirming there.
- The "non-local TCP upstream" case exercises the "loopback and not opted in" arm rather than a genuinely non-loopback upstream, which would need `127.0.0.x` loopback aliases.
- A file or shm payload exceeding `max_message` returns **503**, consistent with how rspamd already reports oversized compressed input, whereas the capability gate returns 400. Happy to make size violations client errors too if preferred — it is a one-line change.

### DIFF
--- a/conf/worker-controller.inc
+++ b/conf/worker-controller.inc
@@ -17,3 +17,19 @@ password = "q1";
 secure_ip = "127.0.0.1";
 secure_ip = "::1";
 static_dir = "${WWWDIR}";
+
+# Allow client-named message sources: the `File`/`Path`/`Shm`/`Shm-Offset`/
+# `Shm-Length` headers and the v3 `file`/`shm`/`shm_offset`/`shm_length`
+# metadata keys, which make rspamd read a path or map a shared memory object
+# chosen by the client. TCP listeners only: unix sockets always permit these
+# inputs, since filesystem permissions control access there.
+# Do not enable on a listener reachable from untrusted networks. Neither
+# `password`/`enable_password`, nor a `secure_ip` match, nor an encrypted
+# connection authorizes this capability: it is a separate one.
+# `true` here for backwards compatibility; it becomes `false` by default in the
+# next major release.
+allow_file_and_shm_inputs = true;
+
+# Upper bound on concurrent connections served by a controller process.
+# 0 (the default) means unlimited; a finite value bounds concurrency.
+# max_connections = 0;
--- a/conf/worker-normal.inc
+++ b/conf/worker-normal.inc
@@ -13,3 +13,14 @@
 # Module documentation: https://rspamd.com/doc/workers/normal.html
 
 mime = true;
+
+# Allow client-named message sources: the `File`/`Path`/`Shm`/`Shm-Offset`/
+# `Shm-Length` headers and the v3 `file`/`shm`/`shm_offset`/`shm_length`
+# metadata keys, which make rspamd read a path or map a shared memory object
+# chosen by the client. TCP listeners only: unix sockets always permit these
+# inputs, since filesystem permissions control access there.
+# Do not enable on a listener reachable from untrusted networks. Authentication
+# and encryption do NOT authorize this capability.
+# `true` here for backwards compatibility; it becomes `false` by default in the
+# next major release.
+allow_file_and_shm_inputs = true;
--- a/conf/worker-proxy.inc
+++ b/conf/worker-proxy.inc
@@ -42,3 +42,19 @@ discard_on_reject = false; # Discard message instead of rejection
 quarantine_on_reject = false; # Tell MTA to quarantine rejected messages
 spam_header = "X-Spam"; # Use the specific spam header
 reject_message = "Spam message rejected"; # Use custom rejection message
+
+# Allow client-named message sources: the `File`/`Path`/`Shm`/`Shm-Offset`/
+# `Shm-Length` headers, the `File` query argument, and the v3
+# `file`/`shm`/`shm_offset`/`shm_length` metadata keys, which make rspamd read
+# a path or map a shared memory object chosen by the client. TCP listeners
+# only: unix sockets always permit these inputs, since filesystem permissions
+# control access there.
+# Do not enable on a listener reachable from untrusted networks. Authentication
+# and encryption do NOT authorize this capability.
+# `true` here for backwards compatibility; it becomes `false` by default in the
+# next major release.
+allow_file_and_shm_inputs = true;
+
+# Upper bound on concurrent client connections served by a proxy process.
+# 0 (the default) means unlimited; a finite value bounds concurrency.
+# max_connections = 0;
--- a/src/controller.c
+++ b/src/controller.c
@@ -51,6 +51,13 @@
 /* Bounds the memory used by the throttling state */
 #define AUTH_FAILURES_CACHE_SIZE 1024
 
+/*
+ * Admission control limits, both disabled by default so that this release
+ * behaves exactly as the previous ones unless an operator opts in.
+ */
+#define DEFAULT_MAX_CONNECTIONS 0
+#define DEFAULT_MAX_CONNECTIONS_PER_SOURCE 0
+
 /* HTTP paths */
 #define PATH_AUTH "/auth"
 #define PATH_SYMBOLS "/symbols"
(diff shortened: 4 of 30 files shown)


vstakhov:
Both findings were real and are fixed in d32102699, each with a regression test that was confirmed to fail against the pre-fix code.

**[P1] Snapshot race with truncation**

Correct, and the test gap was exactly as described. The window is now read straight into pool storage with a `pread` loop handling short reads and `EINTR`, so a concurrent `ftruncate` returns fewer bytes rather than faulting, and `seg->data_len` reflects what was actually read instead of what was requested.

One portability wrinkle worth recording: POSIX shared memory descriptors do not accept positional reads everywhere — I measured `pread` on an `shm_open` fd returning `ESPIPE` on macOS, and some BSDs return `ENODEV`. A blind switch to `pread` would have broken the shm flavour on those platforms. The mapping path is therefore retained purely as a fallback for descriptors that reject positional reads. On Linux, where `HAVE_SANE_SHMEM` is unconditionally set and shm objects live on tmpfs, the read path is the one taken and the race is closed.

Regression test: a thread toggles a 2 MiB backing object between full size and 4 KiB while the main thread takes 600 snapshots of a 512 KiB window at the end of it. Against the pre-fix code this dies with SIGBUS (exit 138, `rspamd_shmem_segment_map` → `__asan_memcpy`), reproduced 3/3; against the fix it passes, stressed 30x standalone. The object is zero-filled so that any byte sourced from the allocation rather than the object is detectable, and the thread is bounded by both a stop flag and a wall-clock deadline and joined from a destructor, so a throwing assertion cannot leave it running.

A deterministic non-racy short-read case turned out not to be reachable through the public API — the up-front validation rejects `length > st_size - offset`, so the only way the object can be short at read time is the race itself. Rather than fabricate one, that semantics is pinned by assertions that hold on every iteration of the race test (`data_len` never exceeds the request; the buffer holds exactly `data_len` real bytes).

**[P2] Stripping Shm corrupts remaining query arguments**

Confirmed against the bundled `http_parser`: for `/checkv2?From=a@b.c&Shm=/x`, `UF_QUERY.off` is 9, so the copied prefix was `/checkv2?` and the appended delimiter produced `/checkv2??From=a@b.c`. The prefix now stops at `off - 1`, with a defensive guard for `off == 0`. The existing trailing-separator erase already handles the case where every argument is stripped.

Regression test: a request now carries an ordinary query argument alongside a privileged one, with `From` deliberately not also sent as a header so nothing can mask a mangled name. The upstream probe reports both `From` and `?From`, so a failure names the corruption. Against the pre-fix code it fails with `has options ['?From=qsurvive@example.net'] but expected ['From=qsurvive@example.net']`.

**Validation after the fixes**

- C++ suite 303/303 (77152 assertions), up one case from the new race test
- Functional suite 572: 9/9
- Lua 1173 passed / 0 failed; luacheck clean
- clang-format and `git diff --check` clean; signature valid

PR #5334, opened 2025-02-05
[Fix] properly close multipart/related boundary when adding text footer

FreddleSpl0it:
When adding a footer to an HTML part with an inline image,  
`lua_mime.add_text_footer` messes up the boundaries.  
As a result, Outlook doesn’t show other attachments.

EML without using `lua_mime.add_text_footer`
```
Content-Type: multipart/mixed; boundary="----=_=-_OpenGroupware_org_NGMime-55-1730359886.137956-7------"



------=_=-_OpenGroupware_org_NGMime-55-1730359886.137956-7------
Content-Type: multipart/alternative; boundary="----=_=-_OpenGroupware_org_NGMime-55-1730359886.137208-6------"

------=_=-_OpenGroupware_org_NGMime-55-1730359886.137208-6------
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable
Content-Length: 47


before inline image

after inline image
=C2=A0

------=_=-_OpenGroupware_org_NGMime-55-1730359886.137208-6------
Content-Type: multipart/related; boundary="----=_=-_OpenGroupware_org_NGMime-55-1730359886.136648-5------"

------=_=-_OpenGroupware_org_NGMime-55-1730359886.136648-5------
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: quoted-printable
Content-Length: 192

<html><body><p>before inline image</p><p><img src=3D"cid:37-67233280-F-=
468CF900" type=3D"image/gif" width=3D"354" height=3D"245"/></p><p>after=
 inline image</p><p>&#160;</p></body></html>

------=_=-_OpenGroupware_org_NGMime-55-1730359886.136648-5------
Content-Transfer-Encoding: base64
Content-Type: image/gif; name="37-67233280-F-468CF900"
Content-ID: <37-67233280-F-468CF900>
Content-Disposition: inline; filename="37-67233280-F-468CF900"
Content-Length: 41800

R0lGODlhY...

------=_=-_OpenGroupware_org_NGMime-55-1730359886.136648-5--------

------=_=-_OpenGroupware_org_NGMime-55-1730359886.137208-6--------

------=_=-_OpenGroupware_org_NGMime-55-1730359886.137956-7------
Content-Type: image/jpeg
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="Untitled-1.jpg"
Content-Length: 89761

/9j/4RAfRXh...

[...]

PR #6173, opened 2026-08-07
[Fix] lua_mime: unwind every boundary level when returning from nested parts

### LINKED ISSUE #6170: lua_mime.add_text_footer: boundary stack desynchronises on nested parts, truncating the message mid-body
### Summary

`lua_mime.add_text_footer()` desynchronises its `boundaries` stack when the part walk returns from a nested part to a shallower one. It pops a fixed number of entries — one, or two in the `multipart/related` special case added by #5334 — instead of unwinding every level actually left behind.

Once the stack drifts, `boundaries[#boundaries]` names an unrelated boundary, and a closing delimiter for the **top-level** boundary is emitted into the middle of the body. Every MIME parser stops there. Attachments after that point are still present in the byte stream but unreachable.

The failure is silent: no error, no log line, and the message grows rather than shrinks, so size checks do not catch it. In the deployment where we hit this, a forwarded bundle of 18 messages was delivered as 5, and it went unnoticed for some time because nothing anywhere reported a problem.

Reported downstream as mailcow/mailcow-dockerized#7399, where it has been independently reproduced by a second party on a separate stand.

### Affected code

Current `master` (`lualib/lua_mime.lua`), unchanged from the versions we tested.

The `is_message` branch pops exactly one (L523–531):

```lua
    elseif part:is_message() then
      if boundary then
        if cur_boundary and boundary ~= cur_boundary then
          -- Need to close boundary
          out[#out + 1] = string.format('--%s--%s',
              boundaries[#boundaries].boundary, newline_s)
          table.remove(boundaries)
          cur_boundary = nil
        end
```

The leaf branch pops one, or two when the top of the stack is `multipart/related` (L566–578):

```lua
      if boundary then
        if cur_boundary and boundary ~= cur_boundary then
          -- Need to close boundary
          out[#out + 1] = string.format('--%s--%s',
              boundaries[#boundaries].boundary, newline_s)
          -- Need to close previous boundary, if ct_subtype is related
          if #boundaries > 1 and boundaries[#boundaries].ct_type == "multipart" and boundaries[#boundaries].ct_subtype == "related" then
            out[#out + 1] = string.format('--%s--%s',
                boundaries[#boundaries - 1].boundary, newline_s)
            table.remove(boundaries)
          end
          table.remove(boundaries)
          cur_boundary = boundary
        end
```

Neither branch knows how many levels the walk actually crossed. The number of boundaries to close is a property of the message structure, not a constant.

Note that #5334 fixed a real instance of this same underlying problem — the report there is also "boundaries messed up, Outlook does not show other attachments". The two-pop special case makes that particular shape work, but it is a fix for one arrangement rather than for the stack handling, so adjacent shapes still fail.

### Trigger

A nested `message/rfc822` whose body nests one level deeper, with `multipart/related` sitting **directly** under the inner `multipart/mixed`:

```
multipart/mixed                <- outer container
  message/rfc822               <- forwarded message
    multipart/mixed            <- its body
      multipart/related        <- the trigger
        text/html + inline image
      text/plain attachment
```

Only attachments **after** such a message are lost.

The condition is narrow, which is why this shows up on some forwards and not others:

| nested structure | corrupted |
|---|---|
| `mixed > related` | yes |
| `mixed > alternative > related` | no |
| no inline images | no |

The `alternative` case was checked separately by the second reporter in the downstream thread (18 nested messages, all 18 survived). It is consistent with the code: the two-pop branch inspects only the top of the stack, so an extra `alternative` level between `mixed` and `related` changes which entry that test sees. Depth alone is not the trigger.

### Reproducer

Standalone, no real data, 6.5 KB. Builds 8 nested `message/rfc822` parts where item #3 carries the `multipart/related`:

<details>
<summary><code>mk_repro.py</code></summary>

```python
#!/usr/bin/env python3
"""Minimal reproducer for the boundary-stack desynchronisation in
lua_mime.add_text_footer."""
CRLF = "\r\n"
OUTER = "----=_outer_boundary_here_=="


def nested_plain(i):
    """A forwarded message with a flat multipart/mixed body."""
    b = f"=_inner_{i}_=="
    return [f"From: sender{i}@example.org",
            "To: recipient@example.com",
            f"Subject: Nested message {i}",
            "Date: Mon, 03 Aug 2026 12:00:00 +0000",
            f"Message-ID: <nested-{i}@example.org>",
            "MIME-Version: 1.0",
            f'Content-Type: multipart/mixed; boundary="{b}"',
            "",
            f"--{b}",
            "Content-Type: text/plain; charset=utf-8",
            "Content-Transfer-Encoding: 7bit",
            "",
            f"Body of nested message {i}.",
            "",
            f"--{b}",
            "Content-Type: application/octet-stream",
            "Content-Transfer-Encoding: base64",
            f'Content-Disposition: attachment; filename="file{i}.bin"',
            "",
            "cGF5bG9hZCBkYXRhCg==",
            "",
            f"--{b}--",
            ""]


def nested_related(i):
    """A forwarded message whose body nests multipart/related one level
    deeper. This is the shape that desynchronises the stack."""
    b_mixed = f"=_inner_mixed_{i}_=="
    b_rel = f"=_inner_related_{i}_=="
    return [f"From: sender{i}@example.org",
            "To: recipient@example.com",
            f"Subject: Nested message {i} (related)",
            "Date: Mon, 03 Aug 2026 12:00:00 +0000",
            f"Message-ID: <nested-{i}@example.org>",
            "MIME-Version: 1.0",
            f'Content-Type: multipart/mixed; boundary="{b_mixed}"',
            "",
            f"--{b_mixed}",
            f'Content-Type: multipart/related; boundary="{b_rel}"',
            "",
            f"--{b_rel}",
            "Content-Type: text/html; charset=utf-8",
            "Content-Transfer-Encoding: 7bit",
            "",
            f'<html><body><p>Body {i}</p>'
            f'<img src="cid:img{i}@example.org"></body></html>',
            "",
            f"--{b_rel}",
            "Content-Type: image/png",
            "Content-Transfer-Encoding: base64",
            f"Content-ID: <img{i}@example.org>",
            "",
            "iVBORw0KGgo=",
            "",
            f"--{b_rel}--",
            "",
            f"--{b_mixed}",
            "Content-Type: text/plain; charset=utf-8",
            "Content-Transfer-Encoding: base64",
            f'Content-Disposition: attachment; filename="file{i}.txt"',
            "",
            "cGF5bG9hZCBkYXRhCg==",
            "",
            f"--{b_mixed}--",
            ""]


def build(path, n_before, n_after):
    L = ["From: forwarder@example.org",
         "To: recipient@example.com",
         "Subject: Forwarded bundle",
         "Date: Mon, 03 Aug 2026 13:00:00 +0000",
         "Message-ID: <outer@example.org>",
         "MIME-Version: 1.0",
         f'Content-Type: multipart/mixed; boundary="{OUTER}"',
         "",
         f"--{OUTER}",
         "Content-Type: text/plain; charset=utf-8",
         "Content-Transfer-Encoding: 7bit",
         "",
         "Forwarded messages attached.",
         ""]

    idx = 0

    def attach(body_lines, i):
        L.append(f"--{OUTER}")
        L.append("Content-Type: message/rfc822")
        L.append("Content-Transfer-Encoding: 8bit")
        L.append(f'Content-Disposition: attachment; filename="nested{i}.eml"')
        L.append("MIME-Version: 1.0")
        L.append("")
        L.extend(body_lines)

    for _ in range(n_before):
        idx += 1
        attach(nested_plain(idx), idx)

    idx += 1
    attach(nested_related(idx), idx)          # <-- the trigger

    for _ in range(n_after):
        idx += 1
        attach(nested_plain(idx), idx)

    L.append(f"--{OUTER}--")
    L.append("")

    data = CRLF.join(L).encode()
    open(path, "wb").write(data)
    print(f"{path}: {len(data)} bytes, {idx} nested message/rfc822")


build("/tmp/repro.eml", n_before=2, n_after=5)
```
</details>

Apply a footer through `add_text_footer()` — via `rspamadm mime modify -t/-H`, or through a plugin — and count how many `message/rfc822` parts a strict parser can still reach.

### Results

Measured with Python's `email` parser under `policy.default`, cross-checked against the raw byte offset of the top-level closing delimiter, so it is not an artefact of a lenient reader.

| build | reachable of 8 | top-level closing delimiter |
|---|---|---|
| input (before footer) | 8 | end of message |
| stock 4.1.0 | **3** | mid-body |
| stock 4.1.4 | **3** | byte 3956 of 7539 — 52.5%, mid-body |
| patched (below) | **8** | end of message |

The 4.1.4 row is the independent reproduction from the downstream thread, on a separate stand via `rspamc -m`.

On the original 4.4 MB production message the stack went **empty 14 times** during a single walk, and the closing delimiter landed at byte 3,815,438 of 4,494,002 — roughly 700 KB before the end.

### Suggested fix

Unwind to the boundary being re-entered instead of popping a fixed count. Falls back to the previous single-pop behaviour when the target is not on the stack, so the `message/rfc822` case that opens a boundary of its own is unchanged:

```lua
  -- Emit closing delimiters for all boundaries we are leaving behind, stopping
  -- when the top of the stack is `target` (the boundary about to be re-entered).
  -- If `target` is not on the stack at all -- which happens when a nested
  -- message/rfc822 opens a boundary of its own -- close exactly one level, so
  -- the behaviour stays as before for that case.
  local function close_boundaries_until(out, boundaries, target, newline_s)
    local found = false
    for i = 1, #boundaries do
      if boundaries[i].boundary == target then
        found = true
        break
      end
    end

    if not found then
      if #boundaries > 0 then
        out[#out + 1] = string.format('--%s--%s',
            boundaries[#boundaries].boundary, newline_s)
        table.remove(boundaries)
      end
      return
    end

    while #boundaries > 0 and boundaries[#boundaries].boundary ~= target do
      out[#out + 1] = string.format('--%s--%s',
          boundaries[#boundaries].boundary, newline_s)
      table.remove(boundaries)
    end
  end
```

Both call sites then become:

```lua
      if boundary then
        if cur_boundary and boundary ~= cur_boundary then
          close_boundaries_until(out, boundaries, boundary, newline_s)
          cur_boundary = boundary   -- `nil` in the is_message branch, as before
        end
```

This subsumes the `multipart/related` special case from #5334 — that shape is simply one instance of "more than one level to close" — so the explicit two-pop test can go.

### Regression check

Ten messages of differing structure (flat `text/plain`, flat `text/html`, `multipart/alternative`, `multipart/related`, `mixed > related` + attachment, nested forwards with and without the trigger, and the `mixed > alternative > related` shape) produced byte-identical output before and after the patch, except for the cases that were previously corrupted. The footer is still applied in every case, and sub-part encodings are untouched.

Also verified on live traffic on a production host for a day: a 14-message / 28-attachment forward that previously lost most of its attachments now arrives complete, footer applied.

### Environment

- Rspamd 4.1.0 and 4.1.4, defect present in current `master`
- Reached through mailcow's domain-wide footer feature, but the mailcow wrapper is not involved — the downstream reproduction drove `rspamc -m` directly, and `rspamadm mime modify` shows it too

I am happy to open a PR with the above if the approach looks right.

smpaz7467:
Fixes #6170.

### The defect

`lua_mime.add_text_footer()` closes a fixed number of boundaries when the part walk returns from a nested part to a shallower one — one in the `is_message` branch, or two in the `multipart/related` special case added by #5334. The number of levels to close is a property of the message structure, so a constant cannot be right. Once the stack drifts, `boundaries[#boundaries]` names an unrelated boundary and a closing delimiter for the **top-level** one is emitted into the middle of the body.

Everything after that delimiter is epilogue per RFC 2046, so attachments are still in the byte stream but unreachable to any strict parser. The failure is silent: no error, no log line, and the message *grows* rather than shrinks, so size checks do not catch it.

One point worth drawing out, because it changes how the special case should be read: for `multipart/mixed > multipart/related` with a trailing attachment, the two-pop branch is what *emits* the premature top-level delimiter. Without it that shape already worked. So #5334 did not merely under-fix the stack handling — it broke an adjacent shape while fixing its own. That is the argument for replacing the special case rather than extending it.

### The fix

`close_boundaries_until()` unwinds to the boundary being re-entered instead of popping a fixed count. When the target is not on the stack at all — a nested `message/rfc822` that opens a boundary of its own — it closes exactly one level, so that path behaves as before. This subsumes the `multipart/related` special case, which is simply one instance of "more than one level to close".

The approach is @Viktort-t's, proposed in #6170; I have packaged it with tests and an A/B.

### Two independent reproductions

This was reproduced twice, by different people, with different methods, before any patch was written — worth recording since the two routes fail in different directions.

**@Viktort-t** hit it in production via mailcow's domain-wide footer feature and reported it downstream as mailcow/mailcow-dockerized#7399: a forwarded bundle of 18 messages delivered as 5. On a 4.4 MB production message the stack went empty 14 times in a single walk, with the closing delimiter landing ~700 KB before the end. Because that host is production they could not swap `lua_mime.lua` for a side-by-side control, so they loaded each version by path with `dofile` and called `add_text_footer` directly, md5-verifying the live file was untouched throughout.

**I** reproduced it independently on a disposable stand, driving Rspamd 4.1.4 (`ghcr.io/mailcow/rspamd:4.1.4-1`) through `rspamc -m` and `rspamadm mime modify` so that only Rspamd was in the path. That established the defect is live on 4.1.4 as well as 4.1.0 — the original report cited only 4.1.0, so upgrading is not a workaround — and it narrowed the trigger: an 18-part `mixed > alternative > related` bundle did **not** corrupt, which is what pointed at `related` sitting *directly* under the inner `mixed` rather than at nesting depth as such.

Two method notes that cost us time and may save someone else's:

- Shadowing the module via `LUA_PATH` does **not** work. `rspamadm` resolves `lua_mime` to the compiled-in `/usr/share/rspamd/lualib/` path regardless, so both arms of an A/B silently load the same file and produce byte-identical output — it fails in the direction that looks like a passing test. The module has to be replaced outright and the md5 of the loaded file confirmed to have changed.
- The smaller trigger below needs no `message/rfc822` nesting at all, and is the easier one to reason about.

### Verification

A/B on Rspamd 4.1.4 with only `add_text_footer` patched — `multipattern_text_replace` and `remove_attachments` contain the same "Need to close boundary" code and were deliberately left untouched, so the change under test is isolated. `lualib/lua_mime.lua` on current `master` is md5-identical to the one shipped in 4.1.4 (`b11ccc51…`), so these numbers carry over to `master` unchanged.

Reachability measured by raw byte offset of the top-level delimiter *and* independently by Python's `email` parser under `policy.default`:

| structure | build | top-level close at | leaf parts reachable |
|---|---|---|---|
| `mixed > related` + attachment | stock | 75.2% — mid-body | 2 |
| | patched | 98.2% — end | **3** |
| `mixed > alternative > related` + attachment | stock | 98.6% | 4 |
| | patched | 98.6% | 4 — **byte-identical to stock** |
| nested `message/rfc822` + attachment | stock | 86.2% — mid-body | 4 |
| | patched | 99.0% — end | **6** |

Exactly one closing delimiter in every build, so this is a premature close rather than a duplicate. The `mixed > alternative > related` row is byte-identical before and after, so the shape #5334 addressed is not regressed — @Viktort-t independently confirmed that same control on their side.

On @Viktort-t's original 8-part reproducer, reachable `message/rfc822` parts go from 3/8 (stock 4.1.0 and 4.1.4 alike) to 8/8.

### Tests

`test/lua/unit/lua_mime.add_text_footer.lua` covers the three shapes that distinguish the behaviours: the three-part `mixed > related` trigger, the `mixed > alternative > related` shape from #5334, and a nested `message/rfc822` bundle. Each asserts invariants — exactly one top-level closing delimiter, every part before it, footer still applied — rather than exact bytes, so they stay meaningful if footer placement or encoding changes.

The tests were checked against the unpatched module as well, so they are known to actually catch this:

```
                                                    unpatched   patched
mixed > related + attachment                          FAIL        PASS
mixed > alternative > related + attachment (#5334)    PASS        PASS
nested message/rfc822 + attachment                    FAIL        PASS
```

The two failures report the defect directly, e.g. `part ATTACHMENTMARKER lands after the top-level closing delimiter at 361/537, so it is RFC 2046 epilogue and unreachable`. Full Lua unit suite with the patch applied: 1194 tests, 1191 passed, 0 failed, 0 errors (3 pre-existing unassertive).

### Notes for review

- `ct_type` / `ct_subtype` are now written into the `boundaries` entries but never read, since the special case that consumed them is gone. I left them, and the `part_ct` parse that feeds them, in place to keep the diff to the actual behaviour change — happy to strip both if you would prefer.
- The same fixed-count close pattern appears in `multipattern_text_replace` and `remove_attachments`. Those are plausibly affected by the same class of bug, but I have not reproduced it there and have not touched them; that felt like it should be its own change rather than an unverified drive-by.

### DIFF
--- a/lualib/lua_mime.lua
+++ b/lualib/lua_mime.lua
@@ -430,6 +430,46 @@ local function do_append_footer(task, part, footer, is_multipart, out, state)
 
 end
 
+-- Emit closing delimiters for every boundary we are leaving behind, stopping once
+-- the top of the stack is `target` -- the boundary that is about to be re-entered.
+--
+-- The number of levels to unwind is a property of the message structure, so it
+-- cannot be a fixed count. Popping one entry (or two, as the previous
+-- multipart/related special case did) desynchronises the stack as soon as the part
+-- walk returns from a nested part to a shallower one; `boundaries[#boundaries]`
+-- then names an unrelated boundary and a closing delimiter for it -- in practice
+-- the top-level one -- is emitted into the middle of the body.
+--
+-- When `target` is not on the stack at all, which happens when a nested
+-- message/rfc822 opens a boundary of its own, close exactly one level so the
+-- behaviour for that case is unchanged.
+local function close_boundaries_until(out, boundaries, target, newline_s)
+  local found = false
+
+  for i = 1, #boundaries do
+    if boundaries[i].boundary == target then
+      found = true
+      break
+    end
+  end
+
+  if not found then
+    if #boundaries > 0 then
+      out[#out + 1] = string.format('--%s--%s',
+          boundaries[#boundaries].boundary, newline_s)
+      table.remove(boundaries)
+    end
+
+    return
+  end
+
+  while #boundaries > 0 and boundaries[#boundaries].boundary ~= target do
+    out[#out + 1] = string.format('--%s--%s',
+        boundaries[#boundaries].boundary, newline_s)
+    table.remove(boundaries)
+  end
+end
+
 --[[[
 -- @function lua_mime.add_text_footer(task, html_footer, text_footer)
 -- Adds a footer to all text parts in a message. It returns a table with the following
@@ -523,10 +563,8 @@ exports.add_text_footer = function(task, html_footer, text_footer)
     elseif part:is_message() then
       if boundary then
         if cur_boundary and boundary ~= cur_boundary then
-          -- Need to close boundary
-          out[#out + 1] = string.format('--%s--%s',
-              boundaries[#boundaries].boundary, newline_s)
-          table.remove(boundaries)
+          -- Need to close every boundary we are leaving behind
+          close_boundaries_until(out, boundaries, boundary, newline_s)
           cur_boundary = nil
         end
         out[#out + 1] = string.format('--%s',
@@ -565,16 +603,10 @@ exports.add_text_footer = function(task, html_footer, text_footer)
 
       if boundary then
         if cur_boundary and boundary ~= cur_boundary then
-          -- Need to close boundary
-          out[#out + 1] = string.format('--%s--%s',
-              boundaries[#boundaries].boundary, newline_s)
-          -- Need to close previous boundary, if ct_subtype is related
-          if #boundaries > 1 and boundaries[#boundaries].ct_type == "multipart" and boundaries[#boundaries].ct_subtype == "related" then
-            out[#out + 1] = string.format('--%s--%s',
-                boundaries[#boundaries - 1].boundary, newline_s)
-            table.remove(boundaries)
-          end
-          table.remove(boundaries)
+          -- Need to close every boundary we are leaving behind. This subsumes the
+          -- former multipart/related special case: that shape is simply one
+          -- instance of "more than one level to close".
+          close_boundaries_until(out, boundaries, boundary, newline_s)
           cur_boundary = boundary
         end
         out[#out + 1] = string.format('--%s',
--- a/test/lua/unit/lua_mime.add_text_footer.lua
+++ b/test/lua/unit/lua_mime.add_text_footer.lua
@@ -1,0 +1,306 @@
+--[[
+Boundary handling in lua_mime.add_text_footer.
+
+Regression coverage for the boundary stack desynchronising when the part walk
+returns from a nested part to a shallower one: the walk closed a fixed number of
+boundaries (one, or two in the former multipart/related special case) rather than
+every level it had actually left behind. Once the stack drifted,
+`boundaries[#boundaries]` named an unrelated boundary and a closing delimiter for
+the top-level one was emitted into the middle of the body. Everything after that
+point is epilogue per RFC 2046, so attachments were still in the byte stream but
+unreachable to any strict parser -- silently, with the message growing rather
+than shrinking.
+
+Each case asserts the same three invariants rather than exact bytes, so the tests
+stay meaningful if footer placement or encoding changes:
+  * exactly one top-level closing delimiter
+  * every part appears before it
+  * the footer is still applied
+
+Assertions live in the test bodies rather than in the helpers: telescope injects
+assert_* into the environment of the test function only, so a helper defined at
+context scope would see them as nil.
+]]
+
+context("lua_mime.add_text_footer", function()
+  local rspamd_task = require "rspamd_task"
+  local rspamd_util = require "rspamd_util"
+  local rspamd_test_helper = require "rspamd_test_helper"
+  local lua_mime = require "lua_mime"
+
+  rspamd_test_helper.init_url_parser()
+  local cfg = rspamd_util.config_from_ucl(rspamd_test_helper.default_config(),
+      "INIT_URL,INIT_LIBS,INIT_SYMCACHE,INIT_VALIDATE,INIT_PRELOAD_MAPS")
+
+  local html_footer = '<p>FOOTERMARKER</p>'
+  local text_footer = 'FOOTERMARKER'
+
+  -- Reassemble the body the same way rspamadm mime does: bare strings get a
+  -- newline appended, table entries carry an explicit "add a newline" flag.
+  local function body_to_string(rewrite)
+    local buf = {}
+
+    for _, o in ipairs(rewrite.out) do
+      if type(o) == 'string' then
+        buf[#buf + 1] = o
+        buf[#buf + 1] = rewrite.newline_s
+      elseif type(o) == 'table' then
+        buf[#buf + 1] = tostring(o[1])
+
+        if o[2] then
+          buf[#buf + 1] = rewrite.newline_s
+        end
+      else
+        buf[#buf + 1] = tostring(o)
+      end
+    end
+
+    return table.concat(buf)
+  end
+
+  local function count_substr(haystack, needle)
+    local n = 0
+    local pos = 1
+
+    while true do
+      local s, e = string.find(haystack, needle, pos, true)
+
+      if not s then
+        break
+      end
+
+      n = n + 1
+      pos = e + 1
+    end
+
+    return n
+  end
+
+  -- Apply the footer and return a list of everything wrong with the result, so
+  -- the test body can assert on it. Empty list means the message is intact.
+  local function footer_problems(message, markers)
+    local res, task = rspamd_task.load_from_string(message, cfg)
+
+    if not res or not task then
+      return { 'failed to load message' }
+    end
+
+    task:process_message()
+
+    local rewrite = lua_mime.add_text_footer(task, html_footer, text_footer)
+
+    if not rewrite or not rewrite.out then
+      task:destroy()
+      return { 'add_text_footer did not rewrite the message' }
+    end
+
+    local body = body_to_string(rewrite)
+    task:destroy()
+
+    local problems = {}
+    local n_close = count_substr(body, '--outerbnd--')
+
+    if n_close ~= 1 then
+      problems[#problems + 1] = string.format(
+          'expected exactly one top-level closing delimiter, got %d', n_close)
+    end
+
+    local close = string.find(body, '--outerbnd--', 1, true)
+
+    if not close then
+      problems[#problems + 1] = 'top-level boundary is never closed'
+    else
+      for _, marker in ipairs(markers) do
+        local pos = string.find(body, marker, 1, true)
+
+        if not pos then
+          problems[#problems + 1] = string.format('part %s is missing entirely',
+              marker)
+        elseif pos > close then
+          problems[#problems + 1] = string.format(
+              'part %s lands after the top-level closing delimiter at %d/%d, ' ..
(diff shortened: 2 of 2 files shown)


smpaz7467:
Note on the one red check, in case it saves someone a look: `webui-e2e-playwright / e2e (1.45.3, legacy, 22)` fails with

```
Error: Throughput counter should be one of [2, 3, 4, 5], got 10
```

That is a WebUI throughput-graph assertion, and this change touches only `lualib/lua_mime.lua` and `test/lua/unit/` — no web UI code. The other Playwright variant (`latest, latest`) passed on the same commit, as did all six platform `test` jobs and all three linters. The same legacy job also fails on unrelated branches while `master` is green, so it reads as a flake in that pinned Playwright/Node combination rather than anything here.

Happy to rebase or re-run if you would rather see a clean board — I did not want to force-push over an otherwise green build just for that.

vstakhov:
It's all good, thank you!

PR #6175, opened 2026-08-07
[Fix] mime: CED false positive on computed parts

moisseev:
PDF text extraction injects a synthetic text part whose bytes are not
valid UTF-8 and belong to no legacy charset. Running CED on them produced
a non-IANA name ("BIG5-CP950") that ucnv_open rejected (U_FILE_ACCESS_ERROR).

- Keep computed/extracted text parts raw instead of CED re-detection
  and conversion, so undecodable glyph data is excluded from tokenization
  and language detection.
- Route the CED-detected charset through rspamd_mime_detect_charset and
  map "big5-cp950" to "big5" so legitimate Big5 content converts.


moisseev:
@review

vstakhov:
1. The COMPUTED guard regresses legacy-encoded PDF text (main issue)

pdf.lua handles UTF-16BE/LE itself, but /ToUnicode CMap processing is commented out as NYI (lualib/lua_content/pdf.lua:823), and font /Encoding (WinAnsi/MacRoman) is never applied. So for a simple-font PDF, the extracted bytes are CP1252/CP1251/etc. text — not glyph garbage. That text used to be detected and converted; now it isn't.

Reproduced with a temporary suite on the branch (real CP1252 French text, ~600 bytes, injected exactly like inject_part does):

computed:     utf flag = 0   real_charset = UTF-8        ← kept raw (this PR)
non-computed: utf flag = 1   real_charset = ISO-8859-1   ← detected + converted

The two differ only by the RSPAMD_MIME_PART_COMPUTED bit. So every accented-Latin, Cyrillic or Greek PDF whose fonts lack ToUnicode loses UTF-8 conversion — that's a much more common corpus than CP950-extension Big5.

Also note the guard fires on !valid_utf8, and valid_utf8 is force-reset to FALSE at mime_encoding.c:838 for any declared non-UTF-8 charset. Today inject_part hardcodes charset=utf-8 so it can't bite, but if inject_part ever gains a charset argument, computed parts would stop converting even with a correctly declared charset. If the blanket policy is intended, it'd be worth stating it as "inject_part declares utf-8; content that isn't utf-8 is the producer's bug" rather than leaving it implicit in the control flow.

Options, roughly in order of how surgical they are:
- Drop the guard and rely on the alias fix alone — but then glyph garbage converts as Big5 into mojibake, which is why you added it. So: not that.
- Keep CED for computed parts but require reliability: ced_encoding_detect already returns is_reliable, and rspamd_mime_charset_find_by_content (mime_encoding.c:611) throws it away. Plumbing it through would separate "confident windows-1252" from "coin-flip Big5 on glyph noise" — this is the fix I'd suggest.
- Or gate the guard on the CJK/multi-byte case specifically, where mojibake from glyph noise is the actual failure mode.

Sorry for large LLM generated output, but I've seen utf16 pdf files sadly quite often.

PR #6176, opened 2026-08-08
[Fix] hfilter: compare HELO addresses against the connecting IP

### LINKED ISSUE #6165: [BUG] HFILTER_HELO_IP_A never performs the intended IP comparison
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:

**Describe the bug**

The description for `HFILTER_HELO_IP_A` is `"Helo A IP != hostname IP"`. However, it currently only gets added when the HELO domain doesn't have either an A or AAAA record. It has a comparison of the resolved addressed against the hostname IP but it only runs it inside of the branch handling neither A or AAAA records being available. Here's the relevant code:

```lua
    if failed_address >= 2 then
      -- No A or AAAA records
      if eq_ip and eq_ip ~= '' then
        for _, result in pairs(resolved_address) do
          if result == eq_ip then
            return true
          end
        end
        task:insert_result('HFILTER_' .. symbol_suffix .. '_IP_A', 1.0, host)
      end
      task:get_resolver():resolve_mx({
        task = task,
        name = host,
        callback = check_host_cb_mx
      })
    end
```

The comparison against the resolved addresses is clearly meant to be in an `else` branch where A and/or AAAA resolution succeeded and `HFILTER_HELO_IP_A` still gets added if neither matches.

I think this is what's intended to be there:

```lua
    if failed_address >= 2 then
      -- No A or AAAA records
      if eq_ip and eq_ip ~= '' then
        task:insert_result('HFILTER_' .. symbol_suffix .. '_IP_A', 1.0, host)
      end
      task:get_resolver():resolve_mx({
        task = task,
        name = host,
        callback = check_host_cb_mx
      })
    else
      if eq_ip and eq_ip ~= '' then
        for _, result in pairs(resolved_address) do
          if result == eq_ip then
            return true
          end
        end
        task:insert_result('HFILTER_' .. symbol_suffix .. '_IP_A', 1.0, host)
      end    
    end
```

This would add the symbol when there's a mismatch while still adding it when neither A or AAAA resolve.

I could submit a pull request for this myself but I don't have an existing setup for testing it.

### Steps to Reproduce

1. Send an email visible to Rspamd using a HELO mismatched A and AAAA records
2.  `HFILTER_HELO_IP_A` is not added

**Expected behavior**

`HFILTER_HELO_IP_A` should be added when IP != (A or AAAA).

### Versions

I've checked that the issue is still present in the `master` branch.

```
Rspamd daemon version 4.1.1

CPU architecture x86_64; features: avx2, avx, sse2, sse3, ssse3, sse4.1, sse4.2, rdrand
Hyperscan enabled: TRUE
Jemalloc enabled: TRUE
LuaJIT enabled: TRUE (LuaJIT version: LuaJIT 2.1.1780076327)
ASAN enabled: FALSE
BLAS enabled: TRUE
Fasttext enabled: TRUE (built-in)
```

smpaz7467:
Fixes #6165, reported by @thestinger, whose diagnosis located the problem exactly.

## The defect

`HFILTER_HELO_IP_A` is described as `"Helo A IP != hostname IP"`, but in `check_host_cb_a()` the comparison that would detect a mismatch sits inside the `failed_address >= 2` branch:

```lua
if failed_address >= 2 then
  -- No A or AAAA records
  if eq_ip and eq_ip ~= '' then
    for _, result in pairs(resolved_address) do
      if result == eq_ip then
        return true
      end
    end
    task:insert_result('HFILTER_' .. symbol_suffix .. '_IP_A', 1.0, host)
  end
```

That branch is reached only when both the A and the AAAA lookup failed, and `resolved_address` is appended to only when a lookup *succeeds*. So in the one branch that reads it, the list is necessarily empty: the loop body never executes and its `return true` is unreachable.

The symbol therefore fires **if and only if the HELO host resolves to neither A nor AAAA** — a no-resolve condition already covered by `HFILTER_HELO_NORES_A_OR_MX` — and never on the address mismatch it claims to report.

## Why moving the comparison to an `else` branch is not enough

This is the part worth flagging, since it is the natural reading of the code and it is what the issue proposes.

`check_host_cb_a` is the **shared** callback for both lookups:

```lua
task:get_resolver():resolve('a',    { task = task, name = host, callback = check_host_cb_a })
task:get_resolver():resolve('aaaa', { task = task, name = host, callback = check_host_cb_a })
```

It runs twice, in arbitrary order. An `else` branch therefore executes on the *first* callback to return, when `resolved_address` holds at most one address family. I built that variant and ran it against the tests added here — 1 of 6 passed:

- **A matching host gets flagged.** With `match.test` publishing `A=1.2.3.4` and the client connecting from `1.2.3.4`, the symbol was inserted: the AAAA NXDOMAIN callback returned first, the `else` branch ran against a still-empty `resolved_address`, and the A callback matched too late to prevent it. This is the ordinary IPv4-only-HELO case. It depends on completion order, so it is a race rather than a certainty.
- **The score doubles.** On a genuine mismatch both callbacks reach the insert, so the symbol lands twice — 2.0 instead of 1.0 in the runs here. That part is order-independent; the hfilter group sets no `one_shot` and `DEFAULT_MAX_SHOTS` is 100.

## What this changes

`failed_address >= 2` was implicitly serving as a "both lookups have finished" join. This makes that explicit with a completion counter, performs the comparison once both callbacks have returned, and leaves the MX fallback on the both-failed path.

A HELO that resolves to no addresses at all is **still flagged**, exactly as today. Whether that should remain `HFILTER_HELO_IP_A`'s job rather than `HFILTER_HELO_NORES_A_OR_MX`'s is a scoring question affecting every install, so I have not changed it — a test pins the current behaviour, and it is easy to flip if you would prefer the symbol to stay silent when there is nothing to compare against.

## Tests

hfilter had no functional tests, so this adds the first ones, using `fake_records` for deterministic DNS: mismatch, match, dual-stack matching on AAAA, dual-stack matching on A, dual-stack matching neither, and the no-records case.

Two notes on how they are built:

- The dual-stack cases exist specifically to pin the callback-ordering behaviour described above; they are the ones the `else`-branch variant fails.
- `hfilter.lua` registers its symbols at score `0.0` and relies on `conf/scores.d/hfilter_group.conf` for real weights, which the functional configs do not load, so the test config mirrors the weights in a `group "hfilter"` block. Without that, the score assertions would compare `0.0` against `0.0` and the double-insertion check would be vacuous.

## Verification

Against `master` (`0a8e795a`), building and swapping only `src/plugins/lua/hfilter.lua` between runs (md5 confirmed to change each time):

| Variant | 171_hfilter_helo |
|---|---|
| unmodified `master` | 4 passed, 2 failed |
| `else`-branch variant from the issue | 1 passed, 5 failed |
| this PR | 6 passed, 0 failed |

The two failures on unmodified `master` are the mismatch cases. The no-records case *passes* there, which is what demonstrates the comparison is dead rather than merely misplaced.

Full functional suite with this change, both CI phases: **785 + 101 = 886 tests, 0 failures**. `luacheck -q --no-color` on the modified file via `pipelinecomponents/luacheck`: 0 warnings, 0 errors.

## Not addressed

`check_host_cb_mx` counts `failed_mx_address` across a shared callback in a similar way, once per MX target. I have not reproduced a defect there and have left it alone rather than change behaviour speculatively.

### DIFF
--- a/src/plugins/lua/hfilter.lua
+++ b/src/plugins/lua/hfilter.lua
@@ -225,6 +225,7 @@ end
 local function check_host(task, host, symbol_suffix, eq_ip, eq_host)
   local failed_address = 0
   local completed_address = 0
+  local address_lookup_failed = false
   local resolved_address = {}
 
   local function check_host_cb_mx(_, to_resolve, results, err)
@@ -264,9 +265,15 @@ local function check_host(task, host, symbol_suffix, eq_ip, eq_host)
       end
     end
   end
-  local function check_host_cb_a(_, _, results)
+  local function check_host_cb_a(_, to_resolve, results, err)
     if not results then
-      failed_address = failed_address + 1
+      if err and (err ~= 'requested record is not found' and
+          err ~= 'no records with this name') then
+        lua_util.debugm(N, task, 'error looking up %s: %s', to_resolve, err)
+        address_lookup_failed = true
+      else
+        failed_address = failed_address + 1
+      end
     else
       for _, result in pairs(results) do
         table.insert(resolved_address, result:to_string())
@@ -283,7 +290,7 @@ local function check_host(task, host, symbol_suffix, eq_ip, eq_host)
       return
     end
 
-    if eq_ip and eq_ip ~= '' then
+    if not address_lookup_failed and eq_ip and eq_ip ~= '' then
       local matched = false
       for _, result in pairs(resolved_address) do
         if result == eq_ip then
@@ -298,7 +305,7 @@ local function check_host(task, host, symbol_suffix, eq_ip, eq_host)
       end
     end
 
-    if failed_address >= 2 then
+    if not address_lookup_failed and failed_address >= 2 then
       -- No A or AAAA records
       task:get_resolver():resolve_mx({
         task = task,
--- a/src/plugins/lua/hfilter.lua
+++ b/src/plugins/lua/hfilter.lua
@@ -224,6 +224,7 @@ end
 -- eq_host: host for comparing or empty string
 local function check_host(task, host, symbol_suffix, eq_ip, eq_host)
   local failed_address = 0
+  local completed_address = 0
   local resolved_address = {}
 
   local function check_host_cb_mx(_, to_resolve, results, err)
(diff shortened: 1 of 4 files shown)


thestinger:
Thanks for working on this.

You can change the Reported-by to `Reported-by: Daniel Micay <daniel.micay@grapheneos.org>` instead of using a username.

smpaz7467:
Done — amended to `Reported-by: Daniel Micay <daniel.micay@grapheneos.org>`. Force-pushed, so the PR now points at 5f5c176; the diff is unchanged from the previous revision, commit message only.

Issue #6190, opened 2026-08-15
[BUG] HFILTER_HOSTNAME_UNKNOWN is missing forward-confirmed DNS check due to misc.lua rDNS fallback


thestinger:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Read about bug reporting in general: https://rspamd.com/doc/faq.html#how-to-report-bugs-found-in-rspamd
    * Enabled relevant debugging logs: https://rspamd.com/doc/faq.html#how-to-debug-some-module-in-rspamd 
    * Checked the FAQs about Core files in case of fatal crash: https://rspamd.com/doc/faq.html#how-to-figure-out-why-rspamd-process-crashed
    * Tried ASAN package and obtained the ASAN report (if possible): https://rspamd.com/doc/faq.html#asan-builds 
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aissue+user%3Arspamd
    * Checked that there is not already an experimental package or master branch

**Describe the bug**

The description for `HFILTER_HOSTNAME_UNKNOWN` is "Unknown client hostname (PTR or FCrDNS verification failed)". However, it currently only ends up being inserted when rDNS is missing rather than when forward-confirmed DNS fails.

Rspamd normally receives the hostname from the MTA. For the milter interface, it's only supposed to be passed if forward-confirmed DNS is successful. That would make `HFILTER_HOSTNAME_UNKNOWN` work correctly with an MTA handling this properly, but Rspamd fills in the missing value itself in `rules/misc.lua` without doing forward-confirmed DNS. The code in `rules/misc.lua` adds `RDNS_DNSFAIL` or `RDNS_NONE` for those failures but then simply sets the hostname value `HFILTER_HOSTNAME_UNKNOWN` assumes has been verified.

In `src/lua/lua_task.c`, `lua_task_get_hostname` has a branch for a hostname starting with `[` which it documents is the bare IP enclosed in square brackets if the MTA either couldn't resolve the hostname or it failed forward-confirmed reverse DNS verification. The `set_hostname` function is therefore being used incorrectly by `rules/misc.lua` by not performing FCrDNS verification prior to setting it.

### Steps to Reproduce

1. Send email to Postfix + Rspamd from an IP with a reverse DNS name not matching the hostname
2. Postfix will pass the bare IP to Rspamd as `[A.B.C.D]` which will make get_hostname return nil
[...]

PR #6191, opened 2026-08-16
[Fix] misc: verify FCrDNS before adopting a resolved PTR name as hostname

smpaz7467:
Fixes the bug reported in #6190, and adds the temporary-failure symbol asked for in the same thread as a follow-up to #6165.

## The bug

`RDNS_CHECK` in `rules/misc.lua` fills in the hostname when the MTA passed none, straight from the PTR answer and with no forward confirmation:

```lua
task:set_hostname(results[1])
```

An MTA publishes a client hostname only once it has confirmed that the name resolves back to the connecting address. Per the milter interface a name that fails that check arrives as the bare IP in square brackets instead, which `lua_task_get_hostname()` deliberately reports as `nil` — the comment in `src/lua/lua_task.c` quotes the milter documentation for exactly this. So the fallback runs on precisely the clients the MTA has just judged unverifiable, and replaces that verdict with an unverified name.

Two consumers depend on the contract it breaks:

* **`HFILTER_HOSTNAME_UNKNOWN`**, described as *"Unknown client hostname (PTR or FCrDNS verification failed)"*, is inserted only in the `else` branch of `if hostname` in `hfilter.lua`. Once the fallback has installed a name, the symbol can never fire for the FCrDNS half of what it documents.
* **`fuzzy_sf_ptr()`** in `src/plugins/fuzzy_check.c` reasons that *"a plain name here means forward confirmed"* when producing `RSPAMD_FUZZY_SF_PTR_CONFIRMED`, and notes that `RSPAMD_FUZZY_SF_PTR_PRESENT` is reserved *"for the day rspamd resolves the PTR record on its own"*. That day has already arrived in `misc.lua`, without the verification, so the shingle feature is currently told "confirmed" about names nobody checked.

## The fix

Resolve `A` and `AAAA` for the PTR name and compare against the client address before calling `set_hostname()`.

The two lookups share a callback and complete in arbitrary order, so the verdict is taken only once both have returned — otherwise an address published in the slower family is not visible yet and a host that *does* confirm gets rejected depending on DNS timing. This is the same hazard as in `check_host_cb_a`. The comparison is on the printed forms because `ip:equal()` compares ports as well, and the client address carries one whereas a resolved address does not.

* **Verified** — hostname is installed, exactly as before.
* **Provably not confirmed** (mismatch, or no address records) — the name is not installed, leaving the task in the state it would have been in had the MTA done the check itself, plus `RDNS_FCRDNS_FAIL` for visibility.
* **Resolver failure** — not evidence against the client, so the PTR keeps being trusted as before and `RDNS_FCRDNS_DNSFAIL` is published instead. This is what gives a deployment enforcing `HFILTER_HOSTNAME_UNKNOWN` something to soft reject on rather than treating a timeout as a pass.

Both new symbols are scored 0.0: the weight for an unverified client already comes from `HFILTER_HOSTNAME_UNKNOWN`, and scoring here too would count the same fact twice.

## One deliberate behaviour change — worth your ruling

A name that fails verification is not installed, so hfilter's generic/dynamic **hostname pattern checks no longer score it either**. On the shipped weights, a dynamic-pool client with a generic PTR that fails FCrDNS loses `HFILTER_HOSTNAME_5` (3.0) and gains `HFILTER_HOSTNAME_UNKNOWN` (2.5) — a net −0.5.

I chose this because it is identical to what already happens whenever the MTA performs the check itself and sends the bracketed literal: in that case rspamd never sees the name and never pattern-scores it. The fallback path was the inconsistent one. Both halves are pinned by tests so the behaviour cannot drift by accident.

The alternative would be to install the name anyway for pattern scoring and carry the FCrDNS verdict separately, which nets +2.5 instead. That is a scoring decision for you rather than something to change unilaterally, so I have left it as the consistent option — happy to rework it if you prefer the other.

## Second commit: transient DNS failures in `check_host()`

`check_host()` suppresses every verdict when an address lookup fails transiently — `HFILTER_<suffix>_IP_A` is not inserted, and the MX fallback behind `NORES_A_OR_MX` / `NORESOLVE_MX` is not attempted. That is correct, but it leaves silence as the only outcome, and silence is indistinguishable from a clean pass. A deployment enforcing those symbols cannot tell "verified, no problem" from "could not verify", so a resolver outage quietly degrades into accepting everything.

`HFILTER_<suffix>_DNSFAIL` is published for that case, inserted after the both-callbacks join so a host whose `A` and `AAAA` both fail is reported once rather than once per callback. NXDOMAIN does not qualify — that is an answer, already reported as `HFILTER_<suffix>_NORES_A_OR_MX`. Scored 0.0.

This is the follow-up @thestinger asked for in #6190 against #6165. It is a separate commit and can be dropped or split out if you would rather keep this PR to the hostname fix alone.

## Verification

Built and run against current master (`6d84be9ab`, reports 4.1.6).

**New suite `172_rdns_fcrdns.robot`, covering `RDNS_CHECK`, which had no functional tests at all — 4 pass / 5 fail on stock master, 9/9 with the fix.** Four of the five stock failures are literally `Symbol HFILTER_HOSTNAME_UNKNOWN wasn't found in result`, i.e. the reported bug, reproduced through the mismatch case, the bracketed-IP-literal case an MTA actually sends, the no-address-records case, and the pattern-scoring case.

The suite deliberately opens with a control — a client with no PTR at all must still reach `RDNS_NONE` — which passes on stock, proving the prefilter really runs and the other cases are not green for free. The other controls (forward-confirmed client, dual-stack confirming on the slower family, generic name that *does* confirm) also pass on stock, so the fix is not simply always inserting.

| | stock master | with fix |
|---|---|---|
| `172_rdns_fcrdns` (new) | 4 / 9 | **9 / 9** |
| `171_hfilter_helo` (+4 new cases) | 9 / 11 | **11 / 11** |
| full functional, pabot + serial | — | **907 / 907, 0 failed** |
| `ninja test` | — | **8 / 8** |
| luacheck (`pipelinecomponents/luacheck`, `-q --no-color .`) | — | **0 warnings / 0 errors**, 262 files |

The both-families-fail case asserts on the *score* rather than presence, so an insertion made once per resolver callback shows as 2.00 and fails.

Issue: #6190

### DIFF
--- a/conf/scores.d/hfilter_group.conf
+++ b/conf/scores.d/hfilter_group.conf
@@ -82,6 +82,10 @@ symbols = {
         weight = 2.0;
         description = "Helo not FQDN";
     }
+    "HFILTER_HELO_DNSFAIL" {
+        weight = 0.0;
+        description = "Helo address lookup DNS error";
+    }
     "HFILTER_FROMHOST_NORESOLVE_MX" {
         weight = 0.5;
         description = "MX found in FROM host and no resolve";
@@ -93,6 +97,10 @@ symbols = {
     "HFILTER_FROMHOST_NOT_FQDN" {
         weight = 3.0;
         description = "FROM host not FQDN";
+    }
+    "HFILTER_FROMHOST_DNSFAIL" {
+        weight = 0.0;
+        description = "FROM host address lookup DNS error";
     }
     "HFILTER_FROM_BOUNCE" {
         weight = 0.0;
@@ -112,6 +120,10 @@ symbols = {
     "HFILTER_MID_NOT_FQDN" {
         weight = 0.5;
         description = "Message-id host not FQDN";
+    }
+    "HFILTER_MID_DNSFAIL" {
+        weight = 0.0;
+        description = "Message-id host address lookup DNS error";
     }
 */
     "HFILTER_HOSTNAME_UNKNOWN" {
--- a/conf/scores.d/hfilter_group.conf
+++ b/conf/scores.d/hfilter_group.conf
@@ -138,4 +138,12 @@ symbols = {
         weight = 0.0;
         description = "PTR verification DNS error";
     }
+    "RDNS_FCRDNS_FAIL" {
+        weight = 0.0;
+        description = "PTR name does not resolve back to the sender's IP";
+    }
+    "RDNS_FCRDNS_DNSFAIL" {
+        weight = 0.0;
+        description = "FCrDNS verification DNS error";
+    }
 }
--- a/rules/misc.lua
+++ b/rules/misc.lua
@@ -807,6 +807,74 @@ rspamd_config.COMPLETELY_EMPTY = {
 -- Preserve compatibility
 local rdns_auth_and_local_conf = lua_util.config_check_local_or_authed(rspamd_config, 'once_received',
   false, false)
+
+-- An MTA is expected to publish a client hostname only after forward
+-- confirmed reverse DNS: per the milter interface a name that fails that
+-- check is reported as the bare IP in square brackets instead, which
+-- `task:get_hostname()` deliberately reports as nil. When Rspamd resolves the
+-- PTR itself it has to perform the same verification, otherwise an
+-- unverified name silently replaces the MTA's verdict and
+-- HFILTER_HOSTNAME_UNKNOWN, documented as covering exactly this ("PTR or
+-- FCrDNS verification failed"), can never be inserted.
+local function rdns_verify_forward(task, task_ip, ptr_name)
+  local ip_str = task_ip:to_string()
+  local completed = 0
+  local matched = false
+  local tempfail = false
+
+  local function forward_dns_cb(_, to_resolve, results, err)
+    if err and (err ~= 'requested record is not found' and
+          err ~= 'no records with this name') then
+      rspamd_logger.infox(task, 'error verifying PTR name %s: %s', to_resolve, err)
+      tempfail = true
+    elseif results then
+      for _, addr in ipairs(results) do
+        -- Compare printed forms: `ip:equal()` compares ports as well, and
+        -- the client address carries one whereas a resolved address does not
+        if addr:to_string() == ip_str then
+          matched = true
+        end
+      end
+    end
+
+    -- The A and the AAAA lookups share this callback and complete in
+    -- arbitrary order, so judge only once both have returned: an address
+    -- published in the slower family is not visible yet and a host that does
+    -- confirm would be rejected depending on DNS timing
+    completed = completed + 1
+    if completed < 2 then
+      return
+    end
+
+    if matched then
+      rspamd_logger.infox(task, 'source hostname has not been passed to Rspamd from MTA, ' ..
+        'but we could resolve source IP address PTR as "%s"', ptr_name)
+      task:set_hostname(ptr_name)
+    elseif tempfail then
+      -- Verification could not be completed, which is not evidence against
+      -- the client, so keep trusting the PTR as before. The symbol lets a
+      -- deployment that enforces HFILTER_HOSTNAME_UNKNOWN soft reject these
+      -- rather than treat a resolver failure as a pass
+      task:insert_result('RDNS_FCRDNS_DNSFAIL', 1.0, ptr_name)
+      task:set_hostname(ptr_name)
+    else
+      -- The name provably does not belong to this client, so it is not set:
+      -- this is the same state the task would be in had the MTA done the
+      -- verification itself
+      task:insert_result('RDNS_FCRDNS_FAIL', 1.0, ptr_name)
+    end
+  end
+
+  for _, rr in ipairs({ 'a', 'aaaa' }) do
+    task:get_resolver():resolve(rr, {
+      task = task,
+      name = ptr_name,
+      callback = forward_dns_cb,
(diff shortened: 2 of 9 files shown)


vstakhov:
I'm still not convinced that it should be Rspamd job. Sorry, this AI slope is too hard to read but I see no good reason for this PR to be merged.

smpaz7467:
Fair enough on the writing, I'll keep it shorter.

The one thing I'd note: misc.lua already resolves the PTR and calls set_hostname() with the result, so rspamd is doing this today — the patch only adds the forward check before it trusts the name. Removing the fallback instead would make HFILTER_HOSTNAME_UNKNOWN fire for every client in deployments that pass no hostname at all, so that seemed worse.

Happy to close if you'd still rather not have it.

vstakhov:
The only reason why Rspamd performs PTR check was that Postfix that we've used as MTA was not fast/reliable in doing that, whilst hostname was useful in spam checks. Full FCrDNS loop is surely possible but we need to be sure when to enable it. I'm not convinced it should be 1) in hfilter; 2) enabled by default. In fact, the whole hfilter plugin has quite an outdated design and should be probably refactored.

thestinger:
@vstakhov I think the best solution would probably be splitting the hostname into the FCrDNS confirmed one from the MTA and one resolved by Rspamd without FCrDNS. No need to do FCrDNS again when the MTA already did it. The field should just be separate so that use cases expecting FCrDNS get the one from the MTA.

thestinger:
The advantage of doing FCrDNS in Rspamd would be the ability to differentiate a DNS lookup failure from failed verification. Force actions can be used to soft reject on DNS lookup failures while being able to do score-based reject or forced reject for the verification actually failing. A lot of spammers do retry on soft reject and often a lot more aggressively than legitimate senders. It's nice to tell them to go away for good which is why I wanted to handle this via Rspamd instead of `reject_unknown_client_hostname` in Postfix which must be a temporary error code for correctness.

smpaz7467:
@thestinger the fallback already only runs when the MTA passed no usable hostname (`if not task:get_hostname()`), so a name the MTA confirmed is never re-checked.

On differentiating the failures, that's what the two new symbols do: `RDNS_FCRDNS_FAIL` when the PTR resolves but no A/AAAA matches the client, which is permanent and safe to reject on, and `RDNS_FCRDNS_DNSFAIL` for SERVFAIL or timeout during the forward lookup, which is the soft reject case. `RDNS_NONE` and `RDNS_DNSFAIL` keep their meanings for the PTR lookup itself.

Splitting confirmed from unconfirmed is better than what I did. Right now an unverified name simply isn't installed, so it also stops feeding the `HFILTER_HOSTNAME_*` pattern checks. A separate field keeps those working and leaves `get_hostname()` meaning "confirmed", which `fuzzy_sf_ptr()` already assumes.

@vstakhov the verification is in rules/misc.lua next to the existing fallback. Commit 2 is the only hfilter.lua change and I'll drop it, so this adds nothing to hfilter's design debt either way. I can gate the forward check behind an option defaulting to off.

Want me to redo it as a separate unconfirmed-hostname field, option-gated, with nothing in hfilter?

thestinger:
@smpaz7467 Yes, the overall approach looks fine and would fix my problem. I haven't reviewed the details of the code though. I'm not familiar with Rspamd's codebase and haven't dealt with Lua for years so I'm not really capable of easily reviewing whether it's a good implementation.

What you're implementing would be the ideal approach from my perspective of wanting to `soft reject` on failure to resolve the PTR or perform FCrDNS while likely wanting to `reject` for an empty PTR or failed forward verification. I also want the soft reject to defer to score-based reject or other force_actions reject reasons which means it has to be in Rspamd, not Postfix.

Reject is a lot better at getting spammers to go away than soft reject where they not only often keep retrying for the current message but try more later including from new servers. I'm convinced that a lot of them take note of getting hard rejections and start leaving us alone as it's a waste of their time.

vstakhov:
> @vstakhov I think the best solution would probably be splitting the hostname into the FCrDNS confirmed one from the MTA and one resolved by Rspamd without FCrDNS. No need to do FCrDNS again when the MTA already did it. The field should just be separate so that use cases expecting FCrDNS get the one from the MTA.

Yeah, that was my point as well. I just think I need to rework the hfilter into some sane set of plugins/libraries, so we can configure that in a more sane way.

thestinger:
MX Check can fully replace the `FROMHOST` checks already but it isn't enabled by default.

I think the MX Check scores need an overhaul to make MAIL FROM worth a lot more than From and Reply-To along with eliminating the negative score rewarding having 3 unique domains for MAIL FROM, From and Reply-To. It really needs a lot more scoring changes than that. It could also have the main symbol split between MAIL FROM and HELO which could then replace the HELO MX check part of HFILTER.

The hfilter Message-ID FQDN check is better than the one outside it but that could just be moved. I don't see much reason to have the MX check for Message-ID since there's no particular reason to expect that. It's perfectly reasonable for it to be the FQDN of the server sending mail via a mail server which isn't a mail server itself. It's common to set it to the mail server's FQDN in clients but I don't think it's a spam signal for it to not be a mail server itself.

That would get rid of a lot of what's in hfilter. I do think it would be useful to be able to hard reject on failed forward-confirmed DNS but soft reject on DNS failures as this pull request implements but I'm not sure that it's the right way to do it.

PR #6196, opened 2026-08-20
[Fix] spf: return permerror when a domain has multiple SPF records

dragoangel:
Only the first record starting with the version string was taken and the rest of the answer ignored, so a domain publishing two SPF records was evaluated as if it published one. RFC 7208 4.5 requires permerror instead. An RRset carries no order, so which record won depended on the order the resolver happened to answer in: the same message could pass here and permerror at the next receiver.

Count the records rather than breaking out on the first one and set `permfail`, as is already done for the DNS limits. The duplicate branch reports the record as handled so that the caller does not fall back to `na`: an na element added there would mask the error, since na is checked before permerror when the result is dispatched.

Only records starting with the version string count, so a domain publishing one SPF record next to unrelated TXT records - vendor verification tokens and the like - is unaffected, which the second test added here guards.

### DIFF
--- a/src/libserver/spf.c
+++ b/src/libserver/spf.c
@@ -885,6 +885,7 @@ spf_process_txt_record(struct spf_record *rec, struct spf_resolved_element *reso
 {
 	struct rdns_reply_entry *elt, *selected = NULL;
 	gboolean ret = FALSE;
+	unsigned int nspf1 = 0;
 
 	/*
 	 * We prefer spf version 1 as other records are mostly likely garbage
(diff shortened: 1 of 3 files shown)


src/libserver/spf.c
```diff
@@ -894,17 +895,41 @@ spf_process_txt_record(struct spf_record *rec, struct spf_resolved_element *reso
 	{
 		if (elt->type == RDNS_REQUEST_TXT) {
 			if (strncmp(elt->content.txt.data, "v=spf1", sizeof("v=spf1") - 1) == 0) {
```
vstakhov: [P2] Match the complete SPF version section. This bytewise prefix check misses mixed-case SPF version sections and counts strings such as `v=spf10` as SPF1 records. RFC 7208 section 4.5 makes the literal case-insensitive and requires it to be followed by SP or end-of-record; it explicitly says `v=spf10` must be discarded. As written, mixed-case duplicates can escape permerror, while an unrelated future-version record can cause a false permerror. Please use an exact case-insensitive predicate and add regression cases for both forms. https://www.rfc-editor.org/rfc/rfc7208.html#section-4.5

vstakhov (changes requested):
Found one actionable protocol-correctness issue in the new SPF record-selection check. The multiple-record detection must match the complete SPF version section case-insensitively; details are in the inline comment.

Validation: the branch builds successfully; CTest passed 8/8, the Lua suite reported 1214/1217 passed with 0 failures and 3 unassertive tests, clang-format/diff/EOL checks passed, and the head commit is GitHub-verified. The local SPF Robot suite could not start because this macOS build lacks the configured hs_helper worker, so no SPF functional test executed locally.

src/libserver/spf.c
dragoangel: @vstakhov thank you for a review, will fix 👍

vstakhov: we have caseless string compare in str_util.h

dragoangel: Fixed: the check is now an exact caseless match on the `v=spf1` literal plus the SP-or-end-of-record terminator RFC 7208 4.5 requires, so a duplicate spelled `V=SPF1` permerrors and `v=spf10` no longer does. This also aligns the selection with `start_spf_parse`, which has always matched the version caselessly.

On `str_util.h`: I stayed with `g_ascii_strncasecmp` because `rspamd_lc_cmp` takes a length and rdns exposes TXT data as a bare NUL-terminated `char *` with no length (`union rdns_reply_element_un.txt`), so there is none to pass safely. It reads the first four bytes of both operands before comparing any of them, and the buffer is `malloc(datalen + 1)`, so a TXT string of one character or none - both legal - would be read out of bounds. That is also why `RSPAMD_LEN_CHECK_STARTS_WITH` guards on length first, and `g_ascii_strncasecmp` is what this file already uses for the same literal two functions down.

Regressions for both forms: `V=SPF1` alone -> `R_SPF_ALLOW`; `V=SPF1` + `v=spf1` -> `R_SPF_PERMFAIL`; `v=spf10` + `v=spf1` -> `R_SPF_ALLOW`, no permerror. 117_spf is 37/37 (34 before), CTest 8/8 in the ubuntu-ci image.

vstakhov:
Ok, this looks good indeed. The original logic was to filter the common junk like site verifications or other pre-historic DNS garbage. However, multiple *valid* or look-alike valid SPF records is a clear standard violation and a bad sign (or, sigh, a sign of another idiotic sender in the Internet).

PR #6197, opened 2026-08-20
[Feature] dkim: decide From alignment once, in the module that knows

dragoangel:
Whether a signature authorises the author rather than merely some domain was decided in two places. The dkim module holds the signing domain and knows whether the signature verified; dmarc recomputed the comparison over DKIM_TRACE options with its own eSLD lookup. Two implementations of one rule can only drift apart, and nothing outside dmarc could ask the question at all.

Decide it where the facts are. The dkim module compares each signing domain against the From domain and reports the best verdict it reached, `strict` for an exact match and `relaxed` when only the organisational domains agree, as the R_DKIM_ALIGNED symbol and in the task mempool. Signatures that only tempfailed are tracked apart from those that verified, since a policy must not report a definitive failure while an aligned signature is still unresolved.

dmarc consumes those facts and drops both of its comparisons, keeping the DKIM_TRACE walk only for the per-domain lists its reports need. The comparison itself is unchanged: `rspamd_util.get_tld` wraps `rspamd_url_find_tld`, which is what the module calls, and the From domain comes from the same `from_mime` field either way. The existing alignment tests pass untouched, which is the intended proof that no DMARC verdict moved.

Split the credit a passing signature earned between the two symbols rather than adding to it: R_DKIM_ALLOW goes from -0.2 to -0.1 and R_DKIM_ALIGNED takes the other -0.1. A signature aligned with the author is worth what a passing signature used to be worth on its own, while one that merely verifies for some unrelated domain now earns half of that.


src/plugins/dkim_check.c
```diff
@@ -1216,6 +1242,22 @@ dkim_module_check(struct dkim_check_result *res)
...
 		unsigned int nres = 0, i = 0;
+		/* 0 = none, 1 = relaxed, 2 = strict */
+		int best_align = 0, best_align_tempfail = 0;
+		const char *from_domain = NULL;
+		gsize from_domain_len = 0;
+		rspamd_ftok_t from_tld = {0};
+
+		if (MESSAGE_FIELD_CHECK(task, from_mime) &&
+			MESSAGE_FIELD(task, from_mime)->len > 0) {
+			struct rspamd_email_address *from_addr =
```
vstakhov: BTW, what will happen, if we have rewritten that from address somehow?

dragoangel: Good question, and the googlemail.com -> gmail.com rewrite is exactly the path
worth checking, since the aliases module does it via `task:set_from(..., 'alias')`.

Alignment is computed against `MESSAGE_FIELD(task, from_mime)[0]`. A rewrite cannot
move that entry: `lua_task_set_from` appends the new address and flags the
pre-existing ones with RSPAMD_EMAIL_ADDR_ORIGINAL, so index 0 remains the address
as parsed from the message. That is the same view DMARC has evaluated since
db9288928, which reads `get_from{'mime', 'orig'}`, so an alias rewrite changes
neither the alignment verdict nor the policy lookup.

src/plugins/dkim_check.c
```diff
@@ -1216,6 +1242,22 @@ dkim_module_check(struct dkim_check_result *res)
...
 		struct rspamd_dkim_check_result **pres;
 		unsigned int nres = 0, i = 0;
+		/* 0 = none, 1 = relaxed, 2 = strict */
+		int best_align = 0, best_align_tempfail = 0;
+		const char *from_domain = NULL;
+		gsize from_domain_len = 0;
+		rspamd_ftok_t from_tld = {0};
+
+		if (MESSAGE_FIELD_CHECK(task, from_mime) &&
+			MESSAGE_FIELD(task, from_mime)->len > 0) {
```
vstakhov, line 1258: [P1] Require an unambiguous original From before publishing alignment. DMARC explicitly rejects `(from or E)[2]` as `Duplicate From header`, while this accepts any non-empty array and chooses element 0. A message with multiple From fields/addresses can therefore receive `R_DKIM_ALIGNED` and the `dkim_alignment` mempool verdict even though it has no single RFC5322.From author domain. Since this is now a reusable authentication signal (and has a negative default score), please gate it on exactly one original From address, matching DMARC's predicate, and cover duplicate/multi-address From cases. The rewrite case remains compatible because `task:set_from(..., 'alias')` preserves the original entry and appends the alias.

test/functional/cases/001_merged/116_dkim_alignment.robot
```diff
@@ -0,0 +1,39 @@
...
+  ...  From domain is aligned in the relaxed sense only
+  Scan File  ${RSPAMD_TESTDIR}/messages/dmarc/onsubdomain_fail_alignment.eml
+  ...  Settings=${SETTINGS_DKIM}
+  Expect Symbol  R_DKIM_ALLOW
+  Expect Symbol With Option  R_DKIM_ALIGNED  relaxed
+
+DKIM NOT ALIGNED
+  [Documentation]  A signature that verifies for an unrelated domain authorises
+  ...  nothing about the author, so no alignment is reported
+  Scan File  ${RSPAMD_TESTDIR}/messages/ed25519-broken.eml
```
vstakhov: [P1] This fixture does not test a verified but unaligned signature. `ed25519-broken.eml` changes the signed From domain from `football.example.com` to `evil.example.com`; the existing `Cases.116 Dkim / DKIM Verify ED25519 REJECT` control confirms that it produces `R_DKIM_REJECT`. Consequently this assertion passes without entering the new `DKIM_CONTINUE`/non-aligned path described by the test. Please use a fixture with a valid signature whose `d=` is unrelated to the From domain, assert `R_DKIM_ALLOW`, and then assert that `R_DKIM_ALIGNED` is absent.

vstakhov (changes requested):
The centralization itself looks sound and does not introduce large-data conversions: alignment compares bounded domain strings, and the TLD helper stays stack-backed for ordinary domains. The exact ARM64 head builds cleanly; all 390 C++ unit cases and the focused 26 DKIM/DMARC functional cases pass.

I found two gaps that should be fixed before merging. The new public alignment signal needs the same single-author precondition as DMARC, and the purported verified/non-aligned regression currently tests a rejected signature instead. Please also add an aligned-DKIM-tempfail case, since `dkim_alignment_tempfail` is a new cross-module contract and is otherwise uncovered.

Checks: `git diff --check`, clang-format diff check, and luacheck are clean. All four commits are GitHub-verified. GitHub currently reports no CI checks for this head.

dragoangel:
Thank you @vstakhov for review 🙏, will check what need to be changed and ship required changes 😊

dragoangel:
@vstakhov while building the multi-address From fixture for the alignment gate I ran into something adjacent. It is out of scope for this PR, but worth checking separately.

For a From field naming more than one mailbox, dmarc.lua bails out with `DMARC_NA` on `(from or E)[2]`. That is one of the two options RFC 7489 6.6.1 allows, but it lands as "no policy applies" rather than as a refusal. The section calls such messages "typically rejected", and where a _receiver does evaluate them_ it specifies applying the check "using each of those domains found in the RFC5322.From field as the Author Domain and apply the *most strict policy* selected among the checks that fail".

The consequence is that a strict policy can be side-stepped by adding a second address to From: the policy is never fetched, so `p=reject` never applies. What remains is score, `MULTIPLE_FROM` at `8.0` and `MULTIPLE_UNIQUE_HEADERS` at `7.0` for the duplicate-field variant, and score is not the refusal the domain asked for. A permissive threshold or a significant negative symbol (f.e. `IP_REPUTATION_HAM` or `BAYES_HAM` with high confidence) may absorbs it. For users who using force action based on DMARC its a gap I think.

Both fixes are within the RFC: evaluate every domain in the field and keep the most strict failing policy, or treat an ambiguous author as a failure of the policy that would otherwise apply. I can open a separate issue or PR if you think it is worth pursuing.

dragoangel:
For completeness, what this PR does in that case today: nothing is published. With an ambiguous author the comparison is never entered, so neither `dkim_alignment` nor `dkim_alignment_tempfail` is set and `R_DKIM_ALIGNED` is absent. A consumer therefore cannot tell "the author could not be determined" from "the signature is not aligned", which is the same shape as the `DMARC_NA` point above, one level down.

If that distinction is worth having, alignment can carry a separate state next to `strict` and `relaxed`, so a policy acts on it rather than inferring it from silence. The contract is introduced by this PR, so adding the state now costs nothing but agreeing - we want it and if yes - how to name it.

PR #6200, opened 2026-08-24
[Feature] metadata_exporter: add redis_list pusher

tomsommer:
Adds a `redis_list` pusher alongside the existing `redis_pubsub` and `redis_stream` backends: `RPUSH` of the formatted payload to `list_key` (same `per_recipient` semantics as `redis_stream`), plus an optional `max_len` that `LTRIM`s the list as a bounded-memory safety cap.

**Motivation:** Logstash's redis input consumes only *lists* and *channels* — no stream support — so feeding Logstash (a very common metadata destination) currently forces a choice between `redis_pubsub` (unbuffered: events published while consumers restart are lost, and multiple identical consumers each receive every event) and a custom `push` hook. A list gives buffered delivery with atomic competing consumers (BLPOP) using only stock components on both ends.

**Implementation:** mirrors `redis_stream` — same `backend_required_elements` validation (`list_key`), same `backend_check` redis-params handling, same error/`maybe_defer` semantics. This exact logic is currently running in production at a mail hosting provider (15 scanners, ~600k msg/day) as a custom `push` hook feeding Logstash via a Redis list; upstreaming it so it's a supported backend.

### DIFF
--- a/src/plugins/lua/metadata_exporter.lua
+++ b/src/plugins/lua/metadata_exporter.lua
@@ -545,12 +545,12 @@ local pushers = {
           -- recoverable, a duplicated export is not, so we log at error level
           -- and let the operator act.
           local trim_upstream
-          local function redis_ltrim_cb(err)
-            if err then
+          local function redis_ltrim_cb(trim_err)
+            if trim_err then
               rspamd_logger.errx(task, 'cannot trim list %s to %s entries on server %s: %s; '
                   .. 'max_len is NOT being enforced',
                 list_key, rule.max_len,
-                trim_upstream and trim_upstream:get_addr() or 'unknown', err)
+                trim_upstream and trim_upstream:get_addr() or 'unknown', trim_err)
             end
           end
           -- note: assigns (not redeclares) trim_upstream so the callback's
--- a/src/plugins/lua/metadata_exporter.lua
+++ b/src/plugins/lua/metadata_exporter.lua
@@ -525,6 +525,80 @@ local pushers = {
       read = false,
     })
   end,
+  redis_list = function(task, formatted, rule)
+    local function do_rpush(list_key)
+      local _, ret, upstream
+      local function redis_rpush_cb(err)
+        if err then
+          rspamd_logger.errx(task, 'got error %s when pushing to list on server %s',
+            err, upstream:get_addr())
+          return maybe_defer(task, rule)
+        end
+        if rule.max_len then
+          -- Trim failures must be loud: unlike XADD MAXLEN, RPUSH cannot cap
+          -- the list itself, so a persistently failing LTRIM silently defeats
+          -- max_len and the list grows without bound.
+          --
+          -- Deliberately NO defer here. The RPUSH above already succeeded, so
+          -- the message has been exported; deferring would re-run the whole
+          -- exporter on retry and push a duplicate. A breached memory cap is
+          -- recoverable, a duplicated export is not, so we log at error level
+          -- and let the operator act.
+          local trim_upstream
+          local function redis_ltrim_cb(err)
+            if err then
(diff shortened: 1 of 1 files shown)


src/plugins/lua/metadata_exporter.lua
```diff
@@ -525,6 +525,56 @@ local pushers = {
...
+        if rule.max_len then
+          lua_redis.redis_make_request(task,
+            redis_params,
+            nil,
+            true,
+            function()
+            end,
+            'LTRIM',
+            { list_key, tostring(-rule.max_len), '-1' }
+          )
```
vstakhov: [P2] Handle failures of the list-size cap

Both the LTRIM request result and callback error are ignored. Reproduced with max_len=2 and an ACL denying LTRIM: RPUSH succeeds for all five messages, the list reaches length five, and no exporter error or defer occurs. Persistent trim failures therefore silently defeat the advertised memory cap. Check scheduling and callback failures, define the defer behavior, and add regression coverage.

vstakhov (changes requested):
Reviewed 9bf3eb58759e5d07957b2a8acce71040fe7ecb65.

Requesting changes for the optional list-size cap's failure handling, detailed inline.

Separately, GitHub reports this commit as unsigned; repository rules require GPG-signed commits. Please sign the commit.

Validation: luacheck, git diff --check, and LF checks passed. A focused harness under LuaJIT and Lua 5.4 passed normal pushes, successful trimming, per-recipient routing, and RPUSH failure handling, and reproduced ignored LTRIM failures. The harness used real Redis-compatible commands against an isolated Valkey instance with mocked request scheduling; the full integration suite was not run.

src/plugins/lua/metadata_exporter.lua
vstakhov: In fact, I also need to check if our streams exporter is not doing the same.

tomsommer:
Thanks for the detailed review — the reproduction with an ACL denying LTRIM was exactly the case I had not thought about.

**Fixed**, and the commit is now signed (apologies for the earlier close/reopen churn — that was me resetting the branch to rebuild it as a single commit, which GitHub interpreted as an empty PR).

### LTRIM failure handling

Both the scheduling result and the callback error are now checked and logged at error level, naming the list, the cap and the server. The underlying asymmetry is that `XADD MAXLEN` caps the stream atomically, whereas `RPUSH` cannot cap anything, so the trim is a second command with a second failure path I had simply not written.

### Defer behaviour — a deliberate choice, and I may have it wrong

A trim failure **does not defer**. By the time it happens the `RPUSH` has already succeeded, so the message *is* exported; deferring would re-run the exporter on retry and push a duplicate. My reasoning was that a breached memory cap is recoverable and observable, whereas a duplicated export is neither.

That does mean a persistently failing trim is now loud rather than silent, but the cap is still breached. If you would rather it defer and accept possible duplicates, say so and I will change it — it is one line.

### The atomic alternative

The cleaner fix is arguably to do `RPUSH` + `LTRIM` in a single `EVAL` via `lua_redis.exec_redis_script`. That collapses it to one round trip and one error path, and makes defer unambiguously correct because either both commands happen or neither does — no duplicate risk.

I did not do that here because you asked for the minimal change and it is your call on scope. Happy to rework it that way if you prefer.

### Regression coverage

I have not added a test, and I would rather ask than guess at something unmergeable. The plugin reads `rspamd_config:get_all_opt` at load time, so it cannot be required from a unit test the way `test/lua/unit/metadata_exporter_structured.lua` works, and a functional test would need a Redis ACL fixture to deny LTRIM. Which shape would you like — a functional case with an ACL-restricted instance, or would you rather the pusher be refactored so the trim path is testable in isolation?

### On the streams exporter

For what it is worth, I do not think `redis_stream` has the same problem: `MAXLEN` is part of the `XADD` itself, so there is no separate command to fail. The bug is specific to needing two commands.

PR #6207, opened 2026-08-28
Bound UCL parsing of untrusted input and free trees iteratively

vstakhov:
Hardens UCL parsing of untrusted input. Two problems: freeing a parsed tree
was recursive, and nothing bounded how much tree a request could ask for.

## Stack exhaustion while freeing

`ucl_object_free_internal` descended into every child, so cleanup depth was
proportional to the nesting depth of the document. A deeply nested document
could therefore take the process down *during destruction* — after the parser
had already returned success.

Destruction now walks an explicit worklist threaded through the dying objects'
`next` pointers. That is safe because an object only reaches the worklist once
its refcount hits zero, meaning it is already detached and unobservable, so the
link field is free to reuse. Allocation free, constant stack.

Verified rather than assumed: with the old recursive destructor restored behind
a temporary flag, freeing a 20 000-level tree is a hard `SIGSEGV`; with the
worklist it frees cleanly. That case is now a test.

## Nesting was unbounded

`UCL_MAX_NESTING` existed but was only ever consulted by the msgpack and csexp
parsers — brace containers passed the dotted-key `level` through untouched and
were never depth-checked. Container depth is now tracked separately from
`level` and checked before anything is allocated, so the 1024 default applies
to every parse type.

## Parser budgets

New `ucl_parser_set_limits()` / `ucl_parser_get_limits()`:

| field | bounds |
| --- | --- |
| `max_depth` | nesting depth of objects and arrays |
| `max_nodes` | elements in the parsed tree |
| `max_alloc` | aggregate bytes attributed to the tree |
| `max_key_length` | length of a single key |
| `max_string_length` | length of a single string value |

Zero means unlimited, and only `max_depth` is on by default — the others would
reject large but perfectly valid inputs (maps, language models, big replies),
so callers handling untrusted data opt in.

Both the UCL and the MessagePack parser charge every element they create.
MessagePack matters here: a container costs one byte, so it amplifies harder
than JSON — 100 000 nesting levels fit in 100 KB.

## Applying it in rspamd

`rspamd_ucl_parser_new_untrusted(inlen)` sets depth 64, 1M elements, 1KB keys,
16MB strings and `min(96 * inlen + 256KB, 64MB)` of tree. The factor is derived,
not guessed: an element costs ~128 bytes and the densest JSON describing one is
two bytes (`1,`), so real input tops out near 64x. The absolute cap is what does
the work — today 50MB of `[1,1,1,...]` builds gigabytes of objects.

Wired into checkv3 metadata (JSON and msgpack), both controller body parsers and
both proxy parsers.

## Bugs fixed on the way

- `ucl_copy_or_store_ptr()` returned `false` instead of `-1` on allocation
  failure, leaving a value object with a stale union pointer and `len == 0`.
- The sibling-chain loop released a sibling's children *before* checking its
  refcount, so a shared object could have its children freed underneath it.
- `ucl_parser_append_elt()` dereferenced an unchecked allocation and ignored
  both `ucl_array_append()` results. It is also the `NO_IMPLICIT_ARRAYS` path
  that `SAFE_FLAGS` selects, so its wrapper node now counts against the budget.
- A NULL container was unconditionally reported as a syntax error, masking the
  real cause; msgpack's `CONSUME_RET` did the same to every `-1`.
- `ucl_msgpack_get_next_container()` unwound finished containers by tail
  recursion, making parse depth a stack-depth problem of its own.

## Testing

- 379/379 C++ unit tests, 27 of them new in `rspamd_cxx_unit_ucl_limits.hxx`
  covering each limit for both JSON and msgpack, deep-tree destruction, and
  shared subtrees outliving their container
- 1218 Lua tests, 0 failures
- 440/440 functional across `001_merged` (which carries `430_checkv3` and
  `420_content_negotiation`, the msgpack protocol tests), `121_json`,
  `140_proxy`, `220`/`221_http`, `560_metadata_exporter`
- `rspamadm configtest` on a full production config: `syntax OK`, so the 1024
  default does not disturb real configs
- Everything runs under ASAN

## Known gap

`ucl_sexp.c` still has its own hardcoded `UCL_MAX_NESTING` and manipulates
`parser->stack` directly without touching the depth counter. It is internally
consistent (it neither increments nor decrements, so nothing desyncs) and CSEXP
is not reachable from any of the network paths wired here, which only use
`UCL_PARSE_UCL` and `UCL_PARSE_MSGPACK`. Worth unifying if that ever changes.

### DIFF
--- a/contrib/libucl/ucl.h
+++ b/contrib/libucl/ucl.h
@@ -138,7 +138,8 @@ typedef enum ucl_error {
 	UCL_EMACRO,    /**< Error processing a macro */
 	UCL_EINTERNAL, /**< Internal unclassified error */
 	UCL_ESSL,      /**< SSL error */
-	UCL_EMERGE     /**< A merge error occurred */
+	UCL_EMERGE,    /**< A merge error occurred */
+	UCL_ELIMIT     /**< Input exceeds one of the configured parser limits */
 } ucl_error_t;
 
 /**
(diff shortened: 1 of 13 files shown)


vstakhov:
Also needs backporting to libucl...

PR #6213, opened 2026-08-30
Rework public suffix matching: suffix hash lookup and two-pass URL discovery

vstakhov:
This reworks how the public suffix list (`effective_tld_names.dat`) is matched. Previously one ~10.5k-pattern hyperscan database served two very different workloads: free-text URL discovery and per-host registrable-domain resolution. The per-host workload ran a full multipattern scan over every parsed hostname, reported a callback for every matching suffix and emulated longest-match by keeping the maximum, while `!` exception rules were skipped entirely.

## What changed

**New `tld_lookup` component** (`src/libserver/tld_lookup.{h,cxx}`): loads the suffix file once into flat hash maps and resolves the registrable domain by probing host label suffixes right to left (bounded by the deepest rule, at most 7 probes). It implements full public suffix list semantics: wildcard rules, `!` exception rules (previously a TODO) and the ICANN/private section split, and keeps the distinct final labels of all rules for discovery anchoring.

**Host resolution** (`rspamd_url_find_tld`, the URL parse path, and transitively `util.get_tld` and the HTML phishing checks) now uses direct hash lookups instead of scanning the host with the multipattern engine.

**Free-text URL discovery** is now two passes: the small static-matcher multipattern (~20 patterns, always compiled synchronously in memory) provides scheme/prefix candidates, and TLD candidates are anchored on `.<known final label>` occurrences found with a memchr-driven probe of the 1610 distinct final labels. Since the existing matcher logic only accepts TLD candidates at the end of a domain token, anchoring on the final label is equivalent to matching the full suffix. Candidates from both passes are merged in ascending match-end order, which the dedup logic of the processing callbacks relies on.

**Removed**: the url scanner no longer builds the largest hyperscan database in rspamd and no longer registers it in the hs_helper pending-compilation queue. `RSPAMD_MULTIPATTERN_TLD` stays in multipattern as it is exposed through the Lua trie API.

**Lua exposure**: the new `rspamd_tld_lookup` module queries the loaded suffix list (`registrable`, `suffix` with wildcard/exception/private flags, `is_final_label`) and builds custom suffix-like rule sets. The RBL URL composition maps (`lua_urls_compose`) now run on such a rule set instead of per-TLD regex buckets with optional hyperscan tries (−124 lines, same semantics, covered by the existing unit and functional tests).

## Performance

Standalone microbenchmark (vectorscan 5.4.11, Apple M-series, real suffix list, 200k synthetic hosts, 2.1 MiB corpus of the functional-test messages):

| workload | before | after |
|---|---|---|
| host → registrable domain | 632 ns (hyperscan), 158 ns (ACISM fallback) | **116 ns** |
| text scan | ~3.6 GB/s (~0.25 GB/s on domain-dense text) | **1.4–23 GB/s** |
| startup cost | 1.3 s compile, 6.5 MiB serialised db | ~5 ms, ~600 KiB |

Notably the ACISM "fallback" was already ~5x faster than the compiled hyperscan database for short host inputs — the fixed `hs_scan` overhead dominates there, so the async compilation machinery was actively counterproductive for this workload.

## Behaviour changes (deliberate)

1. `!` exception rules are honoured now. E.g. `foo.city.kobe.jp` resolves to `city.kobe.jp` (the exception cancels `*.kobe.jp`); previously the whole host was returned.
2. A host that is itself a public suffix resolves to the whole host instead of being attributed to a shorter suffix (e.g. bare `s3.amazonaws.com` no longer yields `amazonaws.com` for reputation purposes).

## Validation

- New doctest suite for the component (wildcards, exceptions, unicode + punycode, private section, edge cases).
- Full C++ (388 cases) and Lua (1227 tests) unit suites pass.
- Functional merged + URL suites: 444/444 pass.
- Differential `rspamadm mime urls` and `rspamadm mime urls -t` over all 182 functional-test messages against master: zero differences.

### DIFF
--- a/lualib/lua_urls_compose.lua
+++ b/lualib/lua_urls_compose.lua
@@ -1,5 +1,5 @@
 --[[
-Copyright (c) 2022, Vsevolod Stakhov <vsevolod@rspamd.com>
+Copyright (c) 2026, Vsevolod Stakhov <vsevolod@rspamd.com>
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


vstakhov:
```
┌───────────────────────────┬──────────────────────┬──────────────────────┬─────────────────────┬───────────┐
│         workload          │ hs-full (old engine) │ acism (old fallback) │ tldlkp (production) │ prototype │
├───────────────────────────┼──────────────────────┼──────────────────────┼─────────────────────┼───────────┤
│ host → registrable domain │ 770 ns               │ 164 ns               │ 174 ns (4.4×)       │ 141 ns    │
├───────────────────────────┼──────────────────────┼──────────────────────┼─────────────────────┼───────────┤
│ text scan, real corpus    │ 3.6 GB/s             │ 1.0 GB/s             │ 23 GB/s (6.5×)      │ 30 GB/s   │
├───────────────────────────┼──────────────────────┼──────────────────────┼─────────────────────┼───────────┤
│ text scan, domain-dense   │ 202 MB/s             │ 435 MB/s             │ 1.0 GB/s (5×)       │ 1.4 GB/s  │
├───────────────────────────┼──────────────────────┼──────────────────────┼─────────────────────┼───────────┤
│ load/compile              │ 1.3–6 s + 6.5 MiB db │ 27 ms                │ 5.3 ms              │ —         │
└───────────────────────────┴──────────────────────┴──────────────────────┴─────────────────────┴───────────┘
```

PR #6214, opened 2026-08-31
[Fix] Converge cold passive workers on HTTP maps quickly

vstakhov:
## Problem

Scanner workers never fetch HTTP maps themselves: only the primary controller does (`active_http`), and scanners wait for it to announce the downloaded map via the shared memory cache, polling that cache on their own periodic timers.

On a cold start (no cached map files on disk, so the pre-fork `rspamd_map_preload()` is a no-op) every scanner checks the shm cache exactly once at startup — before the controller has finished its first download — finds nothing, and then sleeps through a **full jittered poll interval** (`[poll_timeout, 2*poll_timeout)`, i.e. minutes at defaults) while serving no map data.

The observable result in production: sibling workers of the same instance disagree about multimap lookups for the first 1–2 poll intervals, each forked scanner independently cold and converging on its own jittered schedule, while the controller reports every map as loaded (the `shared->loaded`/`cached` flags are instance-wide shm flags set by whichever process loads first — they say nothing about per-worker state).

## Fix

A new per-process `map->has_data` flag tracks whether *this* process has delivered map data to its consumer (set in the periodic dtor after a successful fin callback, and in `rspamd_map_preload()` pre-fork so children inherit it, including maps loaded from fallback files):

* **Cold fast retry**: a passive worker with an HTTP map and no data reschedules its check every 2–4 s (jittered `min_timer_interval`) instead of a full poll interval, until the first successful load. Cheap: one atomic read per tick, and it stops as soon as data arrives.
* A cold worker that finds the shm cache available forces a read even when the `last_modified` comparison alone would skip it.
* **Disk cache fallback**: when the shm announcement is absent but `shared->cached` indicates the active worker has already saved the map cache file (e.g. its re-download swap window, or shm cache expiry), a cold passive worker loads directly from that file via `rspamd_map_read_http_cached_file()` instead of skipping the backend.

## Verification

* All unit tests pass (390 C++ cases, 1254 Lua tests).
* Live reproduction: fresh empty DBDIR, 2 scanner workers + controller, HTTP map server with a 5 s response delay. Before: scanners would sleep a full jittered poll interval after their startup miss. After: scanners cold-poll at 2–4 s and converge **1–2 seconds** after the controller publishes, then return to the normal cadence. Maps with shipped fallback files correctly never enter cold mode (preload marks them warm pre-fork).

### DIFF
--- a/src/libserver/maps/map.c
+++ b/src/libserver/maps/map.c
@@ -1395,6 +1395,11 @@ rspamd_map_periodic_dtor(struct map_periodic_cbdata *periodic)
 		/* Need to notify the real data structure */
 		periodic->map->fin_callback(&periodic->cbdata, periodic->map->user_data);
 
+		if (periodic->need_modify && !periodic->cbdata.errored &&
+			periodic->cbdata.cur_data != NULL) {
+			map->has_data = true;
+		}
+
 		if (map->on_load_function) {
 			map->on_load_function(map, map->on_load_ud);
 		}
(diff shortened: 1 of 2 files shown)


vstakhov:
@review deep

PR #6216, opened 2026-09-01
[Feature] Show git build id in WebUI cluster table

moisseev:
Expose the build-time git id (RID, -DGIT_ID builds) as "git_id" in /stat and render it in a dedicated WebUI cluster-table column, hidden when no live server reports one. Differing ids get a drift badge like version/configuration mismatches, so dev builds of different commits can be told apart in a cluster.


moisseev:
@review

PR #6218, opened 2026-09-01
[Feature] rbl: merge duplicate DNS queries produced by multiple checks

c-rosenberg:
### Summary

Requests that resolve to the same DNS query (case-insensitive, ignoring a
trailing dot) are now deduplicated and sent once instead of once per check.

By default, checks sharing a query also share a single symbol insertion:
their labels are combined into one comma-separated option and the symbol's
weight is counted only once.

### Behaviour change

Since `default_merge_checks` is `true`, this changes existing scoring: a rule
that previously scored once per check for the same value now scores once per
matched value instead. Set `merge_checks = false` (or `default_merge_checks =
false`) to keep the pre-4.1.6 behaviour of one insertion per check.

### Additional fixes

- Whitelisting now tracks every RBL rule that whitelists a given value
  instead of only the last one processed.
- `resolve_ip` carries origins/labels from all merged sender-side checks
  through to the second-phase query and to failure reports.
- Empty lookup values are skipped instead of being queried against the bare
  RBL suffix.

### Testing

Added functional coverage for merged DNS requests: `merge_checks` scoring and
disabling it, `symbols_prefixes` grouping, case-insensitive and trailing-dot
deduplication, `resolve_ip` origin propagation, and whitelisting of merged
requests.

### DIFF
--- a/src/plugins/lua/rbl.lua
+++ b/src/plugins/lua/rbl.lua
@@ -254,45 +254,81 @@ matchers.glob = function(_, to_match, _, map)
 end
 
 local function rbl_dns_process(task, rbl, to_resolve, results, err, resolve_table_elt, match)
-  local function make_option(ip, label)
+  local function make_option(origin, ip, label)
     if ip then
       return string.format('%s:%s:%s',
-          resolve_table_elt.orig,
+          origin,
           label,
           ip)
     else
       return string.format('%s:%s',
-          resolve_table_elt.orig,
+          origin,
           label)
     end
   end
 
-  local function insert_result(s, ip, label)
-    if rbl.symbols_prefixes then
-      local prefix = rbl.symbols_prefixes[label]
-
-      if not prefix then
-        rspamd_logger.warnx(task, 'unlisted symbol prefix for %s', label)
-        task:insert_result(s, 1.0, make_option(ip, label))
-      else
-        task:insert_result(prefix .. '_' .. s, 1.0, make_option(ip, label))
-      end
-    else
-      task:insert_result(s, 1.0, make_option(ip, label))
-    end
-  end
+  -- merge_checks false restores the pre-4.1.6 one-insertion-per-check behavior
+  local merge_checks = rbl.merge_checks ~= false
 
   local function insert_results(s, ip)
-    for label in pairs(resolve_table_elt.what) do
-      insert_result(s, ip, label)
+    -- Only the first insertion of a symbol carries score; later spellings just add options
+    local scored_symbols = {}
+
+    local function insert_origin_labels(sym_name, origin, labels)
+      table.sort(labels)
+
+      if merge_checks then
+        local weight = scored_symbols[sym_name] and 0.0 or 1.0
+        scored_symbols[sym_name] = true
+        task:insert_result(sym_name, weight,
+            make_option(origin, ip, table.concat(labels, ',')))
+      else
+        for _, label in ipairs(labels) do
+          task:insert_result(sym_name, 1.0, make_option(origin, ip, label))
+        end
+      end
+    end
+
+    for origin, origin_labels in pairs(resolve_table_elt.origins) do
+      if rbl.symbols_prefixes then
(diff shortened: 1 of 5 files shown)


vstakhov:
@review deep

src/plugins/lua/rbl.lua
```diff
@@ -254,45 +254,81 @@ matchers.glob = function(_, to_match, _, map)
...
   if err and (err ~= 'requested record is not found' and
       err ~= 'no records with this name') then
     rspamd_logger.infox(task, 'error looking up %s: %s', to_resolve, err)
-    task:insert_result(rbl.symbol .. '_FAIL', 1, string.format('%s:%s',
-        resolve_table_elt.orig, err))
+    -- Report every merged origin, but only the first insertion carries weight
+    local failed_origins = lua_util.keys(resolve_table_elt.origins)
+    table.sort(failed_origins)
+    for i, origin in ipairs(failed_origins) do
+      task:insert_result(rbl.symbol .. '_FAIL', i == 1 and 1.0 or 0.0,
```
vstakhov, line 329: [P2] Honor `merge_checks = false` for failure symbols too. This branch always gives score weight only to the first merged origin, independently of `merge_checks`, although line 270 documents the option as restoring one insertion per check. With the existing two-selector SERVFAIL fixture plus `merge_checks = false`, an exact ARM64 run reports both options but scores `RBL_MERGED_ERR_FAIL` as 1.0 instead of the pre-change/opt-out value 2.0. Please apply the same merge policy here and cover the failing-query opt-out case.

src/plugins/lua/rbl.lua
```diff
@@ -426,33 +475,37 @@ local function gen_rbl_callback(rule)
         if processed then
           nreq = {
             forced = forced,
-            n = processed,
+            n = tostring(processed):lower(),
```
vstakhov, line 478: [P2] Canonicalize `process_script` output consistently. The non-script path removes the DNS root dot, but this path only lowercases, and the final dedupe pass also only lowercases. A temporary rule whose script returns `example.org.test9.uribl` for one selector and the same name with a trailing dot for another issued two equivalent queries and scored the symbol 4.0 instead of 2.0. That defeats both the default scoring change and the DNS-query reduction for supported scripted rules. Please normalize the single root dot after `process_script` as well, ideally through one shared query-canonicalization helper.

src/plugins/lua/rbl.lua
```diff
@@ -426,33 +475,37 @@ local function gen_rbl_callback(rule)
...
             what = { [label] = true },
+            origins = { [req_str] = { [label] = true } },
           }
           requests_table[req] = nreq
         end
       else
         local to_resolve
-        local origin = req
+        -- Strip trailing dots first so hashing treats "foo." and "foo" alike
+        local origin = tostring(req):gsub('%.+$', '')
```
vstakhov, line 489: [P2] Do not sanitize invalid repeated dots into another valid name. `gsub('%.+$', '')` removes every trailing dot before `validate_dns`, so `malformed..` becomes `malformed` and bypasses the validator's explicit `..` rejection. In a negative control with only `malformed.test9.uribl` listed, the selector value `malformed..` produced `RBL_SELECTOR_MALFORMED`. Strip at most the one legal root dot (and re-check for an empty result) so malformed values cannot be reinterpreted as a different lookup.

vstakhov (changes requested):
I found three reproducible correctness gaps in the new canonicalization/scoring paths. The PR's focused suite passes 26/26 on an exact ARM64 build, both Lua files pass luacheck, diff-check is clean, all CI jobs are green, and both commits are GitHub-verified. I also audited the new conversions and aggregation: the built-in inputs are strings or small IP userdata, so I found no message-sized `tostring` materialization or performance blocker; the extra lowercase/dedup pass is over bounded DNS request names. The negative controls below fail specifically at the documented opt-out, `process_script`, and malformed-input boundaries.

PR #6219, opened 2026-09-01
[Feature] metadata_exporter: custom variables, MIME/multipart alerts, SMTP hardening, and rule schema validation

c-rosenberg:
### Summary

A batch of related `metadata_exporter` improvements built on top of `lua_smtp`
hardening, covering routing/templating flexibility, correct MIME output, and
config-time validation.

### Custom variables & selector expansion
- Add a `custom_variables` config block (Lua functions evaluated per message,
  like `custom_select`/`custom_format`/`custom_push`).
- Add `$name` / `${expr}` placeholder expansion for a small allowlist of
  per-message routing options: `mail_from`, `mail_to`, `helo`, `channel`,
  `stream_key`, plus the `email_alert` `email_template` body. `${expr}` falls
  back to a `lua_selectors` expression when no matching custom variable exists.
- Connection, credential, timeout and boolean options are never expanded, even
  if the configured value contains a literal `$`.
- Invalid addresses produced by `mail_from`/`mail_to` expansion are validated
  and dropped instead of sent; CR/LF in expanded values is collapsed to
  prevent header/protocol injection.
- Add `email_alert_sender_variable` to route an extra alert copy to an address
  produced by a named custom variable.

### MIME correctness for email_alert
- Auto-encode non-ASCII headers (RFC 2047) via the new
  `email_auto_encode_headers` rule setting (default `true`); address-list
  headers are parsed per-address so only the display name is encoded.
- Add `email_parts` / `email_parts_type` so an alert can be built as
  `multipart/<type>` with per-part `content_type`, `filename`, `disposition`,
  and transfer encoding (`auto`, `base64`, `quoted-printable`, `7bit`, `8bit`),
  with defensive encoding selection and RFC 2231 filename encoding.
- Auto-group inline, unnamed `text/plain` + `text/html` parts into a
  `multipart/alternative` (new `auto_grouping` option, default `true`, plain
  first per RFC 2046 5.1.4).

### Config validation
- Rules and `email_parts` are now schema-validated
  (`lualib/plugins/metadata_exporter.lua`) at config load time instead of
  ad-hoc checks, reporting failures via `push_config_error` so `rspamadm
  configtest` fails on invalid config.
- Include the rule name in debug log messages.

### lua_smtp hardening (shared with metadata_exporter's SMTP backend)
- Dot-stuff SMTP DATA content per RFC 5321 4.5.2, fixing silent truncation of
  message bodies containing a bare `.` line.
- Negotiate EHLO/8BITMIME with HELO fallback: read the EHLO response to
  completion before deciding to fall back, and match the 8BITMIME capability
  by exact keyword instead of substring search.
- Opt into separate connect/read/write timeouts (falling back to the overall
  timeout when unset), wired through metadata_exporter's `send_mail` backend.

### Supporting fix
- `lua_util.callback_from_string`: add `chunkname`/`return_chunk` params so
  whole-chunk callback snippets (used by `custom_variables`/`custom_select`/
  `custom_push`) compile correctly instead of being mis-wrapped as a single
  expression.

### Testing
- Functional coverage (`560_metadata_exporter_structured.robot`) for custom
  variable/selector expansion, non-ASCII header encoding, multipart
  `email_parts` assembly and dot-stuffing, EHLO/8BITMIME negotiation and HELO
  fallback, schema validation errors, content/content_from_variables, and
  multipart/alternative auto-grouping.
- Unit tests for the schema validator and `callback_from_string`.
- Fix a CPU-spin in the dummy SMTP test server helpers on shutdown.

### DIFF
--- a/lualib/lua_smtp.lua
+++ b/lualib/lua_smtp.lua
@@ -122,13 +122,72 @@ local function sendmail(opts, message, callback)
           return msg
         end
 
+        -- RFC 5321 4.5.2 transparency: a line starting with '.' is transmitted
+        -- with an extra leading '.', which the receiver strips. Without this a
+        -- lone '.' line in the body would end the DATA phase early and
+        -- truncate the message. Runs after CRLF normalization, so every line
+        -- is known to end in CRLF.
+        -- `at_line_start` carries line state across chunks of a split message
+        -- so a '.' at the head of a chunk is only stuffed when the previous
+        -- chunk actually ended a line.
+        local function dot_stuff(msg, at_line_start)
+          local mtype = type(msg)
+          local len
+
+          if mtype == 'userdata' then
+            len = msg:len()
+          elseif mtype == 'string' then
+            len = #msg
+          else
+            return msg, at_line_start
+          end
+
+          if len == 0 then
+            return msg, at_line_start
+          end
+
+          -- Matching on LF alone also covers CRLF, so stuffing stays correct
+          -- even if the content reaches us with mixed line endings
+          local leading_dot, embedded_dot
+          if mtype == 'userdata' then
+            -- len/at/find inspect the text in place; only stuffing copies it
+            leading_dot = at_line_start and msg:at(1) == 46
+            embedded_dot = msg:find('\n.') ~= nil
+          else
+            leading_dot = at_line_start and string.sub(msg, 1, 1) == '.'
+            embedded_dot = string.find(msg, '\n.', 1, true) ~= nil
+          end
+
+          local ends_line
+          if mtype == 'userdata' then
+            ends_line = msg:at(len) == 10
+          else
+            ends_line = string.sub(msg, -1) == '\n'
+          end
+
+          if not leading_dot and not embedded_dot then
+            return msg, ends_line
+          end
+
+          local str = mtype == 'userdata' and msg:str() or msg
+          str = string.gsub(str, '\n%.', '\n..')
+          if leading_dot then
+            str = '.' .. str
+          end
+
+          return str, ends_line
+        end
+
         if type(message) == 'string' or type(message) == 'userdata' then
           message = normalize_to_crlf(message)
+          message = dot_stuff(message, true)
           conn:add_write(pre_quit_cb, { message, CRLF .. '.' .. CRLF })
         else
           -- Array of chunks
+          local at_line_start = true
           for i = 1, #message do
             message[i] = normalize_to_crlf(message[i])
+            message[i], at_line_start = dot_stuff(message[i], at_line_start)
           end
           table.insert(message, CRLF .. '.' .. CRLF)
           conn:add_write(pre_quit_cb, message)
--- a/lualib/lua_smtp.lua
+++ b/lualib/lua_smtp.lua
@@ -240,23 +240,80 @@ local function sendmail(opts, message, callback)
         conn:add_read(from_done_cb, CRLF)
       end
     end
-    local function hello_done_cb(merr, mdata)
-      if no_error_read(merr, mdata) then
-        stage = 'from'
-        conn:add_write(from_cb, string.format(
-            'MAIL FROM: <%s>%s', opts.from, CRLF))
-      end
+    local supports_8bitmime = false
+
+    local function send_mail_from()
+      stage = 'from'
+      local mail_params = supports_8bitmime and ' BODY=8BITMIME' or ''
+      conn:add_write(from_cb, string.format(
+          'MAIL FROM: <%s>%s%s', opts.from, mail_params, CRLF))
     end
 
-    -- HELO stage
-    local function hello_cb(merr)
+    -- HELO stage (fallback used when EHLO is rejected or unsupported)
+    local function helo_done_cb(merr, mdata)
+      if no_error_read(merr, mdata) then
+        send_mail_from()
+      end
+    end
+    local function helo_cb(merr)
       if no_error_write(merr) then
-        conn:add_read(hello_done_cb, CRLF)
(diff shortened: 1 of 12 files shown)


src/plugins/lua/metadata_exporter.lua
```diff
@@ -68,8 +78,761 @@ Symbols: $symbols]],
...
+  local requested = {}
+  for name in string.gmatch(tmpl, '%$([%w_]+)') do
+    requested[name] = true
+  end
+  for name in string.gmatch(tmpl, '%${([%w_]+)}') do
+    requested[name] = true
+  end
+  for name in pairs(requested) do
+    if meta[name] == nil and variables[name] then
+      meta[name] = variables[name](task)
```
vstakhov, line 249: [P1] Sanitize custom values before template substitution. `expand_selectors` intentionally leaves placeholders already present in `meta` untouched, and `lua_util.template` then inserts this callback result verbatim. In a negative-control run, a custom variable returning `safe\r\nBcc: injected@example.com` produced a separate `Bcc:` header on the wire. Please route custom values through the same scalar/stringification and CR/LF policy as selector expansion (at least while rendering the header block), and add a regression test that proves a custom variable cannot create a new header.

src/plugins/lua/metadata_exporter.lua
```diff
@@ -68,8 +78,761 @@ Symbols: $symbols]],
...
+  ['resent-message-id'] = true,
+  ['return-path'] = true,
 }
 
+local function is_ascii(str)
+  return not string.find(str, '[\128-\255]')
+end
+
+local function encode_address_header(task, value)
+  if string.find(value, '[():]') then
```
vstakhov, line 347: [P1] Do not bypass encoding for every address header containing a comment/group marker. This returns the complete value unchanged even when its display name contains non-ASCII bytes; the new functional test consequently accepts raw `Cc: Jörg ... (Kommentar)`. The SMTP client negotiates 8BITMIME only, not SMTPUTF8, so this is not a valid way to transmit that header and contradicts `email_auto_encode_headers`. Please encode the supported portions without lossy reconstruction, or reject/fold a syntax you cannot safely encode, and make the test require no raw UTF-8 header bytes.

lualib/lua_smtp.lua
```diff
@@ -181,23 +240,80 @@ local function sendmail(opts, message, callback)
         conn:add_read(from_done_cb, CRLF)
       end
     end
-    local function hello_done_cb(merr, mdata)
+    local supports_8bitmime = false
+
+    local function send_mail_from()
+      stage = 'from'
+      local mail_params = supports_8bitmime and ' BODY=8BITMIME' or ''
```
vstakhov, line 247: [P1] Handle raw 8-bit content when 8BITMIME is unavailable. Omitting the `BODY=8BITMIME` parameter does not make the unchanged DATA payload 7-bit-safe: the `EMAIL_EXPORT_PARTS_DEFAULT` test sends `<p>Grüße</p>` with `Content-Transfer-Encoding: 8bit` to a peer advertising only `X8BITMIME`, then asserts merely that the BODY parameter is absent. Please arrange for such content to be transfer-encoded before DATA (or fail the send) when EHLO/HELO does not establish 8BITMIME support, and cover the actual transmitted octets in the fallback test.

src/plugins/lua/metadata_exporter.lua
```diff
@@ -68,8 +78,761 @@ Symbols: $symbols]],
...
+-- RFC 5322 998-octet limit break framing there. Quoted-printable encodes and
+-- folds both away. A bare "." line needs no handling here: dot-stuffing is the
+-- SMTP transport's job and lua_smtp does it (RFC 5321 4.5.2).
+local function needs_encoding_for_transport(content)
+  -- Plain find of a literal NUL: the %z class was removed in Lua 5.2 and
+  -- matches the letter 'z' there instead
+  if string.find(content, '\0', 1, true) then
+    return true
+  end
+  for line in (content .. '\n'):gmatch('(.-)\n') do
```
vstakhov, line 469: [P2] Avoid copying and rescanning the complete part body here. `(content .. '\n')` allocates another full-size string and the non-greedy capture creates per-line strings; for a large ASCII text part, auto selection calls this once in `guess_text_cte` and then again in `sanitize_cte`. That adds two full scans and two payload-sized temporary copies before MIME assembly/encoding. Please scan newline indexes/lengths without concatenating or capturing the body, and carry the computed safety result forward so auto mode evaluates it once.

vstakhov (changes requested):
I found four blocking issues in the new mail-template/MIME path: custom-variable header injection, invalid raw UTF-8 address headers, raw 8-bit DATA when the peer does not advertise 8BITMIME, and avoidable full-payload copies/repeated scans in transfer-encoding selection.

Validation at `c3a665c76211456121318e9327cdf43cc75b046b` (merge base `eaac8e914ef09f837fe53cc4973e0b2312ab1118`): ARM64 Debug build/install passed; the installed binary links `/opt/homebrew/opt/luajit/lib/libluajit-5.1.2.dylib`; all Lua tests passed (1255 tests, 1252 asserted passes, 3 unassertive); changed production Lua passed luacheck with 0 warnings/errors; changed Python passed `py_compile`; all 13 new focused functional cases passed (10 SMTP/MIME/schema cases plus the 3 Redis cases rerun with the Redis Python client available). A negative-control functional run confirmed that a custom variable containing CRLF creates a separate `Bcc:` header in the SMTP transcript. All 16 commits are GitHub-verified and their subjects have project tags.

GitHub currently reports the PR as conflicting with `master`, with no status checks and no earlier review discussion, so it also needs a rebase and a fresh CI run after these fixes.

PR #6221, opened 2026-09-01
[Feature] force_actions: composite-aware postfilter promotion and priority arbitrationFeature/force actions

c-rosenberg:
### Summary

Fixes `force_actions` rules that depend on composite symbols and adds
priority-based arbitration between multiple matching rules.

### Auto-promote composite-dependent rules to postfilter
`register_dependency()` from a `normal`-stage symbol on a composite silently
fails, since composites run in their own stage after all normal filters — so
a rule whose expression referenced a composite symbol never actually waited
for it. Composite atoms are now detected via
`rspamd_config:get_symbol_flags()`, and such rules are registered as
postfilters instead, same as the existing `require_action`/`honor_action`
handling. Applied to both the `rules` and legacy `actions` config paths.

### Priority arbitration
- Fix `has_composite_atom()`: `get_symbol_flags()` returns an array of flag
  names, not a keyed map, so it now iterates the array instead of testing a
  non-existent `.composite` field.
- `lua_config`: `get_symbol_flags()` was stripping the
  `SYMBOL_TYPE_COMPOSITE` bit before returning flags to Lua; restored so
  composite detection actually works.
- Add an optional `priority` rule setting (`low`/`normal`/`high`/`critical` or
  `0`-`3`) to arbitrate which of several matching rules' passthrough result
  wins, with validation, range clamping, and a warning when combined with
  `least`.
- Register rules in a stable, name-sorted order instead of relying on UCL's
  unspecified `pairs()` iteration order.
- Expose the configured priority as a `priority:N` symbol option alongside
  the existing action-name option, so it's visible in scan output next to the
  fired `FORCE_ACTION_*` symbol.

### Known limitation
A rule matching on a composite can still be starved if an earlier pre-result
was set without `process_all`, since that suppresses composite evaluation
entirely; set `process_all = true` on whichever rule fires first.

### Testing
Expanded `360_force_actions.robot` functional coverage: composite-triggered
postfilter promotion, priority ordering (named, numeric, mixed main/postfilter
stages), and priority validation warnings.

### DIFF
--- a/src/lua/lua_config.c
+++ b/src/lua/lua_config.c
@@ -2147,6 +2147,13 @@ lua_config_get_symbol_flags(lua_State *L)
 		flags = rspamd_symcache_get_symbol_flags(cfg->cache,
 												 name);
 
+		/* Item type bits are stripped off the stored flags when a cache item is
+		 * created, so re-add the composite one that callers rely upon */
+		if (rspamd_symcache_get_symbol_stage(cfg->cache, name) ==
+			SYMBOL_TYPE_COMPOSITE) {
+			flags |= SYMBOL_TYPE_COMPOSITE;
+		}
+
 		if (flags != 0) {
 			lua_push_symbol_flags(L, flags, LUA_SYMOPT_FLAG_CREATE_ARRAY);
 		}
--- a/src/plugins/lua/force_actions.lua
+++ b/src/plugins/lua/force_actions.lua
@@ -201,6 +201,11 @@ local function gen_cb(params)
       else
         task:set_pre_result { action = params.act, module = N, flags = flags, priority = params.priority }
       end
+
+      if params.priority then
+        return true, params.act, string.format('priority:%s', params.priority)
+      end
+
       return true, params.act
     end
 
--- a/src/plugins/lua/force_actions.lua
+++ b/src/plugins/lua/force_actions.lua
@@ -35,18 +35,89 @@ local lua_selectors = require "lua_selectors"
 -- 'normal' symbol cannot depend on them via register_dependency (the cache
 -- rejects such cross-stage deps). Any rule referencing a composite atom must
 -- therefore be registered as a postfilter instead.
+-- Such a rule is still starved if any earlier pre-result was set without
+-- `process_all`, since that suppresses composite evaluation entirely.
 local function has_composite_atom(atoms)
   for _, a in ipairs(atoms) do
-    local fl = rspamd_config:get_symbol_flags(a)
-    if fl and fl.composite then
-      return true
+    -- get_symbol_flags returns an array of flag names, not a keyed map
+    for _, fl in ipairs(rspamd_config:get_symbol_flags(a) or {}) do
+      if fl == 'composite' then
+        return true
+      end
     end
   end
   return false
 end
 
+-- Passthrough priority band from src/libmime/scan_result.h:
+-- 0 = low, 1 = normal (default), 2 = high, 3 = critical (used by the core for
+-- broken messages). The C side stores it as unsigned int, so a negative value
+-- would wrap around and outrank everything.
+local min_priority, max_priority = 0, 3
+local named_priorities = {
+  low = 0,
+  normal = 1,
+  high = 2,
+  critical = 3,
+}
+
+local function check_priority(name, priority)
+  if priority == nil then
+    return nil
+  end
+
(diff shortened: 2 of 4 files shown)


src/lua/lua_config.c
```diff
@@ -2147,6 +2147,13 @@ lua_config_get_symbol_flags(lua_State *L)
 		flags = rspamd_symcache_get_symbol_flags(cfg->cache,
 												 name);
 
+		/* Item type bits are stripped off the stored flags when a cache item is
+		 * created, so re-add the composite one that callers rely upon */
+		if (rspamd_symcache_get_symbol_stage(cfg->cache, name) ==
+			SYMBOL_TYPE_COMPOSITE) {
+			flags |= SYMBOL_TYPE_COMPOSITE;
+		}
```
vstakhov, line 2155: I think it should be done properly by preserving that flag on a symbol.

src/plugins/lua/force_actions.lua
```diff
@@ -111,10 +196,16 @@ local function gen_cb(params)
       if type(params.message) == 'string' then
         -- process selector expressions in the message
         local message = string.gsub(params.message, '(${(.-)})', process_message_selectors)
-        task:set_pre_result { action = params.act, message = message, module = N, flags = flags }
+        task:set_pre_result { action = params.act, message = message, module = N, flags = flags,
+                              priority = params.priority }
       else
-        task:set_pre_result { action = params.act, module = N, flags = flags }
+        task:set_pre_result { action = params.act, module = N, flags = flags, priority = params.priority }
```
vstakhov, line 202: [P1] Do not skip a configured higher-priority result merely because its action is currently selected. This old shortcut is no longer semantics-preserving now that rules arbitrate by priority: with three name-sorted, matching `process_all` rules (`low reject`, `critical reject`, then `normal add header`), the critical rule returns here without adding its passthrough result; the final action becomes `add header`, and `FORCE_ACTION_B_SAME_ACTION_CRITICAL` is absent. I reproduced that result against this head. Please disable this shortcut for prioritized rules (or compare/update the selected passthrough priority) and add a regression containing two same-action priorities followed by a competing action.

test/functional/cases/360_force_actions.robot
```diff
@@ -42,3 +42,56 @@ FORCE ACTIONS from add header to reject
...
+  Expect Symbol  FORCE_ACTION_FORCE_MIX_POST_WINS_ADD_HEADER
+  Expect Symbol  FORCE_ACTION_FORCE_MIX_POST_WINS_REJECT
+
+FORCE ACTIONS priority validation
+  ${log} =  Get File  ${RSPAMD_TMPDIR}/rspamd.log  encoding_errors=ignore
+  Should Contain  ${log}  priority -10 is out of the allowed range [0..3], adjusted to 0
+  Should Contain  ${log}  priority 10 is out of the allowed range [0..3], adjusted to 3
+  Should Contain  ${log}  unknown priority name "invalid"
+  Should Contain  ${log}  priority 1.5 is not an integer, adjusted to 1
+  Should Contain  ${log}  priority is NaN; ignored
```
vstakhov, line 92: [P2] Make the NaN validation portable to the supported standard-Lua runtimes. The current Fedora job runs with `ENABLE_LUAJIT=OFF` and fails here because converting the string `nan` with `tonumber` does not yield NaN there, so the configuration is ignored as an unknown priority name and this message is never logged. LuaJIT passing locally does not cover the required Lua 5.3/5.4 behavior. Please feed an actual NaN through a suitable unit-level test or adjust/remove this runtime-specific assertion.

vstakhov (changes requested):
Requesting changes for the priority-arbitration bug below and the standard-Lua CI failure. This is in addition to the existing review thread about reconstructing the composite type bit in the Lua getter instead of preserving/fixing the symbol representation at its source.

Performance audit: I found no per-message conversion of message bodies or other large data. The new rule-name sort and symbol-flag scans run once at configuration load; the only new matching-path allocation is the small priority symbol option.

Validation on this exact head: native ARM64 build using Homebrew LuaJIT; ctest 8/8; luacheck clean; clang-format check clean; focused 360_force_actions Robot suite 13/13. A focused negative control with low reject -> critical reject -> normal add header failed with add header and omitted the critical rule symbol. All three commits are GitHub-verified. Fedora's non-LuaJIT job is currently red on the NaN assertion.

PR #6222, opened 2026-09-01
[Feature] lua_magic: expand format detection coverage (archives, media, documents) and fix content-type inconsistencies

c-rosenberg:
### Summary

Significantly widens `lua_magic`'s content-based file type detection and
fixes a handful of pre-existing content-type mismatches, plus keeps
`lua_mime_types`'s cross-check whitelist in sync.

### New format detection
- MHTML/MHT archives, detected from content (`MIME-Version` +
  `multipart/related`) or the `.mht`/`.mhtml` extension, not just the
  declared `Content-Type`.
- VHD/VHDX, OneNote, OLE-encrypted packages, URL shortcuts, PGP keys/
  signatures (now split into separate `asc`/`sig` patterns, matching
  `file(1)`'s distinct `application/pgp-keys` vs `application/pgp-signature`
  mime types), XML/XSLT, LHA/LZ archives, WMF/EMF images, and a range of
  audio/video containers (FLAC, MP3, OGG/OGV, AVI, WAV, AIFF, FLV, ASF/WMA,
  MKV/MKA/WEBM/WEBA, MP4/M4A/MOV/3GP/3G2).

### Content-type fixes (⚠️ potentially breaking)
`lnk`, `bz2` and `iso` already had `types.lua` entries with incorrect `ct`
values: `lnk` copy-pasted the `exe` entry's `x-ms-application` instead of
`x-ms-shortcut`, and `bz2`/`iso` used non-standard `x-bzip`/`x-iso` instead
of the libmagic-standard `x-bzip2`/`x-iso9660-image`.

**This changes the `detected_ext`/content-type string these three
extensions report.** Anything keying off the old values (e.g. `mime_types`
plugin extension_map lookups, score maps by content-type) should be checked
before relying on this.

### mime_types plugin sync
- Register the newly-ported `lua_magic` extensions (`asc`, `bz2`, `emf`,
  `iso`, `lha`, `lnk`, `lz`, `mka`, `mkv`, `oleenc`, `sig`, `url`, `vhd`,
  `vhdx`, `weba`) in `lua_mime_types.full_extensions_map` so the
  `MIME_BAD_ATTACHMENT` ct cross-check covers them too, matching
  `lua_magic/types.lua` exactly.
- Fix three existing entries (`aiff`, `pub`, `xslt`) whose accepted value no
  longer matched `lua_magic`'s (now libmagic-verified) detection, extending
  each to accept both values, following the existing wav/mp3/p7s/m4a
  multi-value pattern.

### Testing
Added `test/lua/unit/lua_magic.mht.lua` unit test coverage for MHTML/MHT
detection.


lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+-- "webm") tells webm from generic mkv, and a scan for known Tracks CodecID
+-- strings tells audio-only mka/weba from video. This is a cheap
+-- approximation of a real EBML track parser: CodecIDs live in the Tracks
+-- element, which for well-formed files appears well within the first 64KiB
+exports.mkv_format_heuristic = function(input, log_obj, pos, part)
+  if not input then
+    return 'mkv', 60
+  end
+
+  local head = tostring(input:span(1, math.min(#input, 65536)))
```
vstakhov, line 752: Why `tostring` here?

lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+end
+
+-- Matroska/WebM heuristic: the EBML DocType element (0x42 0x82 + size +
+-- "webm") tells webm from generic mkv, and a scan for known Tracks CodecID
+-- strings tells audio-only mka/weba from video. This is a cheap
+-- approximation of a real EBML track parser: CodecIDs live in the Tracks
+-- element, which for well-formed files appears well within the first 64KiB
+exports.mkv_format_heuristic = function(input, log_obj, pos, part)
+  if not input then
+    return 'mkv', 60
```
vstakhov, line 749: If no input, why do we return `mkv`?

lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+-- kept in sync, so unrecognised-but-valid brands fall back to mp4 instead
+-- of going undetected entirely
+local ftyp_m4a_brands = {
+  msnv = true, ndas = true,
+  ['f4a '] = true, ['f4b '] = true,
+  ['m4a '] = true, ['m4b '] = true, ['m4p '] = true,
+}
+exports.ftyp_format_heuristic = function(input, log_obj, pos, part)
+  if not input then
+    return 'mp4', 60
```
vstakhov, line 781: Same question

lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+  end
+
+  local brand = tostring(input:span(pos + 1, 4)):lower()
```
vstakhov, line 784: 4 symbols string is ok, but still not sure why do we need it.

lualib/lua_magic/heuristics.lua
```diff
@@ -63,10 +69,10 @@ local zip_patterns = {
 local txt_trie
 local txt_patterns = {
   html = {
-    { [=[(?i)<html[\s>]]=],                   32 },
+    { [=[(?i)<html[\s>]]=],                   40 },
```
vstakhov, line 72: [P1] Do not make a single incidental HTML token sufficient to reclassify plain text. `text_part_heuristic` accepts a result at weight >= 40, so raising `<html...>` and `<!DOCTYPE HTML...>` from 32/33 to 40 changes the semantics from additive evidence to one-token detection. A `text/plain` part containing `Documentation: the root element is <html> and nothing here is HTML markup.` now returns `html, 40`; the MIME pipeline then parses that part as HTML, affecting text and URL extraction. Please retain the previous sub-threshold weights, or require an anchored/document-level structure before crossing the threshold.

lualib/lua_magic/types.lua
```diff
@@ -68,6 +68,10 @@ local types = {
     ct = 'application/x-chm',
     type = 'binary',
   },
+  mht = {
+    ct = 'application/x-mimearchive',
```
vstakhov, line 72: [P1] Keep the detected MHT type consistent with the filename content-type whitelist. This declares `application/x-mimearchive`, but `lua_mime_types.full_extensions_map` still maps both `mht` and `mhtml` exclusively to `message/rfc822`. The `mime_types` plugin compares an attachment's declared type against that map, so a legitimate `.mht` declared with the exact type introduced here is reported as `MIME_BAD_ATTACHMENT`. Please allow both representations in the extension map (or choose one consistently) and add an integration-level regression test.

lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+  if not input then
+    return
+  end
+
+  local b2 = input:byte(2)
+  if not b2 then
+    return
+  end
+
+  if bit.band(b2, 0xF6) == 0xF2 then
```
vstakhov, line 679: [P2] Validate an MPEG frame header rather than two bytes. With this check, a two-byte attachment containing only `FF F2` is immediately classified as `mp3` with weight 40. At minimum, require the complete four-byte header and reject reserved/free bitrate and reserved sample-rate values; preferably validate the computed frame length or a following frame before assigning decisive weight. This needs short/truncated and random-binary negative tests.

lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+
+local mkv_video_codecs = {
+  'V_MPEG', 'V_THEORA', 'V_VP8', 'V_VP9', 'V_AV1',
+  'V_UNCOMPRESSED', 'V_QUICKTIME', 'V_MS/VFW',
+}
+local mkv_audio_codecs = {
+  'A_VORBIS', 'A_OPUS', 'A_AC3', 'A_EAC3', 'A_AAC', 'A_DTS',
+  'A_FLAC', 'A_MPEG', 'A_PCM', 'A_TTA1', 'A_WAVPACK4', 'A_MS/ACM',
+}
+local function mkv_find_any(head, markers)
```
vstakhov, line 733: [P2] Avoid copying and repeatedly scanning the 64 KiB prefix. `rspamd_text:find` already performs a plain search directly on the span, so `tostring(input:span(...))` is unnecessary here; after the copy, the two marker lists can rescan the same 64 KiB up to 20 times. Compile the codec markers into one plain `rspamd_trie`/multipattern with an index-to-kind table and scan the `rspamd_text` span once. The Ogg and ASF helpers can likewise call `rspamd_text:find` directly and avoid their 64-byte/8-KiB copies.

lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+-- ISO-BMFF ftyp brand dispatch table/heuristic: new major brands (iso6,
+-- dash, ...) show up far faster than a hand-maintained regex list can be
+-- kept in sync, so unrecognised-but-valid brands fall back to mp4 instead
+-- of going undetected entirely
+local ftyp_m4a_brands = {
+  msnv = true, ndas = true,
+  ['f4a '] = true, ['f4b '] = true,
+  ['m4a '] = true, ['m4b '] = true, ['m4p '] = true,
+}
+exports.ftyp_format_heuristic = function(input, log_obj, pos, part)
```
vstakhov, line 779: [P1] Check the available bytes and the box header before slicing the major brand. The candidate regexp matches at byte 8, but this function unconditionally requests bytes 9-12. An attachment containing exactly `xxxxftyp` therefore raises `invalid arguments: start offset 9 is larger than text len 8` from `rspamd_text:span`; 9-11 byte inputs fail on the requested length as well. Please reject truncated input and validate the first box size (including the extended-size case) before reading the brand.

lualib/lua_magic/heuristics.lua
```diff
@@ -649,4 +662,152 @@ exports.svg_format_heuristic = function(input, log_obj, pos, part)
...
+    return 'mov', 60
+  end
+
+  if ftyp_m4a_brands[brand] then
+    return 'm4a', 60
+  end
+
+  -- isom, iso2..isoN, avc1, mp41/42, mmp4, dash, m4v , f4v , nds[chmps],
+  -- and any brand not specifically catalogued above
+  return 'mp4', 60
```
vstakhov, line 810: [P1] An unknown ISO-BMFF brand is not necessarily MP4 video. This fallback classifies image and unrelated container brands as `video/mp4`; for example, a structurally sized `ftypavif` input returns `mp4`, and the incomplete HEIF guard also labels the registered `heix` image brand as MP4. The MP4 Registration Authority lists many non-video brands, including `avif`, `avis`, `heic`, `heix`, `heim`, `heis`, `hevc`, `hevx`, `hevm`, and `hevs`: https://mp4ra.org/registered-types/brands. Please dispatch only known brands (including compatible brands if desired) and return no result for unknown families rather than guessing video.

lualib/lua_mime_types.lua
```diff
@@ -63,6 +63,7 @@ exports.full_extensions_map = {
   { "art", "image/x-jg" },
   { "asa", "application/xml" },
   { "asax", "application/xml" },
+  { "asc", "application/pgp-keys" },
```
vstakhov, line 66: [P2] Do not restrict the generic `.asc` armor extension to public keys. GnuPG uses `.asc` for ASCII-armored signatures and other armored output as well, while IANA defines distinct `application/pgp-keys`, `application/pgp-signature`, and `application/pgp-encrypted` types. With this single-value map, legitimate `.asc` signatures or encrypted payloads are reported as `MIME_BAD_ATTACHMENT`. Please accept the relevant OpenPGP media types here, or leave `.asc` unmapped and rely on the content-specific detector. References: https://www.gnupg.org/documentation/manuals/gnupg/gpg.html and https://www.iana.org/assignments/media-types/media-types.xhtml

vstakhov (changes requested):
Reviewed the current head (d215f20a8), including the existing review threads. Luacheck is clean, all 23 focused MHT tests pass, and the full local Lua suite passes (1250 tests; 1247 asserted passes, 3 pre-existing unassertive cases) under the Homebrew ARM64 LuaJIT runtime. However, targeted negative controls expose runtime errors and false classifications in the new media paths, and the type-map inconsistencies can create MIME_BAD_ATTACHMENT false positives. Please also add table-driven positive, truncated-input, and ambiguous-container tests for the newly added formats; the current new tests cover only MHT. The failing latest Playwright job is an unrelated scan-counter mismatch (expected +2, observed +4).

PR #6223, opened 2026-09-01
[Feature] crypto: multi-currency wallet address detection replacing bitcoin.lua

c-rosenberg:
### Summary

Replaces the old Bitcoin-only regexp rule with a proper multi-currency
crypto wallet address detection library, a new rule set, and a selector
extractor built on top of it.

### lua_crypto_addresses library
New `lualib/lua_crypto_addresses.lua` validates and classifies
cryptocurrency wallet addresses found in a message: Bitcoin and forks
(Base58Check, Bech32, Bech32m, CashAddr), Litecoin, Dogecoin, Tron, XRP,
Zcash transparent, Cardano, Cosmos, Stellar, TON (checksum verified), plus
format-only Ethereum/EVM and Monero matching.

`get_addresses(task)` scans the subject and text parts once per task and
memoises the result via the task cache, so both the symbol callbacks and the
`crypto_addresses` selector share a single scan.

### crypto.lua rules (replaces bitcoin.lua)
Registers `CRYPTO_ADDR_CHECK` plus one virtual symbol per currency on top of
`lua_crypto_addresses`, replacing the old `BITCOIN_ADDR`-only regexp rule.
`LEAKED_PASSWORD_SCAM_RE`/`LEAKED_PASSWORD_SCAM` now trigger on any supported
currency instead of Bitcoin alone.

### crypto_addresses() selector
Returns wallet addresses found in a message, optionally filtered to one
currency and/or typed as `"currency:address"`. Computes addresses via
`lua_crypto_addresses` directly, so it needs no dependency on
`CRYPTO_ADDR_CHECK`.

### Testing
- Unit tests for `lua_crypto_addresses` covering the supported address
  formats.
- Functional coverage for every supported currency symbol,
  `CRYPTO_ADDR_CHECK` combined options, the `crypto_addresses()`
  selector/map integration, uppercase Bech32/CashAddr, Taproot, the
  no-bare-Base58-false-positive case, and the non-Bitcoin leaked-password-scam
  composite.


lualib/lua_crypto_addresses.lua
```diff
@@ -0,0 +1,710 @@
...
+
+    -- v0 is only ever P2WPKH (20 bytes) or P2WSH (32 bytes)
+    if witver == 0 and plen ~= 20 and plen ~= 32 then
+      return nil
+    end
+
+    return known.currency
+  end
+
+  if chk == BECH32_CONST then
```
vstakhov, line 319: [P1] Validate the decoded payload, not merely the Bech32 checksum. This branch accepts any checksum-valid payload for the Cardano and Cosmos HRPs. For example, `cosmos1qqqqk0wkhp` and `addr1qqqqxn5hht` both decode to only two payload bytes, yet `classify` returns `cosmos` and `cardano`. Since these symbols participate in the 7-point `LEAKED_PASSWORD_SCAM` composite, an arbitrary checksum-valid string can become a strong false-positive signal. Please enforce the chain-specific payload format and length before returning (Cardano address header/network/type per CIP-19; Cosmos account/address byte length and type). References: https://cips.cardano.org/cip/CIP-19 and https://docs.cosmos.network/sdk/latest/guides/reference/bech32

lualib/lua_crypto_addresses.lua
```diff
@@ -0,0 +1,710 @@
...
+local wallet_re = rspamd_regexp.create([[/\b(?:]] ..
+    [[0x[0-9a-fA-F]{40}|]] ..                                  -- EVM
+    [[G[A-Z2-7]{55}|]] ..                                      -- Stellar
+    [[[EUk0]Q[A-Za-z0-9_-]{46}|]] ..                           -- TON
+    [[[A-Za-z]{2,12}1[qpzry9x8gf2tvdw0s3jn54khce6mua7lQPZRY9X8GF2TVDW0S3JN54KHCE6MUA7L]{10,100}|]] .. -- bech32/bech32m
+    [[[A-Za-z]+:[qpzry9x8gf2tvdw0s3jn54khce6mua7lQPZRY9X8GF2TVDW0S3JN54KHCE6MUA7L]{14,}|]] .. -- cashaddr with prefix
+    [[[qpzry9x8gf2tvdw0s3jn54khce6mua7lQPZRY9X8GF2TVDW0S3JN54KHCE6MUA7L]{42}|]] .. -- cashaddr without prefix
+    [[4[1-9AB][1-9A-HJ-NP-Za-km-z]{93}|]] ..                   -- Monero
+    [[[13LM9ADTrt][1-9A-HJ-NP-Za-km-z]{24,34}]] ..             -- Base58Check family
+    [[)\b/]])
```
vstakhov, line 607: [P1] The shared trailing `\b` drops valid TON addresses ending in `-`. Dash is allowed by TON's URL-safe Base64 encoding, but both the final dash and normal following whitespace are non-word characters, so there is no word boundary. `EQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIOs-` has a valid tag, workchain, 32-byte account and CRC16; `classify` returns `ton`, while `get_addresses` extracts nothing. Please use family-specific boundaries rather than one boundary around the full alternation. Reference: https://docs.ton.org/foundations/addresses/formats

lualib/lua_crypto_addresses.lua
```diff
@@ -0,0 +1,710 @@
...
+local btc_lut = make_b58_lut("123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz")
+-- XRP Base58: same character set, different ordering
+local xrp_lut = make_b58_lut("rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz")
+
+--[[[
+-- Decodes a Base58Check string of `total_len` bytes and verifies the trailing
+-- 4 byte double SHA256 checksum. Returns the decoded bytes (a table of
+-- integers, checksum included) or nil when the input is not valid.
+--]]
+local function base58check_decode(word, lut, total_len)
```
vstakhov, line 69: [P2] Preserve and validate Base58 leading-zero semantics. This fixed-width accumulator ignores extra leading zero symbols, so prepending `1` to the valid Bitcoin address produces `116L5yRNPTuciSgXGHqYwn9N6NeoKqopAu`, which still decodes to the same bytes and passes the checksum here even though it is not the canonical address. The same class of issue applies to each alphabet's zero symbol. Please account for leading zeros and reject overflow/non-canonical encodings, for example by requiring an exact decoded length and canonical re-encoding. Bitcoin Core's decoder explicitly counts leading zeroes: https://github.com/bitcoin/bitcoin/blob/master/src/base58.cpp

lualib/lua_crypto_addresses.lua
```diff
@@ -0,0 +1,710 @@
...
+  'cosmos',
+  'stellar',
+  'ton',
+  'ethereum',
+  'monero',
+}
+
+-- Extraction ----------------------------------------------------------------
+
+-- A single candidate regexp keeps the scan to one pass per text part; the
```
vstakhov, line 596: [P2] Why use one regexp rather than `rspamd_trie`/multipattern here? With regexp patterns plus start-of-match offsets, multipattern can report the matched pattern index, dispatch directly to the corresponding validator, and let each family define its own boundaries; with Hyperscan/VectorScan it remains a single scan. The regex fallback can scan the patterns separately when Hyperscan is unavailable, so that case should be benchmarked rather than assumed, but the current union both loses the family identity and has already coupled the alternatives incorrectly via the shared boundary. Please switch to multipattern, or document and benchmark why the combined PCRE is preferable while still separating the family-specific boundaries.

lualib/lua_crypto_addresses.lua
```diff
@@ -0,0 +1,710 @@
...
+
+  if cached then
+    return cached
+  end
+
+  local found = {}
+  local seen = {}
+
+  for _, hay in ipairs(get_haystacks(task)) do
+    for _, raw_word in ipairs(wallet_re:search(hay) or E) do
```
vstakhov, line 647: [P1] Bound and stream candidate processing. Passing `rspamd_text` into `search` is zero-copy, which is good, but `rspamd_regexp:search` then creates a Lua table containing every match and copies each match to a Lua string before this loop; the Lua search loop has no hit cap. A crafted body containing many distinct address-shaped strings therefore grows the match table plus `seen`/`found`, and runs Base58 decoding and double-SHA256 for every candidate. This normal symbol runs on every message. A multipattern callback can validate as matches arrive and stop after a configured candidate/address budget; keep the haystack as `rspamd_text`, use a zero-copy `span` for the reported offsets, and stringify only that bounded address-sized slice. Also, `raw_word` is already a Lua string here, so the following `tostring` is redundant, though it is not the large-copy problem.

vstakhov (changes requested):
Reviewed the current head (bfd007128). The focused Lua and functional tests pass, but the issues below can produce false positives or false negatives in a high-score scam composite, and the extraction path has an unbounded allocation/validation cost on attacker-controlled bodies. On the regexp-versus-multipattern choice: multipattern is a better fit if it is used with match callbacks and start offsets, while retaining rspamd_text inputs and materialising only bounded candidate slices. Please address these before merge.

lualib/lua_crypto_addresses.lua
vstakhov: I don't think it's important, but IIRC we had some bounding parameters there.

PR #6224, opened 2026-09-01
[Feature] auto-register symbol() dependencies for selector-based rules (rbl, reputation, clustering, multimap, dmarc, ratelimit)

c-rosenberg:
### Summary

Rules built from a `lua_selectors` expression that references
`symbol('OTHER')` previously ran with no guaranteed ordering relative to
`OTHER` unless a manual `register_dependency()` call was added by hand — an
easy thing to forget, and easy to silently get wrong. This PR makes that
dependency automatic wherever such rules are built from a selector.

### symcache: apply post-init dependencies instead of dropping them
`add_delayed_dependency()` previously only queued edges before `init()` ran
and silently discarded any dependency registered afterwards (e.g. from a Lua
`add_on_load` callback, which is how the reputation/RBL selector wiring below
registers deps). It now resolves and applies such edges immediately against
the live graph, guarded by a new `scans_started` flag that rejects
registration once tasks have started scanning against a shared graph.
Also hardens `add_dependency()`/`get_item_by_id(_mut)()` to probe
`items_by_id` via `find_map()` instead of assuming dense/contiguous ids,
which no longer holds once `init()` prunes statically disabled symbols.

### lua_selectors: auto-register symbol() dependencies
`create_selector_closure()` / `create_selector_closure_with_combinator()`
now accept an optional `own_symbol` argument. When given, a dependency is
automatically registered from `own_symbol` on every symbol referenced via a
`symbol()` extractor in the selector. The raw dependency list is also
exposed via new `get_selector_symbol_deps()`/`register_symbol_deps()`
helpers, for callers building several selectors per rule that need to merge
dependencies before registering once.

### Wired into consumers
- **rbl, reputation, clustering, multimap**: pass `own_symbol` through to
  `lua_selectors` so rules built from a `symbol("OTHER"):...` selector
  automatically depend on `OTHER`. `clustering.lua` merges source/cluster
  selector deps into a single `register_symbol_deps()` call.
- **lua_maps_expressions**: accumulate selector dependencies from every rule
  while the object is built, and register them against `own_symbol` only
  once the whole object (all rules + combining expression) has built
  successfully, avoiding a dangling dependency if `create()` bails out.
  Wired through `dmarc` (`munge_map_condition`), `multimap` (combined
  rules), and `reputation` (whitelist).
- **ratelimit**: accumulate `symbol()` dependencies from every rate's
  selector while parsing, and register them against `RATELIMIT_CHECK` once
  known.

### Testing
- Unit tests for `lua_selectors`' dependency auto-registration and
  `lua_maps_expressions`' deferred registration.
- Functional coverage: RBL rules sharing one `symbol()` selector (exercising
  the selector-cache-hit path), a combined multimap rule and a DMARC
  `munge_map_condition` referencing `symbol('SELECTOR_RBL_DEP')`, and a
  ratelimit rule keyed on a deliberately low-priority symbol proving the
  bucket key is only stable once the automatic dependency is in place.

### DIFF
--- a/lualib/lua_maps_expressions.lua
+++ b/lualib/lua_maps_expressions.lua
@@ -102,7 +102,7 @@ exports.schema = T.table({
 })
 
 --[[[
--- @function lua_maps_expression.create(config, object, module_name)
+-- @function lua_maps_expression.create(cfg, obj, module_name, own_symbol)
 -- Creates a new maps combination from `object` for `module_name`.
 -- The input should be table with the following fields:
 --
@@ -120,9 +120,11 @@ exports.schema = T.table({
 -- In case if `expression` is false a `nil` value is returned.
 -- @param {rspamd_config} cfg rspamd config
 -- @param {table} obj configuration table
+-- @param {string} module_name optional module name for logging, defaults to `lua_maps_expressions`
+-- @param {string} own_symbol optional symbol name; registers a dependency on any symbol() referenced by a rule's selector
 --
 --]]
-local function create(cfg, obj, module_name)
+local function create(cfg, obj, module_name, own_symbol)
   if not module_name then
     module_name = 'lua_maps_expressions'
   end
@@ -139,8 +141,15 @@ local function create(cfg, obj, module_name)
     module_name = module_name
   }
 
+  -- Registered only after a successful build, so a bailout below leaves no dangling dependency
+  local selector_symbol_deps = {}
+
   for name, rule in pairs(obj.rules) do
-    local sel = lua_selectors.create_selector_closure(cfg, rule.selector)
+    local sel, sel_deps = lua_selectors.create_selector_closure(cfg, rule.selector)
+
+    for _, dep in ipairs(sel_deps or {}) do
+      selector_symbol_deps[#selector_symbol_deps + 1] = dep
+    end
 
     if not sel then
       rspamd_logger.errx(cfg, 'cannot add selector for element %s in module %s',
@@ -215,6 +224,8 @@ local function create(cfg, obj, module_name)
   end
 
   ret.symbol = obj.symbol
+  ret.selector_symbol_deps = selector_symbol_deps
+  lua_selectors.register_symbol_deps(cfg, own_symbol, selector_symbol_deps)
 
   return ret
 end
--- a/lualib/lua_selectors/init.lua
+++ b/lualib/lua_selectors/init.lua
@@ -634,15 +634,70 @@ end
 
 
 --[[[
--- @function lua_selectors.create_closure(log_obj, cfg, selector_str, delimiter, fn)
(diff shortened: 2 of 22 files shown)


vstakhov (changes requested):
Thanks, this is a well-structured change and the symcache part looks correct to me. One design problem needs a fix in this PR, plus a few minor points.

### Prefilter rules were pulling filter symbols into the prefilter stage

`RATELIMIT_CHECK` is a prefilter by default (`prefilter = true` in `ratelimit.lua`), and the new test config does not override that. With this PR a ratelimit selector such as `symbol('X')` makes the prefilter depend on `X`. In the prefilter stage `check_item_deps()` eagerly *executes* not-started dependencies, so `X` and everything `X` depends on were run inside the prefilter stage, and the stage stalled until they finished. That is what made the new `RATELIMIT CHECK SELECTOR SYMBOL DEPENDENCY` test pass: `SELECTOR_RBL_DEP` was hoisted into the prefilter stage, not ordered before `RATELIMIT_CHECK` in the filter stage.

A prefilter must never depend on a filter, so the symcache now rejects such edges at init and drops them (master cc9038fc9). `rspamadm configtest` reports:

```
cannot add dependency from RATELIMIT_CHECK (prefilter) on SELECTOR_RBL_DEP (filter): a prefilter cannot depend on a symbol from a later stage
```

After a rebase onto master the ratelimit test will fail, since the dependency is no longer applied. Please:

1. Set `prefilter = false` in `test/functional/configs/ratelimit.conf` so the test proves the ordering guarantee it claims.
2. In `ratelimit.lua`, when `settings.prefilter` is true and the collected selector deps are not empty, log an error saying that `symbol()` selectors need `prefilter = false`, instead of registering the dependency. The same guard makes sense for multimap rules with `prefilter = true`.

### Minor

- `add_delayed_dependency()` post-init path skips the `disabled_ids` filter. A dependency whose source was removed by `rspamd_config:disable_symbol()` is skipped quietly at init, but after init the item is gone from both maps, so the same case logs `cannot register delayed dependency ...: source X is missing` as an error at every worker start (reputation registers from `add_on_load`). Keeping the statically disabled names in a member set after init would let you skip these quietly.
- The immediate path reuses the "cannot register delayed dependency" wording; for a dependency applied right away the word "delayed" is misleading.
- `scans_started` is set inside `maybe_resort()`. It works because the runtime constructor is the only caller, but a graph-freeze side effect in a function named "maybe resort" is easy to lose in a later refactor; setting it from `symcache_runtime::create()` would make the intent visible.
- Reputation is the motivating post-init consumer, but the only post-init functional coverage is the synthetic `POSTINIT_DEP` in `deps.lua`. A generic reputation rule with a `symbol()` selector in the functional configs would pin the real path end to end.

PR #6226, opened 2026-09-02
[Fix] lua_maps: include the map type in the cache key of url lists

dragoangel:
map_add_from_ucl() caches maps so that modules pointed at the same feed share one object. A map given as a single string is keyed by its url and its type, but a map given as a list of urls is keyed by the digest of the list alone, so the type requested by the first caller wins and every later caller silently gets a map of a foreign type.

url_redirector asks for a glob map of redirector_hosts_map while the multimap `redirector` rule asks for a hash map of the same two urls. multimap registers first, so url_redirector runs on a hash map and no glob pattern in the feed can ever match.

The collision is visible at startup: the module that did not get its own map logs the one it was handed instead.

```
lua_maps.lua:396: reuse url for complex map definition f4fbdf5a: Redirectors definitions (glob: bare names match exactly, *.foo matches subs)
```

Checked on a running instance with both modules enabled over one shared list: a `*.sub.example` entry never matched while an exact host did, and the same message with the patch applied resolves through the glob, with the reuse line gone.

### DIFF
--- a/lualib/lua_maps.lua
+++ b/lualib/lua_maps.lua
@@ -391,7 +391,8 @@ local function rspamd_map_add_from_ucl(opt, mtype, description, callback)
       return ret
     end
   elseif type(opt) == 'table' then
-    local cache_key = lua_util.unordered_table_digest(opt)
+    -- Type is a part of the key: the same urls can back maps of different types
+    local cache_key = map_hash_key(lua_util.unordered_table_digest(opt), mtype)
     if not callback and maps_cache[cache_key] then
       rspamd_logger.infox(rspamd_config, 'reuse url for complex map definition %s: %s',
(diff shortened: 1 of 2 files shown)


vstakhov:
Thanks, the fix is right. I pushed two commits on top of your branch:

- `[Minor] lua_maps: key url-list maps by their effective type` moves the `glob;`/`hash;`/… prefix stripping ahead of the cache lookup, so the key carries the effective type rather than the one the caller asked for (with your change alone, `{ "glob;/x" }` requested as `hash` and as `set` produced two identical maps instead of one).
- `[Test] lua_maps: cover map cache keys across map types` adds a unit test: the first case fails on master with exactly the collision you hit, the third one fails without the prefix hoist.

Full Lua unit suite passes locally with both.

PR #6229, opened 2026-09-03
[Feature] WebUI: informative All SERVERS cluster row

moisseev:
Derive the aggregate row from cluster data instead of a mix of the first neighbour's values, averages and blanks: majority version, configuration and git ids (with drift badges), youngest uptime, cluster-wide scan-time envelope, slowest-server latency, an up/total health counter, and an expandable details row with summed traffic and memory totals. Fixes the literal "undefined" version shown when the first neighbour is down or legacy (<= 2.5).


moisseev:
@review deep

PR #6239, opened 2026-09-09
[Feature] WebUI: Bayes learns balance bar and min_learns status in the Status tab

moisseev:
## Motivation

The Bayesian statistics table on the Status tab showed raw learn counts but did
not answer two questions an admin actually asks:

- how balanced the training between the classes is (spam vs ham);
- whether the classifier classifies at all: below the `min_learns` threshold a
  class stops participating, and a binary classifier stops classifying
  entirely — with no indication in the UI.

## Changes

1. **WebUI: Bayes learns balance bar** (frontend only, works against an
   unmodified server)
   - Stacked balance bar of the per-class learns share in the classifier cell,
     computed client-side from statfile revisions (the redis backend refreshes
     them asynchronously, ~30 s, so the bar may lag slightly).
   - Segment colors follow the spam/ham row semantics via dedicated
     `--rspamd-balance-*` variables, validated for contrast and CVD separation
     in both light and dark themes; identity is never color-alone — tooltips
     and the neighbouring table rows spell the numbers out.
   - Native `title` tooltips on all table column headers.

2. **WebUI: Bayes min_learns status badge** (backend + frontend)
   - `/stat`: expose `classifier.min_learns` in each statfile entry.
   - Classifier-cell badge reflecting the actual gating in `bayes.c`: a binary
     classifier classifies nothing while any class is below the threshold
     ("not classifying"); a multi-class one excludes the classes below it and
     stops only when all of them are below ("classes excluded").
   - Row-level warning triangle next to a class that has not reached
     `min_learns` yet, with the exact numbers in a tooltip.

## Compatibility

- Both features degrade gracefully on servers that do not send the new
  `classifier.min_learns` field (or predate classifier metadata entirely):
  no badge or marker; the balance bar keeps working off revisions.
- The badge and marker stay inert until the server is rebuilt.
- Gating semantics verified against `src/libstat/classifiers/bayes.c`
  (binary gate and multi-class viable/winner selection).

### DIFF
--- a/interface/css/rspamd.css
+++ b/interface/css/rspamd.css
@@ -641,6 +641,12 @@ input.action-scores {
     background-color: var(--rspamd-balance-special);
 }
 
+/* Class row marker: the class has not reached the classifier min_learns yet;
+   same warning tint as the cluster drift badge */
+#bayesTable .bayes-below-min {
+    color: var(--rspamd-text-warning);
+}
+
 /* Cluster (servers) table: details toggle, last-seen line, details row */
 #clusterTable td.cluster-toggle {
     width: 1%;
--- a/interface/css/rspamd.css
+++ b/interface/css/rspamd.css
@@ -35,6 +35,12 @@
     --rspamd-symbol-positive-hover-bg: #fbd6d1;
     --rspamd-symbol-special-hover-bg: #cddbff;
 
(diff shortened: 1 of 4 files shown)


moisseev:
@review deep

PR #6243, opened 2026-09-10
WebUI: fuzzy storage details and availability; fix empty storages missing from stat

moisseev:
## What

The Status tab fuzzy table now shows each storage's configuration and
availability next to the hash count, and a long-standing protocol quirk
that hid empty storages from the statistics is fixed.

## WebUI

- `read-only` badge for storages that cannot be learned to
- combined tooltip on the storage name with server addresses (unified or
  split read/write) and the symbol → flag mapping, taken from the
  existing `plugins/fuzzy/storages` endpoint
- the storages config is fetched once per page load (and on a
  configuration or up-set change), never in the /stat refresh cycle; a
  missing endpoint (older rspamd, `fuzzy_check` disabled) degrades
  silently to the plain table
- rules configured but missing from `fuzzy_hashes` render as
  `unavailable` with "-" instead of a count: the selected storage did
  not answer FUZZY_STAT in time (down, rate-limited or denied). A
  neighbour that recovers gets its metadata on the next refresh cycle
- header tooltips and a "No fuzzy storages" empty-state row, matching
  the bayes table conventions; hash counts stay raw and copyable

## Protocol fix

An empty fuzzy storage (count 0) was indistinguishable from an error
reply: the storage encoded the hash count in `prob`, so a zero count
arrived as `prob = 0` and the client's `prob > 0.5` guard dropped it —
the rule silently disappeared from `/stat` (`fuzzy_hashes`) and from
`/metrics` (`rspamd_fuzzy_stat`).

- the storage now replies with `prob = 1.0`; the count already travels
  in `value` + `flag` (same semantics as FUZZY_PING)
- both client reply handlers (UDP and TCP) also accept the legacy
  zero-count encoding (`prob = 0` with zero `value`/`flag`), so
  mixed-version clusters upgrade cleanly; error replies always carry a
  nonzero code in `value` and are unaffected

### DIFF
--- a/interface/index.html
+++ b/interface/index.html
@@ -256,9 +256,9 @@
 									<table class="table status-table table-sm table-bordered text-nowrap mb-0" id="fuzzyTable">
 										<thead class="text-secondary">
 											<tr>
-												<th>Server name</th>
-												<th>Storage</th>
-												<th>Hashes</th>
+												<th title="rspamd instance that reported this row">Server name</th>
+												<th title="Fuzzy storage rule; hover a storage for its servers, symbols and flags">Storage</th>
+												<th title="Number of fuzzy hashes in this storage">Hashes</th>
 											</tr>
 										</thead>
 										<tbody>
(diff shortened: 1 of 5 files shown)


moisseev:
@review deep

PR #6250, opened 2026-09-11
[Fix] Drop a grouped selector that cannot be evaluated

dragoangel:
## What breaks today

A selector list drops each of its parts down to the first simple selector that part starts with, and applies the declarations to that.

```css
div.mainbox ul li.spacer, div.mainbox ol li.spacer { font-size: 0 }
```

registers `div`. Every `<div>` in the document gets `font-size: 0`, and no class has to match anything. Without the comma the same rule is correctly ignored, which is why this only shows up in grouped rules.

Confirmed against a document where no class matches anything:

| rule | what disappears |
|---|---|
| `p.nomatch span.nomatch { font-size: 0 }` | nothing |
| `p.nomatch span.nomatch, x.other y.other { font-size: 0 }` | every `<p>` |
| `span.nomatch a.nomatch, q.other r.other { font-size: 0 }` | every `<span>` |
| `div.nomatch i.nomatch, q.other r.other { font-size: 0 }` | every `<div>` |
| `.onlyclass .other, .a .b { font-size: 0 }` | nothing (first part starts with a class) |

## Why it matters

Grouped descendant rules are everywhere in mail templates, and the ones carrying `font-size: 0`, `opacity: 0`, `color: transparent` or `display: none` are the spacer, preheader and mobile-only rules. When one of them collapses to a bare `div` or `td`, the whole body becomes invisible.

A real double opt-in confirmation, DKIM signed, hit this: its `div.mainbox ul li.spacer, div.mainbox ol li.spacer { font-size: 0; color: transparent; opacity: 0 }` became `div { font-size: 0 }`. The HTML part went from 81 words to 0, which is worth about 9 points:

* `HIDDEN_TEXT` 4.06 on 411 invisible characters
* `R_SUSPICIOUS_IMAGES` 5.0, since with zero words a single logo makes the image to text ratio 1.0

## The fix

`cur_selector` survived the transition into the ignore states, so a comma attached it as if it were complete. Reset it there, and let a comma restart parsing whether or not anything was attached.

## Scope and what is left

Nothing that used to match stops matching: plain selector lists (`p, span`, `em,.class,#id`) are unaffected, and a list that mixes both keeps the parts it can evaluate. What goes away is matching that was never asked for.

This does not lose hiding detection either: an ungrouped `div.x span.y { display: none }` is already ignored today, so grouped and ungrouped now behave the same.

Attribute selectors are untouched and still over-broad: the tokeniser folds `[...]` into a block, so the bracket never reaches this state machine and `a[href]` is registered as a bare `a`. Fixing that means changing how the block is consumed, which felt like a separate change. Say the word if you would rather have both in one go.

## Verified

* 410/410 C++ unit cases, 1344 Lua unit tests, no failures
* new `css` selector case covering grouped, compound, combinator and mixed lists
* a Lua unit case asserting the grouped rule no longer blanks the document while `.spacer { font-size: 0 }` still hides
* a functional fixture added to the existing `HTML VISIBILITY - No significant hidden content` template

### DIFF
--- a/src/libserver/css/css_selector.cxx
+++ b/src/libserver/css/css_selector.cxx
@@ -113,30 +113,40 @@ auto process_selector_tokens(rspamd_mempool_t *pool,
 				}
 			}
 			else if (state == selector_process_state::selector_ident_consumed) {
-				if (parser_tok.type == css_parser_token::token_type::comma_token && cur_selector) {
+				if (parser_tok.type == css_parser_token::token_type::comma_token) {
 					/* Got full selector, attach it to the vector and go further */
-					msg_debug_css("attached selector: %s", cur_selector->debug_str().c_str());
-					ret.push_back(std::move(cur_selector));
+					if (cur_selector) {
+						msg_debug_css("attached selector: %s", cur_selector->debug_str().c_str());
+						ret.push_back(std::move(cur_selector));
+					}
 					state = selector_process_state::selector_parse_start;
 				}
 				else if (parser_tok.type == css_parser_token::token_type::semicolon_token) {
 					/* TODO: implement adjustments */
 					state = selector_process_state::selector_ignore_function;
+					cur_selector.reset();
 				}
 				else if (parser_tok.type == css_parser_token::token_type::osqbrace_token) {
 					/* TODO: implement attributes checks */
 					state = selector_process_state::selector_ignore_attribute;
+					cur_selector.reset();
 				}
 				else {
 					/* TODO: implement selectors combinations */
 					state = selector_process_state::selector_ignore_combination;
+					/*
+					 * What follows narrows the selector down and we cannot
+					 * evaluate it, so the simple selector parsed so far no
+					 * longer describes the rule. Drop it: attaching it would
+					 * apply the declarations to every element of that name
+					 */
+					cur_selector.reset();
 				}
 			}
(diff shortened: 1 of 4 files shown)


vstakhov (approved):
Verified locally on top of the branch: the `css` doctest suite and the Lua HTML unit file pass, and the change is contained to the selector state machine.

The analysis is right: `cur_selector` leaked through the ignore states, so a comma (or EOF after an ignored tail) attached a simple selector that was only the prefix of something we never evaluated. Dropping the whole selector is the correct behaviour for a matcher that cannot evaluate it, and the mixed-list cases in the test pin down exactly that.

One pre-existing nit you may have noticed: the `semicolon_token` branch meant to catch pseudo-classes never fires, as the tokeniser emits `colon_token` for `:`. It falls through to the combination branch and gets the same treatment, so no behaviour difference, just a misleading comment.

Regarding the attribute selectors: agreed that it belongs to a separate change. The bigger gap is that compounds and combinators are dropped instead of evaluated, so `div.mainbox li.spacer { font-size: 0 }` is now invisible to hidden-text detection. I am going to add real support for compound selectors and descendant/child/sibling combinators on top of this fix, keeping the whitespace significance the parser currently throws away.

vstakhov:
Superseded by #6252

PR #6251, opened 2026-09-12
docs: add the estate-wide Non-Negotiable Rules to CLAUDE.md

smpaz7467:
Ten rules, identical in every repository — five non-negotiable (breaking them causes incidents) and five standards (expected, reviewable, not tripwires).

Each non-negotiable carries a **Why this exists** naming the real event behind it: the `.env` password in a transcript, the safeguard installed on neither server, the restore that "succeeded" having restored 0 files, the backup guard that fired correctly for 12 days and arrived green every time, the `.gitignore`/`.dockerignore` drift that baked two secrets backups into an image.

**Why identical text in every repo:** one set to learn, not one per app.

**Why the scars:** `server-ops/CLAUDE.md` is 337 lines with two rules at the top that name what they cost, and those are the rules that actually got followed today. The app CLAUDE.md files run to 1,300–2,400 lines and contained neither `a check must be able to fail` nor `not done until it is verified running on the target` — checked before writing, all zero.

Placed at the very top, above existing content, which is otherwise untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RzFzMxrnhaUrk4perzAJyo

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,3 +1,124 @@
+## Non-Negotiable Rules
+
+Breaking one of these causes an incident. Each has already cost us something —
+the "Why this exists" lines are real events, not illustrations.
+
+1. **Security first, always**
+   Before any change, perform a full threat model: what new attack surface is
+   introduced, what data flows in and out, what privileges are required, and
+   what happens if this code is compromised or fed malicious input. If the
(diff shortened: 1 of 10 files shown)


smpaz7467:
Opened in error against upstream — this was meant for an internal fork. Closing immediately; apologies for the noise.

PR #6268, opened 2026-09-22
Fix memory safety and framing defects in the chunked HTTP path

vstakhov:
Five defects in the chunked HTTP path, all reachable from an ordinary
unauthenticated request. Four of them are reproduced by the new unit suite,
which crashes the test binary under ASan without these fixes.

### The body allocation was sized from a number the peer chose

With `Transfer-Encoding: chunked` the parser reports no length when the headers
end, so `rspamd_http_on_body` creates the body storage instead, sized from
`parser->content_length`. For chunked input that field holds what is left of the
*current chunk* — a number the peer picked, with nothing received to back it up.

* A chunk size of `fffffffffffffff1` sized the allocation at
  `0xfffffffffffffff0`, which wrapped once `sizeof(rspamd_fstring_t)` was added.
  `malloc()` returned a tiny buffer while `allocated` kept the enormous length,
  and the first body byte was written outside it — ASan reports a
  heap-buffer-overflow in `rspamd_fstring_sized_new`.
* A merely large size ended the process inside `malloc` via `g_error`.
* Either way the allocation happened *before* `max_size` was consulted, so a
  configured message limit did not bound it.

The announced length is now only a hint, bounded by `max_size` and by a cap of
its own, and `max_size` is enforced over what has been received plus what is
still announced — so an oversized body is refused at its first fragment instead
of after room has been made for it.

`fstring` no longer wraps either: a length that cannot be allocated together
with the header is refused the way this file already treats an allocation it
cannot satisfy, and every intermediate in the growth arithmetic is clamped.
`rspamd_fstring_suggest_size` returns identical values to before for all normal
sizes (checked exhaustively over a grid of lengths up to 4 GiB).

### The chunk-size overflow check was incomplete

Both the decimal `Content-Length` and the hex chunk size were accumulated first
and checked afterwards, by comparing the result against the value it came from.
Unsigned wrapping does not reliably produce a smaller number, so some
overflowing lengths passed: a 17-digit chunk size such as `12000000000000000`
wrapped silently to `0x2000000000000000`. The check now happens before the
multiply.

### Callbacks could arrive after their message was released

Completing a message runs application code, and with keepalive the connection is
reset right there — which releases the message while the parser still has the
bytes that followed it to walk over. In zero-copy mode those bytes live in the
storage the message owned, and retaining the private buffer does not cover it,
because `zc_buf` only borrows that memory. A pipelined request after a chunked
one segfaults inside `http_parser_execute`.

The message is now held for the duration of the parse, and every callback that
needs a message returns when there is none.

### The last trailer was lost and leaked

Trailers use the ordinary header callbacks, but the end of the trailer section
reaches `on_message_complete` rather than `on_headers_complete`, so the last
trailer was never finished: it stayed in `priv->header`, never reached
`msg->headers`, and was never freed — one leak per message with a trailer.

### Chunked bodies inside encrypted requests were treated as contiguous

`rspamd_http_on_body_decrypted` recorded the first fragment's pointer and then
only added fragment lengths, so for a chunked inner body the reported range
contained the chunk framing and omitted part of the payload. The fragments are
now packed together as the plaintext path already does. An inner message that
stopped mid-framing was also accepted, because consuming every byte was taken
for completion; the inner parser now has a completion callback.

### Testing

`test/rspamd_cxx_unit_http_chunked.hxx` drives a server connection over a
socketpair. Against the unfixed code:

| case | before | after |
| --- | --- | --- |
| huge chunk size must not overflow the body allocation | heap-buffer-overflow (ASan) | pass |
| unallocatable chunk size is reported rather than fatal | allocation failure, abort | pass |
| chunk size wider than 64 bits is rejected | allocation failure, abort | pass |
| chunk size beyond max_size is refused | assertion failure | pass |
| message released at completion does not strand the parser | SIGSEGV | pass |
| trailer reaches the finished message | assertion failure | pass |
| multi chunk body is reassembled without framing | pass | pass |

Also green: full C++ unit suite (437 cases), Lua unit suite (1311 tests), and
functional suites 140, 220, 221 and 572 (34 tests) — 221 in particular covers
the early-response and keepalive client paths this touches.


vstakhov:
All three hold. Pushed e1b9c690d, 7ea4872e5, c1ff7c75f.

**P1, retaining the message does not preserve its body allocation** — correct,
and the retain was worse than insufficient: it was unsound. `rspamd_proxy.c`
(2968, 3118) and `http_router.c` call `rspamd_http_message_free()` **directly**,
bypassing the refcount, so the matching `unref` after the parse read a freed
refcount and took the proxy worker down. The milter suites catch that, because
milter mode reaches the backend over HTTP — 8 of them were failing on the branch
and passing on master, and I had nearly written them off as local environment
noise. The retain is gone.

The lifetime is fixed where you pointed: by preserving the input rather than the
object. The real defect was that the zero-copy *window* was sized only by the
room left in the body storage, so it reached past the end of the body and the
bytes after the message were read into it. The window is now bounded by what the
peer has announced is still to come — the rest of the current chunk, or the rest
of `Content-Length` — so a message can never complete with bytes left in the
window, and it no longer matters what a callback does with the storage.

Two arming sites needed it: `rspamd_http_switch_zc`, and a re-arm inside the body
callback that set `zc_remain` straight from the storage size and bypassed it. My
first attempt only fixed the former and the use-after-free still reproduced,
which is how the second one turned up.

Reproducer (`zero copy reads never reach past the end of the body`), before:

```
ERROR: AddressSanitizer: heap-use-after-free ... READ of size 1
    #0 http_parser_execute
0x627000004917 is located 4119 bytes inside of 12220-byte region
freed by thread T0 here:
    #1 rspamd_fstring_free
    #2 rspamd_http_message_storage_cleanup
    #3 rspamd_http_message_set_body
    #4 record_finish            <- the finish handler, as proxy_request_decompress does
```

**P1, chunked messages with Content-Length bypass the new allocation cap** —
confirmed, and worse than stated: `set_body` at header completion also sets
`HAS_BODY`, so the body callback's cap never runs at all. Also worth recording
that **nothing anywhere calls `rspamd_http_connection_set_max_size`**, so
`max_size == 0` in production and plain `Content-Length: <huge>` was equally
unbounded, chunked or not.

```
ERROR: AddressSanitizer: requested allocation size 0x1ff973cafa8010 ...
    #1 rspamd_fstring_sized_new
    #2 rspamd_http_message_set_body
    #3 rspamd_http_on_headers_complete
```

Both halves done: the cap now applies at header completion, and the parser
rejects a message carrying both framings in either header order. rspamd never
emits `Transfer-Encoding` (the proxy strips it) and ICAP builds its chunking over
raw TCP, so nothing internal relies on the combination.

One consequence to call out: bodies larger than the cap now grow geometrically
instead of in one exact allocation. The right way to get the exact size trusted
again is to give `max_size` a real value from `cfg->max_message`, but that starts
enforcing a limit on every HTTP connection in every worker, so I left it as its
own change rather than folding it in here.

**P2, inner completion remains true while a subsequent inner message is
incomplete** — confirmed by reading. The inner parser is now paused at inner
completion and the caller requires `nparsed == dec_len`, so there is exactly one
inner message and it ends exactly where the plaintext does; a second message's
callbacks never run at all. Pausing is safe there because it is a local parser
for that plaintext only — pausing the *connection* parser is not, it breaks the
early-response and keepalive paths in suite 221, which is why the fix for
released-message callbacks guards the callbacks instead.

This one has no unit test: a malformed double-inner-message needs hand-rolled
encryption. Happy to write it if you want it covered rather than reasoned.

Verification: 439/439 C++ unit cases (11 chunked), 1311 Lua. Full functional
suite run twice, on this branch and on master at 8a829eeda, same machine:
954 passed / 21 failed on both, and the two failure sets are **identical** —
no regressions, and the 21 are pre-existing local environment failures
(miltertest, MX loopback aliases, fuzzy map watcher, p0f crashing in
`__redisAsyncHandleConnectFailure`, language detection).

